PR #38, opened 2023-06-17
Add TLA+ spec for shamir secret sharing

pool2win:


docs/specifications/ShamirSecretSharing.tla
```diff
@@ -0,0 +1,130 @@
+------------------------ MODULE ShamirSecretSharing ------------------------
+(***************************************************************************)
+(* Sepcification for simple Shamir Secret Sharing. This is not a veriable  *)
```
nothingmuch, line 3:
```suggestion
(* Specification for simple Shamir Secret Sharing. This is not a verifiable *)
```

pool2win: Thanks. Committed change.

PR #51, opened 2023-09-11
Use clap for command line arguments

mcelrath:


pool2win:
Nice. Thanks for setting up clap.

PR #54, opened 2023-10-07
listen for new block hashes via ZMQ + fetch block templates via RPC

bernardoaraujor:
This PR makes the braidpool node able to listen for new block hashes via ZMQ, and fetch new block templates via RPC.
The node now takes some new CLI args, namely:
- `bitcoin`: IP address for bitcoin node
- `rpcport`: RPC port on bitcoin node
- `rpcuser`: username for RPCs on bitcoin node
- `rpcpass`: password for RPCs on bitcoin node
- `zmqport`: ZMQ port on bitcoin node

We spawn a `tokio` thread called `zmq::listener` that listens to the ZMQ subscriptions. When a new `hashblock` notification comes in, `block_template::fetcher` calls the `getblocktemplate` RPC and sends the new block template into a `mpsc` channel.

A dummy placeholder `tokio` thread called `block_template::consumer` consumes the `mpsc` channel by simply printing the new block templates into the console.


.gitignore
```diff
@@ -15,4 +15,5 @@ proposal/proposal.bbl
 proposal/proposal.blg
 simulations/logs/*
 simulations/*.toolbox
-venv
\ No newline at end of file
+venv
+.idea
```
pool2win, line 19: We can leave this in your local .git/info/exclude. The entire repo can avoid having ignores for all contributors editor dotfiles.

node/src/cli.rs
```diff
@@ -15,4 +15,20 @@ pub struct Cli {
...
+    /// Connect to this bitcoin RPC node
+    #[arg(long, default_value = "0.0.0.0:38332")]
+    pub rpc_url: String,
+
+    /// Use this username for bitcoin RPC
+    #[arg(long, default_value = "braidpooldev")]
+    pub rpc_user: String,
+
+    /// Use this password for bitcoin RPC
+    #[arg(long, default_value = "braidpooldev")]
```
pool2win, line 28: Lets take out the default username and password. Instead lets make these required. So we do not assume or proliferate default passwords that people use everywhere and thus get messed with.

node/src/template.rs
pool2win, line 1: template is very generic, lets call this module "blocktemplate", then we know immediately what we talking about.

node/src/main.rs
```diff
@@ -2,11 +2,15 @@ use clap::Parser;
...
 use tokio::net::{TcpListener, TcpStream};
+use tokio::sync::mpsc;
 use tokio_util::codec::{FramedRead, FramedWrite, LengthDelimitedCodec};
 
 mod cli;
 mod connection;
 mod protocol;
+mod template;
+
+use crate::template::{template_consumer, template_poll};
```
pool2win, line 13: The prefix "template" is being repeated here.

`use crate::block_template::{consumer, poll}` is cleaner.

Also, I prefer the following way to import and use those imports.

```
use crate::block_template
...
tokio::spawn(block_template::poll)
```

node/src/template.rs
```diff
@@ -0,0 +1,52 @@
...
+        .unwrap();
+    template_tx.send(last_template.clone()).await.unwrap();
+
+    loop {
+        // sleep until it's time to poll again
+        sleep(Duration::from_secs(poll_interval.clone())).await;
+
+        let template = rpc
+            .get_block_template(GetBlockTemplateModes::Template, &BLOCK_TEMPLATE_RULES, &[])
+            .unwrap();
```
pool2win, line 35: unwrap here and on L39 will kill the loop and the program. We should explicitly handle the error and then continue with the loop. The BIP also says we should do exponential backoff before trying again.

1. Handle error
2. Do exponential backoff with parameters as module constants
3. If after cumulative retry period we are still seeing an error then we should print meaningful error and do a proper shutdown. We don't have proper shutdown yet, so for now a process::exit will do. Added an issue for this https://github.com/braidpool/braidpool/issues/55

pool2win (commented):
Nice work. Left some nitty comments and some error handling fix requests.

We should move the mutinynet and docker setup files in a separate PR. Since these are dev tools, I am not sure yet if we should have these in the repo.

node/src/template.rs
bernardoaraujor: I understand your point on proper error handling and I agree on the need for a more elegant strategy. I'm working on something.

But just to clarify, L39 would throw an error under different circumstances.
- `template = rpc.get_block_template(...).unwrap()` failure depends on the RPC result.
- `template_tx.send(template.clone()).await.unwrap()` failure depends on the consumer of the `mpsc` channel being dropped. For now I only wrote a dummy `fn consumer` which makes it unclear under which circumstances this scenario will arise.

bernardoaraujor:
From my experience working on this PR, mutinynet is really useful because waiting ~10min for new blocks on a development workflow is not viable. This will likely be the case for future development as well.

But I agree it's perhaps not ideal to impose this into the codebase as a whole, so mutinynet deployment files were moved over to https://github.com/bernardoaraujor/braidpool-mutinynet

node/src/template.rs
bernardoaraujor: just proposed a solution for exponential backoff on https://github.com/braidpool/braidpool/pull/54/commits/98e88116717a5072fb8e567f6942883eb1f78876

the `unwrap` on the `mpsc` channel remains, however.

waiting for your input on this one @pool2win

mcelrath:
Thanks for this, good work! But polling bad. Bitcoin provides a [ZMQ](https://github.com/bitcoin/bitcoin/blob/master/doc/zmq.md) interface:

    https://github.com/bitcoin/bitcoin/blob/master/doc/zmq.md

This will push block notifications to us.

We're going to be very sensitive to latency, with braid bead times that might be as low as 150ms. 

How do you feel about rewriting this to use ZMQ?

bernardoaraujor:
There's a few different options on how to go about this ZeroMQ implementation. Here's a summary of the tradeoffs for each option:
1. [`bitcoincore-zmq`](https://crates.io/crates/bitcoincore-zmq): goes straight to the point, but the project is very new and seems WIP. I did some tests and while it does fill our needs, the behavior of the subscription didn't seem very stable.
2. [`bitcoin-net-zmq`](https://crates.io/crates/bitcoin-net-zmq): also has the advantage of being a Bitcoin+ZMQ implementation. It is actually part of the larger effort [`bitcoin-rs`](https://github.com/klebs6/bitcoin-rs), which is a WIP translation of C++ into Rust. There's no documentation.
4. [`zmq`](https://crates.io/crates/zmq): the most popular crate for ZMQ, and the most likely to recieve long-term maintenance. Doesn't provide `async` out-of-the-box, which forced me to slightly increase the complexity of the code by using things like `tokio::task::spawn_blocking`, `Arc` and `Mutex`.
5. [`async_zmq`](https://crates.io/crates/async_zmq): an async port of the crate mentioned above, which allows for simpler code. The project is relatively smaller than its sync counterpart, and I wonder if it could be abandoned in the future.
6. [`futures-zmq`](https://crates.io/crates/futures-zmq): pretty much dead.
7. [`tokio-zmq`](https://crates.io/crates/tokio-zmq): pretty much dead.

While searching for `zmq` on `crates.io` there's actually many other options. But most of them follow the patterns described above.

If I were to choose the best dependency for the long term, option 3 seems the most appealing to me. Please let me know your thoughts @mcelrath @pool2win 

I have a working implementation based on option 3 that needs some cleaning up but I should be able to push here soon.

bernardoaraujor:
another question that comes to mind is:

will braidpool nodes also need to listen for other ZMQ topcis aside from `hashblock`?

mcelrath:
IMHO, no one should be writing non-async code in the 21st century. I suspect using one of the bitcoin-specific ones might be a better long-term dependency. But ZMQ is pretty universal. Option 4 is using a transpiler which generated a LOT of code, so probably a bad choice.

What's wrong with the `zmq`'s suggestion of using the `async-std-runtime`? I know something @kulpreet wrote used `tokio` as well. I'm leaning towards Option 1 here. What's wrong with using async the way they describe?

As to receiving other messages, I think eventually we will. `getblocktemplate` is a very slow RPC that we need to make fast. I know bitcoin core has done some work on making this faster and I'd like improvements to it to go into bitcoin core instead of braidpool. But things we've discussed include a fast "block diff" that removes txs included in the last block but otherwise re-uses the last block template we got from bitcoin core. This way we can avoid the expensive fee calculations that make getblocktemplate slow and just recompute the Merkle tree. But again I'd like to see that in bitcoin core. But changes to bitcoin core are like molasses and we might find ourselves writing a "block diff" in braidpool to bridge the gap.

There are also potential use cases for seeing mempool txs. Braidpool needs to watch its wallet payouts for instance and will probably want to see mempool txs that affect its wallet. Alerting on non-authorized spends of Braidpool coinbases would be another use case. (If a catastrophic fail, but we'd want to notice)

mcelrath:
Another major reason to actually deserialize a block is to examine the coinbase transaction (which is not broadcast as a regular tx in any other way). We need to determine if this block is a braidpool block and follows the rules that trigger a bunch of other work (adding the miner of this block to the payout FROST signing quorum for instance), as well as updating that miner's shares. A bitcoin block is a valid bead in our DAG. We could for instance receive the block before the corresponding DAG bead (which contains extra metadata [as described in the spec](https://github.com/mcelrath/braidcoin/blob/master/braidpool_spec.md)). If we haven't received the DAG bead yet, we still need to switch work to name that block as a parent.

We might consider embedding useful metadata in the coinbase as well. For instance, we could commit to the parents in the OP_RETURN. There's not enough space to put 32-byte hashes for more than one parent, but we might consider putting `hash({parents})` there. Generally there will only be a couple of possible parents so it's not a huge computation to exhaustively list every possible combination of parents. Or use Fountain codes to include the first few bytes of all the parent block hashes. Collisions here should not be a concern. This would let us keep the DAG structure even if bead metadata is delayed.

mcelrath:
The issue here is that we shouldn't name grandparents of a bead as parents. But if we don't know the parents of the current block, we could do so accidentally.

It's also not a problem to just name the recently mined block as a parent and tie up any dangling parents once the metadata is available. This creates diamond/pentagon graphs in the DAG and increases our latency measurement, but this isn't a problem. (The latency IS actually high and data got delayed, so it's right that the DAG reflect this)

bernardoaraujor:
> I'm leaning towards Option 1 here. What's wrong with using async the way they describe?

I lean towards Option 1 as well, IMO it is the best bet for a long-term sustainable dependency. The current implementation on this PR is based on this dependency.

There's nothing wrong with this approach, it is just slightly more verbose when compared to using Option 2, because we need to call non-async code inside an async function.

Basically, with Option 1 we end up doing this:
```
pub async fn listener(...) {
    ...
    loop {
        let zmq_clone = Arc::clone(&zmq);

        match tokio::task::spawn_blocking(move || {
            let zmq = zmq_clone.lock().expect("lock zmq socket");
            zmq.recv_multipart(0)
        })
        .await
        .expect("receive zmq notification")
        { ... }
    }
}
```

while Option 2 would allow us to achieve the same result with:
```
pub async fn listener(...) {
    ...
    while let Some(msg) = zmq.next().await {
        match msg { ... }
    }
}
```

I just felt I should highlight all the pros and cons of all potential paths, but Option 1 seems worth the trade-off to me.

bernardoaraujor:
in regards to decoding the ZMQ notification, I will make some adaptations so that `listener` is able to deserialize and deal with all possible notification topics, namely:
- `hashtx`
- `hashblock`
- `rawblock`
- `rawtx`
- `sequence`

mcelrath:
@bernardoaraujor I see what you mean. The zmq code is much less readable... I have a hard time understanding what the first example is doing while the second is very clear. I'm a big fan of the way python implemented async/await. It took them a few rounds experimenting with different syntaxes, looks like Rust will be going through that in the next couple of years.

No to further muddy the waters, but there a couple of tokio zmq crates: [TMQ](https://github.com/cetra3/tmq), [zmq-tokio](https://github.com/rotty/zmq-tokio), and [tokio-zmq](https://crates.io/crates/tokio-zmq) (of those it looks like only TMQ is alive). Then there's [rust-zmq](https://github.com/erickt/rust-zmq)...

Well this is a bit of a headache.

Anyway, your rust experience is certainly more extensive than mine and I'm going to take your recommendation here.

bernardoaraujor:
> Then there's [rust-zmq](https://github.com/erickt/rust-zmq)...

[rust-zmq](https://github.com/erickt/rust-zmq) was originally listed as Option 1.

Since we have to decode all ZMQ messages, I reconsidered [`bitcoincore-zmq`](https://crates.io/crates/bitcoincore-zmq) (Option 3). Even though the project is still WIP, it implements everything we need for now (means for decoding `hashtx`, `hashblock`, `rawblock`, `rawtx`, and `sequence` topics).

The crate has a tight [dependency list](https://crates.io/crates/bitcoincore-zmq/1.2.0/dependencies):
- `rust-bitcoin`
- `rust-zmq` (Option 1)

So it starts to make sense in my mind the usage of `bitcoincore-zmq`. The only downside is the fact that it is non-async.
So as a compromise, I forked it into [`tokio-bitcoincore-zmq`](https://github.com/bernardoaraujor/tokio-bitcoincore-zmq), which allowed this concise code:
```
let mut zmq = bitcoincore_zmq::subscribe_single(&zmq_url).await?;

    while let Some(msg) = zmq.recv().await {
        match msg {
            Ok(msg) => {
                match msg {
                    bitcoincore_zmq::Message::HashBlock(_, _) => {
                        // call `getblocktemplate` RPC
                        todo!();
                    }
                    bitcoincore_zmq::Message::HashTx(_, _) => todo!(),
                    bitcoincore_zmq::Message::Block(_, _) => todo!(),
                    bitcoincore_zmq::Message::Tx(_, _) => todo!(),
                    bitcoincore_zmq::Message::Sequence(_, _) => todo!(),
                };
            }
            Err(err) => return Err(err),
        }
    }
```

bernardoaraujor:
after going back and forth with @antonilol and the `bitcoincore-zmq` crate is on its way to support an `async` feature via [this PR](https://github.com/antonilol/rust-bitcoincore-zmq/pull/4/files)

so I'm switching to `bitcoincore-zmq` as dependency and I'm archiving my `tokio-bitcoincore-zmq` fork.

mcelrath:
**tested**

So the standard way to authenticate to bitcoind is to use cookie authentication, which is the file `~/.bitcoin/.cookie` and allows the local user to use bicoind without `rpcuser` and `rpcpass`. Looks like we can do this:

https://github.com/rust-bitcoin/rust-bitcoincore-rpc/blob/master/client/src/client.rs#L198

Before I merge this, can you get cookie auth working by default (and optional rpcuser/rpcpass)?

mcelrath:
Could you add some debugging around the zmq connection? I didn't have zmq enabled in bitcoind and got no complaints from the braipdool node. e.g. fail if we can't connect to zmq.

Secondly can you add some docs around configuring the bitcoin node to publish zmq? I think it's just
```
zmqpubrawblock=tcp://localhost:28332
```

Do you understand the "high water mark" config options bitcoin has for zmq?

antonilol:
> Could you add some debugging around the zmq connection? I didn't have zmq enabled in bitcoind and got no complaints from the braipdool node. e.g. fail if we can't connect to zmq.

this is a problem i have had with the `zmq` crate (as a dependency of my `bitcoincore-zmq` crate). it appears to be a zmq protocol problem later fixed with a socket option (see https://stackoverflow.com/a/75893170/13800918), i will look into this

antonilol:
> Do you understand the "high water mark" config options bitcoin has for zmq?

https://zeromq.org/socket-api/#high-water-mark

antonilol:
i did not get it working with the socket options (`ZMQ_HEARTBEAT_*`, see http://api.zeromq.org/4-2:zmq-setsockopt), but after a lot of google searching and interrogating chat gpt i consulted the ultimate network debugging tool: wireshark. there is apparently a handshake between the subscriber and the publisher, where the subscriber could make up from if the publisher exists. the only problem is that i do not know if zmq exposes this (i think not). it might be possible but it could require patching zmq (only the subscriber, i do not consider any changes to the publisher a viable option)

antonilol:
update on [this](https://github.com/braidpool/braidpool/pull/54#issuecomment-1785871997) comment: it is possible and libzmq exposes functionality for it, and the rust bindings do too.

i made a working proof of concept pr: https://github.com/antonilol/rust-bitcoincore-zmq/pull/7, note that the stream will be reusable eventually, that is just not implemented yet

PR #57, opened 2023-10-24
litep2p

bernardoaraujor:
Exploring usage of [`litep2p`](https://github.com/altonen/litep2p) for the p2p communication layer of a `braidpool` node.

The usage of [`QUIC`](https://datatracker.ietf.org/doc/html/rfc9000) (low-latency general-purpose transport layer) opens some interesting possibilities, given that `braidpool`'s protocol efficiency is sensitive to latency. Even though it is `UDP`-based, `QUIC` has mechanisms for packet loss detection and recovery.

We can use `litep2p::protocol::notification` to announce new beads in a gossip fashion, and `litep2p::protocol::request_response` for syncing the braid.


mcelrath:
I think using s2n-quic is a good idea. libp2p looks like a little used one-man operation and it's not clear to me what it really adds. One less dependency... s2n-quic looks very comprehensive.

bernardoaraujor:
@mcelrath I recently shared comments about `litep2p` on the Matrix chat with extra context.

I believe it's an interesting direction, but I'll leave that for you guys to decide.

mcelrath:
Yeah just logged into Matrix and saw your comments there.

I'm not opposed to litep2p but can you summarize what it brings over QUIC directly?

bernardoaraujor:
`litep2p` is a lightweight rewrite of [`rust-libp2p`](https://github.com/libp2p/rust-libp2p)

`libp2p` is widely adopted in the web3 space ([ethereum just ditched `devp2p` for it after the Merge](https://blog.libp2p.io/libp2p-and-ethereum/), and IPFS is entirely based on top of it).

`libp2p` provides multiple p2p communication abstractions that could prove useful for `braidpool`. This PR is exploring the usage of `litep2p::protocol::notification` for Bead Announcement and `litep2p::protocol::request_response` for Braid Syncing.

@kulpreet mentioned that `libp2p` was considered in the past but discarded because the current codebase is bloated. And I discovered recently that `litep2p` aims to tackle this exact same issue, and since it will receive long term and quality support, it seemed like an interesting path for exploration here.

The benefit for `braidpool` would be mostly reducing the heavy-lifting engineering effort for p2p comms on `braidpool` code. `QUIC` is a nice-to-have that comes out-of-the-box, but as already discussed, could also be achieved without `litep2p`.

mcelrath:
Here are some other resources. Given the lack of FEC support, anywhere, I think we should go with libp2p and QUIC and possibly upgrade it or contribute FEC code to it at a later time. The impact of high latency will be to slow down the bead rate in the DAG, but that's dynamic and measured anyway.

https://airhook.ofb.net/

https://ieeexplore.ieee.org/document/9013401

 https://github.com/pgOrtiz90/quic-go-fec

s2dd:
Quick blog post for reference :

mcelrath:
> Quick blog post for reference : (https://quic.video/blog/quic-powers/)

Would like to see latency measurements. The rejection of FEC in the standard is a big red herring that we might have to fix later. But please show us measurements if we're wrong.

mcelrath:
TL;DR we need low latency. Would prefer not to build it ourselves if there's a good solution out there. But everything I've seen measures latency in dumb ways (multiple requests/multiple responses, video delivery, etc). We're looking for small packet latency, not any kind of aggregate.

s2dd:
Latency is dependent on router's between endpoints,
i guess a stream solution will be the quickest, then considering having a p2p graph where peer only have max(8?) streams live at any time in the p2p network need work.
Big Data that has need for speed uses OWN routers with fiber between (10-20ms). 
This not possible in the open wild, so to give answer about measurements of latency of one kind of impl running on the open inet -the answer would be it depends.

s2dd:
If SQLite is to be used, then to offload some of the node's db work one can consider : https://www.sqlite.org/wal.html
in the search for speed ...

s2dd:
Sample source of a discovery beacon : 
https://github.com/lemunozm/message-io/tree/master/examples/distributed

pool2win:
Thanks for the exploratory effort here.

I am not in a rush to explore optimisations just yet. The focus is on the getting the functionality working and avoid being bogged down by optimisations. Once we are ready to start optimising and decide QUIC will be helpful, we can replace TCPStream with a QUIC implementation like s2n-quic.

Thanks again.

mcelrath:
For those following, we have not made a decision about this yet and hope to discuss in-progress ideas in the #65 discussion. (We're trying "Discussions" because we don't want to have 10k Issues and PRs like Bitcoin core. Thank you very much @bernardoaraujor for your contributions and they could well come back. We haven't rejected this idea.

PR #74, opened 2024-08-14
Check for existence of data directory and create it if necessary.

mcelrath:


node/src/main.rs
```diff
@@ -19,8 +20,19 @@ async fn main() -> Result<(), Box<dyn Error>> {
     setup_logging();
     setup_tracing()?;
 
-    let datadir = args.datadir;
-    log::info!("Using braid data directory: {}", datadir.display());
+    let datadir = shellexpand::full(args.datadir.to_str().unwrap()).unwrap();
```
pool2win, line 23: We should move this to a function and then add a unit test for it. Or we can have an integration test that tests the directory is created properly. Unit test will be better.

PR #76, opened 2025-01-13
Add python tests and example braid files

mcelrath:


tests/braid.py
```diff
@@ -0,0 +1,679 @@
...
+    for p in parents[b]:
+        if p not in ancestors:
+            all_ancestors(p, parents, ancestors)
+        ancestors[b].update(ancestors[p])
+
+def all_ancestors_iterative(b, parents, ancestors):
+    """ Gets all ancestors for a bead <b>, filling in ancestors of any other ancestors
+        encountered, using an iterative algorithm.
+        FIXME broken
+    """
```
Aritra8438, line 91: Hey, @mcelrath.
Do you need a helping hand here?

mcelrath: Pushed an iterative version below and added a test case comparing it to the recursive version. Seems to work but it isn't any faster. Anyway, the Rust translation should probably not use a recursive implementation.

PR #88, opened 2025-03-26
Place beads both above and below the highest work path

Aritra8438:
fixes: `FIXME` place beads both above and below the highest work path.
fixes: `FIXME` will draw lines over other beads when a bead is in both the head and tail.

modifies `layout()` to return the previous cohort tips and the minimum negative height of the bead for each cohort and then lift the whole braid up.

### DIFF
--- a/tests/braid.py
+++ b/tests/braid.py
@@ -322,7 +322,7 @@ def check_cohort_ancestors(cohort, parents, children=None):
         return False
     return True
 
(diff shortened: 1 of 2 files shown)


Aritra8438:
Done @mcelrath.
Sorry, I wanted to keep a commit for every major changes and the two revert commits were due to the revert of `.ipynb` and a commit that exists in `mcelrath/braidpool`.

PR #92, opened 2025-03-30
Bundle node as its own crate

Calisto-Mathias:


mcelrath:
Looks good

mcelrath (changes requested):
Can you get the workflows working?

Calisto-Mathias:
> Can you get the workflows working?

I'll try and see what is the issue with the workflows. Does this have anything to do with my account? GitHub seems to have disabled Actions but only for my account. I have put in a GitHub ticket for this as well. But I don't think that has anything to do with the workflow not running on the repo itself right? since I don't own it.

mcelrath (approved):

PR #94, opened 2025-03-30
Braidpool primitives crate merge

Calisto-Mathias:


.github/workflows/rustfmt.yml
```diff
@@ -0,0 +1,25 @@
+name: Rust Format Check
+
+on:
+  push:
+    branches: [ main, dev ]
+  pull_request:
+    branches: [ main, dev ]
```
Aritra8438, line 7: Do we need to have `dev` on this?

PR #103, opened 2025-04-02
Add Consensus Functions

TypiCal25:
This Pull Request adds in all the functions/error definitions/helper functions for the consensus algorithms implemented in braidpool.


braidpool-primitives/src/braid.rs
```diff
@@ -151,4 +149,29 @@ impl Braid {
     fn calculate_valid_difficulty_for_bead(&self, bead: &Bead) -> CompactTarget {
         unimplemented!()
     }
-}
\ No newline at end of file
+}
+
+use std::fmt;
+#[derive(Debug)]
```
Sansh2356, line 155: Can add the Clone trait also @TypiCal25 ?

PR #106, opened 2025-04-02
feat: add api to run simulator

its-me-abhishek:
- Integrated Flask-SocketIO to stream live braid simulation updates in a json format
- Enabled CORS for frontend calls
- Implemented a background thread to run simulations continuously
- Exposed braid.json via an API route for direct data access via a websocket

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,0 +1,63 @@
+# Simulator API
+
+This README provides instructions on how to run the `simulator_apil.py` script and access its deployed URL, Swagger documentation, and endpoints.
+
+## Prerequisites
+
+Ensure you have the following installed:
+
+- Python 3.8 or higher
+- Required dependencies (install via `requirements.txt`)
+
+## Setup
+
+Install dependencies:
+
+```
+    pip install -r requirements.txt
+```
+
+## Running the Simulator
+
+1. Run the script:
+
+   ```
+   python simulator_api.py
+   ```
+
+2. The API will start, and you can access it at:
+
+   ```
+   http://127.0.0.1:65433
+   ```
+
+3. For testing and API documentation, refer to:
+   ```
+   http://127.0.0.1:65433/apidocs
+   ```
+
+## Accessing the Deployed URL
+
+You can access the API at the provided URL:
+
+Websocket:
+
+```
+http://french.braidpool.net:65433/
+```
+
+## Deployed Swagger Documentation
+
+To view the deployed Swagger UI for API documentation, navigate to:
+
+```
+http://french.braidpool.net:65433/apidocs
+```
+
+## Endpoints
+
+Refer to the Swagger UI for a detailed list of available endpoints and their usage.
+
+## Notes
+
+- Ensure the server is running before accessing the endpoints if running locally.
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -3,3 +3,5 @@ matplotlib==3.10.1
 numpy==2.2.4
 pytest==8.3.5
 scipy==1.15.2
+flask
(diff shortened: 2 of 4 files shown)


rasta-dave:
I've reviewed PR #106 for the simulator API. This looks like a solid foundation for enabling real-time data streaming from the simulator to a frontend dashboard.

Some observations:

1. The implementation with Flask-SocketIO provides a good solution for pushing live updates to the dashboard.

2. The background thread approach allows for continuous simulation without blocking the web server which is good.

3. Some potential improvements to consider:
   - Add error handling around the simulator run to prevent crashes if the simulator encounters issues. Would be very helpful
   - There's no way to stop/pause the simulation from the API but maybe this is not that important

4. For the dashboard integration:
   - Would be helpful to have some documentation on the format of the data being emitted over the socket!
   - Adding a simple endpoint to retrieve the current simulation state without waiting for the next update could be useful

Overall, this PR provides the foundation we need for building the real-time visualization components of the dashboard. With some minor refinements, I think this will work well for our needs

Sansh2356:
@its-me-abhishek looks a good starting point for dashboard , can be improved on the go .

mcelrath:
I'd really prefer that this import the simulator module rather than calling it on the command line and reading its json file. The code below is from the `Braid Examples.ipynb` file which you should take a look at. In non-mining mode it generates beads very fast, but by re-using the same `Network` object and calling with it `n.simulate(nbeads=1, mine=False)` and then e.g. `sleep(0.5)` in a loop, it will generate one bead every 500ms and should push updates using asynchronous websockets that the UI should respond to. The basic objects and how to retrieve them (parents, children, work, hwpath, cohorts) and how to extract them from the simulator are below:

```python
from simulator import *
# A larger network or lower difficulty makes thicker braids
# Re-run this block to generate a different braid.
# The target and hashrate parameters are measured on my CPU (Intel 3.3 GHz) Run `simulator.py --mine` to benchmark your CPU
n = Network(nnodes=50, target=2**240-1, hashrate=800000)
# Use `mine=False` to use the geometric distribution to simulate mining. Use `mine=True` to actually compute sha256d hashes (slower)
n.simulate(nbeads=30, mine=False) 

# Each node has its own braid, look at node 0's braid here.
b = n.nodes[0].braid
# Extract the `parents` dict as integers (hashes) instead of instances of the <Bead> object
hashed_parents = {int(k): set(map(int, v)) for k,v in dict(b).items()}
# Remap integers (hashes) in the parents array to sequential integers for display purposes
parents = braid.number_beads(hashed_parents)
children = braid.reverse(parents)
work = braid.descendant_work(parents)
hwpath = braid.highest_work_path(parents, children)
cohorts = list(braid.cohorts(parents))
print(f"Generated braid: {len(b.beads)} beads with {len(cohorts)} cohorts, {len(b.tips)} tips, "
      f"and {len(hwpath)} beads in the highest work path.")
```

priyashuu:
PR #106 -

Hey ,

PR #106 establishes a strong base for real-time data streaming with Flask-SocketIO, utilizing a background thread to maintain server responsiveness. A few refinements to consider:

 
- **Pause/Stop Simulation**: Not critical but could be useful for debugging.  
- **Data Format Docs**: Helps frontend integration.  
- **Fetch Current State**: An API endpoint to get the latest state without waiting.  
- **Asynchronous Optimization**: Consider `socketio.start_background_task()` for efficiency.  
- **Logging & Monitoring**: Add structured logging for debugging.  

These improvements would enhance reliability and scalability. Let me know if you'd like more details!

its-me-abhishek:
Closed in favour of #126

PR #107, opened 2025-04-02
Initial workflow for testing functions for braid.rs and bead.rs

Sansh2356:
What does this PR do - 

- Adds the separate unit tests for private functions for `bead` and `braid` respectively .
- Adds the integration tests for public functions for `bead` and `braid` respectively.
- Inherit trait such as `Clone` , `debug` etc .  

Further addition of testcases in data directory is to be done along with few changes related to refactoring of functions as well .


braidpool-primitives/src/braid.rs
```diff
@@ -151,4 +151,238 @@ impl Braid {
...
+                    Time::from_consensus(1690001000).expect("invalid time value"),
+                ),
+                (
+                    BeadHash::from_byte_array([0x03; 32]),
+                    Time::from_consensus(1690002000).expect("invalid time value"),
+                ),
+            ],
+            vec![[0x11; 32], [0x22; 32], [0x33; 32], [0x44; 32]],
+            436864982,
+            [0x00; 32],
```
Aritra8438, line 231:
```suggestion
            [0x01; 32],
```

braidpool-primitives/src/braid.rs
```diff
@@ -151,4 +151,238 @@ impl Braid {
...
+            [0xbb; 32],
+            [0xbb; 32],
+            4040404,
+            vec![(
+                test_dag_bead_1.bead_hash,
+                Time::from_consensus(1690000000).expect("invalid time value"),
+            )],
+            vec![[0x11; 32], [0x22; 32], [0x33; 32], [0x44; 32]],
+            436864982,
+            [0x00; 32],
```
Aritra8438, line 256:
```suggestion
            [0x02; 32],
```

Aritra8438 (changes requested):
`BeadHash` is ideally a unique identifier. They are like `blockchain hash` and therefore should be different.

braidpool-primitives/src/braid.rs
Aritra8438: Similarly, update the parents too.

braidpool-primitives/src/utils/test_utils.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+pub fn create_test_bead(
+    block_version: i32,
+    prev_blockhash_bytes: [u8; 32],
+    merkle_root_bytes: [u8; 32],
+    timestamp: u32,
+    difficulty_bits: u32,
+    nonce: u32,
+    bead_hash_bytes: [u8; 32],
+    coinbase_tx_bytes: [u8; 32],
```
Aritra8438, line 74:
```suggestion
    nonce: u32,
    coinbase_tx_bytes: [u8; 32],
```
If I'm not wrong, `bead_hash_bytes` is not used at all and the `bead_hash` is in the `bead_bytes`.

braidpool-primitives/src/braid.rs
Aritra8438: Okk, suitable directory means?

We should ideally keep the unit tests in the same file as the `impl`.

Sansh2356: Yep in the utility directory i will add `data` directory storing the corresponding json files related to test cases.

braidpool-primitives/src/utils/bitcoin.rs
```diff
@@ -1,31 +1,62 @@
...
+        &self,
+        testing_option: Option<(String, Vec<Transaction>)>,
+    ) -> MerkleRoot {
+        if testing_option.is_some() == true {
+            let tuple_val = testing_option.unwrap();
+            let default_order = tuple_val.1;
+            let tx_ids = default_order.iter().map(|obj| obj.compute_txid());
+            let merkel_root = TxMerkleNode::calculate_root(tx_ids);
+            println!("Merkel path proof ---- {:?}", merkel_root);
+            return merkel_root.unwrap();
```
Aritra8438, line 32: Using `unwrap()` without `.is_some()` might panic, is this intended?
What if the `calculate_root` function returns `None`?

Also, I would prefer using this structure.
```rust
if cond {
	...
	merkel_root.unwrap()
}
else {
	...
	TxMerkleNode::from_byte_array(current_hash)
}

But it's fine if you prefer this or if there's a reason.

File: braidpool-primitives/src/utils/bitcoin.rs [hunk 5]
```diff
@@ -1,31 +1,62 @@
...
+        }
+        //transactionid along with its sibling
+        //taking hash and converting it to its raw bytes
+        let irreversed_id = self.transaction_hash.to_string();
+        let reversed_curr_tx_id = reverse_hex_byte_order(&irreversed_id);
+        let mut current_hash = [0u8; 32];
+        let result = hex::decode_to_slice(reversed_curr_tx_id, &mut current_hash);
+        match result {
+            Ok(val) => {
```
Aritra8438, line 41:
```suggestion
            Ok(_) => {
```

Aritra8438: If you think the suggestion is fine, you can just commit it.

braidpool-primitives/src/utils/bitcoin.rs
Sansh2356: Sure @Aritra8438 thought the same, can be done .

braidpool-primitives/src/utils/bitcoin.rs
Sansh2356: Yep this is indeed required also instead of `println` we can do `eprintln` macros ?

braidpool-primitives/src/utils/bitcoin.rs
```diff
@@ -1,31 +1,62 @@
...
+    pub fn calculate_corresponding_merkle_root(
+        &self,
+        testing_option: Option<(String, Vec<Transaction>)>,
+    ) -> MerkleRoot {
+        if testing_option.is_some() == true {
+            let tuple_val = testing_option.unwrap();
+            let default_order = tuple_val.1;
+            let tx_ids = default_order.iter().map(|obj| obj.compute_txid());
+            let merkel_root = TxMerkleNode::calculate_root(tx_ids);
+            println!("Merkel path proof ---- {:?}", merkel_root);
```
Aritra8438, line 31: Can you explaun this?
Given a `transaction with merkle proof`, I would expect to calculate the merkle root by hashing the `txid` with the first element of the vector and then the 2nd one and so on.

Sansh2356: This is not on the basis of `merkel path` though as i said the siblings ordering is left .

PR #109, opened 2025-04-03
Add functions related to consensus

Aritra8438:
This PR adds functions related to consensus.

In this PR, I'll also add tests.


Sansh2356 (approved):
Looks fine with the functions , after adding tests further work can be continued from my part @Aritra8438 .

braidpool-primitives/src/braid.rs
```diff
@@ -86,6 +93,69 @@ impl Braid {
 
 impl Braid {
     // All private functions go here!
+    fn get_parents(&self) -> HashMap<BeadHash, HashSet<BeadHash>> {
```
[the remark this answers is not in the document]
TypiCal25: Also another suggestion I have for this, is that we break this code into a nice set of cleaner functions
1. `fn get_children_for_bead(&self, bead: BeadHash) -> Result<Children, some error type> `. Here the error type would be an enum like the one I defined in my PR in case you put in a BeadHash that hasn't been ever recorded by the node.
2. `fn is_bead_a_parent(&self, bead: BeadHash) -> Result<bool, same error type>`. This function is simply a wrapper on top of the previous function that just returns whether the HashSet that is returned is not empty.

This would make our code a whole lot cleaner I feel.

braidpool-primitives/src/braid.rs
```diff
@@ -86,6 +93,69 @@ impl Braid {
...
+        let mut parents = HashMap::new();
+        for beadhash in self.beads.iter() {
+            let bead = match self.loaded_beads_in_memory.get(beadhash) {
```
TypiCal25, line 99: This should be handled by another function itself. In fact, I think I already added this function in my PR. I think for now, this is fine, but I'll replace it once both our PRs are reviewed and merged to use a single `load_bead` function.

braidpool-primitives/src/braid.rs
```diff
@@ -86,6 +93,69 @@ impl Braid {
...
+                Some(bead) => bead,
+                None => continue, // TODO: Load from the database
+            };
+            let mut this_parents = HashSet::new();
+            for (parent_hash, _) in &bead.parents {
+                this_parents.insert(parent_hash.clone());
+            }
+            parents.insert(beadhash.clone(), this_parents);
```
TypiCal25, line 107: Also, i think in this line, the BeadHash type (which is a type alias for BlockHash) has already implemented the Copy trait. So we can shorten this code, and make it more clear, by removing the `.clone()`.

braidpool-primitives/src/braid.rs
```diff
@@ -86,6 +93,69 @@ impl Braid {
...
+        }
+        parents
+    }
+
+    fn get_children(&self) -> HashMap<BeadHash, HashSet<BeadHash>> {
```
TypiCal25, line 112: As mentioned above, I think it would be good if we split this function for allowing a function that gets us the children for a specific bead. Then we can make use of this function in this function. I feel that could be an improvement in the way we write these functions, but do want to know what you think as well.

braidpool-primitives/src/braid.rs
```diff
@@ -86,6 +93,69 @@ impl Braid {
...
+                    .insert(beadhash.clone());
+            }
+            children
+                .entry(beadhash.clone())
+                .or_insert_with(HashSet::new);
+        }
+        children
+    }
+
+    fn get_geneses(&self) -> HashSet<BeadHash> {
```
TypiCal25, line 128: I think we can rename this to get_genesis_beads. Also a type alias for the return type helps a lot similar to the type aliases written above. I know this might seem like redundant code, but it will help us keep the code super clear. Since these won't be pub types, as well as these are private functions, so we don't need to worry about keeping these types pub(external) either.

TypiCal25: Also, in my PR on adding the genesis bead check, we can make use of that function here, and simply have a single for loop in this function making it an inline function for faster processing speed as well. I think we can refactor this code, once we have both our PRs merged. Or equivalently, we could get that PR merged, and you could rebase on top of that, and replace the function definition equivalently.

image

TypiCal25 (dismissed):

braidpool-primitives/src/braid.rs
```diff
@@ -44,6 +44,13 @@ impl Braid {
         }
     }
 
+    pub fn load_beads_in_memory(&mut self, beads: Vec<Bead>) {
```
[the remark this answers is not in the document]
Aritra8438: I'm fine wth both. At this point, this should not make any difference wrt the working of the `v1`.

braidpool-primitives/src/braid.rs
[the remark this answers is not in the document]
Aritra8438: > Also another suggestion I have for this, is that we break this code into a nice set of cleaner functions

I thought of this and also planned to implement (inspired by your `BeadHash error enum`). If you want to, please go ahead.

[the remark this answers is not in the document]
Aritra8438: > I think renaming `type ParentBeadHash = BeadHash`, `type ChildrenBeadHash = BeadHash` and `type Children = HashSet<ChildrenBeadHash>` would make our code a whole lot cleaner.

@mcelrath even used `Relatives`. That's a good idea too.

braidpool-primitives/src/braid.rs
Aritra8438: Fine, I needed to add this as a `placeholder`.

braidpool-primitives/src/braid.rs
Aritra8438: Sure, fine with me.

braidpool-primitives/src/braid.rs
TypiCal25: yep yep, I agree with this. I'm saying that we still take in the set of beads itself, but we call a function for each bead rather than do all the logic in this one function. It makes for cleaner code. What do you think? Btw, the logic and everything is perfect. I meant only a small refactor for code reability.

braidpool-primitives/src/braid.rs
TypiCal25: oh alright perfect.

braidpool-primitives/src/braid.rs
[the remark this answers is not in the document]
TypiCal25: Yup I don't think it makes too much of a difference. I also do feel like this is a better choice since it clearly shows that the bead vector is being consumed by the function.

braidpool-primitives/src/braid.rs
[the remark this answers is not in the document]
TypiCal25: Alright. We can use `Relatives` if that makes it clearer. For now, I'll write it through child/parent, but ill also define the public type alias of Relatives.

Aritra8438 (commented):
Looks fine! I'l ask @Sansh2356 as it's my PR and can't approve.

braidpool-primitives/src/bead/mod.rs
```diff
@@ -64,6 +69,62 @@ impl Bead {
         // TODO: Implement this function.
         unimplemented!()
     }
+
+    #[inline]
+    pub fn is_parent_of(&self, child_bead_hash: BeadHash) -> bool {
```
mcelrath, line 74: How do you intend to use `is_*` functions? I personally dislike a bunch of helper functions like this, I'd rather write something like (pythonic) `if bead in otherbead.parents { ... }`. Unless this presents a problem with borrowing in Rust...

braidpool-primitives/src/bead/mod.rs
```diff
@@ -64,6 +69,62 @@ impl Bead {
...
+            if braid.beads.contains(parent) == false {
+                return true;
+            }
+        }
+
+        false
+    }
+
+    #[inline]
+    pub fn get_parents(&self) -> Parents {
```
mcelrath, line 109: Why would we need a `get_parents()` function instead of just using `bead.parents` (assuming bead is a public struct).

braidpool-primitives/src/braid/mod.rs
```diff
@@ -85,6 +98,38 @@ impl Braid {
...
+
+        match self.loaded_beads_in_memory.get(&bead_hash) {
+            Some(bead) => Ok(bead),
+            None => Err(BeadLoadError::BeadPruned),
+        }
+    }
+}
+
+impl Braid {
+    fn update_parents_of_bead(&self, child_bead: &Bead) {
```
mcelrath, line 116: The parents of a bead are static, and committed to when the bead is created. Setting the parents should be part of a constructor. We should never be adding parents.

If e.g. we're mining on a bead having parents A and B, and a new bead C comes in, we should delete the previous bead we were mining on and instantiate a new Bead object with A,B,C. Adding the parent C changes the coinbase tx which commits to parents in an OP_RETURN (committed metadata), which changes the merkle root, and changes the bead entirely. There's no need to keep the old (A,B parent) bead around, it will never be mined and can be discarded.

braidpool-primitives/src/braid/mod.rs
```diff
@@ -112,66 +157,23 @@ impl Braid {
         cohorts
```
mcelrath, line 157: Call braid.cohorts() here with the parents HashMap (and cached children hashmap)

braidpool-primitives/src/utils/mod.rs
```diff
@@ -1,8 +1,40 @@
...
-// Type Definitions
+// External Type Aliases
 pub type BeadHash = BlockHash;
 pub type Byte = u8;
 pub type Bytes = Vec<Byte>;
+
+// Internal Type Aliases
+pub(crate) type ParentBeadHash = BeadHash;
+pub(crate) type ChildrenBeadHash = BeadHash;
+pub(crate) type RelativeBeadHash = BeadHash;
```
mcelrath, line 17: Why have three separate types for this? Why not just use BeadHash?

mcelrath (changes requested):

braidpool-primitives/src/bead/mod.rs
```diff
@@ -1,14 +1,16 @@
...
 use bitcoin::{BlockHeader, CompactTarget, Transaction};
 
 // Custom Imports
-use crate::utils::BeadHash;
+use crate::braid::Braid;
 use crate::utils::bitcoin::MerklePathProof;
+use crate::utils::{BeadHash, BeadLoadError, Children, Parents};
 
 // Type Aliases
 type TransactionWithMerklePath = (Transaction, MerklePathProof);
```
mcelrath, line 16: How and when do we plan to use a Merkle Path proof here?

Txs are included in a bead (and there should be a small Merkle tree with the ~2-5 txs so they can be easily committed to).

Merkle proofs are really only used when you need to communicate compactly with another party who doesn't have the same database that we do. As long as we construct our database consistently, we don't need to prove to *ourselves* that a tx was included -- we can just do a db query to verify it, and if we want we can chase the container bead, its merkle tree, etc to verify if we think there's a reason our db is wrong.

I can see ways we might use merkle proofs in the future (settlement/payouts), but I'm not sure how to grok Merkle paths in the basic Braid/Bead structure...

braidpool-primitives/src/braid/mod.rs
```diff
@@ -38,20 +41,29 @@ impl Braid {
...
             tips: genesis_beads.clone(),
-            cohorts: vec![Cohort(genesis_beads)],
+            cohorts: vec![Cohort(genesis_beads.clone())],
             orphan_beads: Vec::new(),
             loaded_beads_in_memory: HashMap::new(),
+            genesis_beads: genesis_beads,
+        }
+    }
+
+    pub fn load_beads_in_memory(&mut self, beads: Vec<Bead>) {
```
ABD-AZE, line 51: I think a sanity check should be done prior to inserting in order to avoid inserting duplicates.

braidpool-primitives/src/braid/mod.rs
```diff
@@ -68,14 +80,15 @@ impl Braid {
...
             self.orphan_beads.push(bead);
             return AddBeadStatus::ParentsNotYetReceived;
         }
 
         self.beads.insert(bead.bead_hash);
         self.remove_parent_beads_from_tips(&bead);
         self.tips.insert(bead.bead_hash);
+        self.update_parents_of_bead(&bead);
 
         self.cohorts = self.calculate_cohorts();
```
ABD-AZE, line 93: Can we somehow use the previous cohort state to update the current one, rather than recomputing the entire thing. This would be a major optimisation

braidpool-primitives/src/braid/mod.rs
```diff
@@ -38,20 +41,29 @@ impl Braid {
...
         Braid {
             beads: previous_dag_braid.tips.clone(),
-            tips: previous_dag_braid.tips,
+            tips: previous_dag_braid.tips.clone(),
             cohorts,
             orphan_beads: Vec::new(),
             loaded_beads_in_memory: HashMap::new(),
+            genesis_beads: previous_dag_braid.tips,
         }
     }
```
ABD-AZE, line 69: We need to improve the overall error handling of the add_bead function. The remove_parent_beads_from_tips must return a Result, in case of an error we must return it (after conversion to the AddBeadStatus enum), same goes for update_parents_of_bead and update_orphan_bead_set. All the functions doing any sort of update deletion must return the status to ensure proper error handling.

PR #114, opened 2025-04-06
fix: correct MerklePathProof hashing order for left and right leaf nodes

TypiCal25:
This commit fixes the `MerklePathProof` implementation to ensure the correct hashing order is used when calculating the Merkle root for both left and right leaf nodes in the Merkle binary tree.

1. **Fix in `get_merkle_hashing_order`**:
   - Correctly handles the hashing order for both left and right leaf nodes.
   - For right leaf nodes (`is_right_leaf == true`):
     - The sibling hash from the `merkle_path` is added first.
     - The `transaction_hash` is added second. - The remaining hashes in the `merkle_path` are appended starting from index `1`.
   - For left leaf nodes (`is_right_leaf == false`): - The `transaction_hash` is added first. - The remaining hashes in the `merkle_path` are appended starting from index `0`.

2. **Improved Clarity**:
   - Introduced the `index_for_starting_the_copy` variable to clearly define where to start iterating over the `merkle_path` for appending additional hashes.
   - Ensures consistent and predictable behavior for both left and right leaf cases.

3. **Integration with `calculate_corresponding_merkle_root`**:
   - The `get_merkle_hashing_order` function is used to generate the correct hashing order.
   - The concatenated hashes are double-hashed using `Sha256d` to compute the Merkle root.
   - The resulting Merkle root is converted into a `TxMerkleNode` for compatibility with Bitcoin's data structures.

- Fixes potential issues with incorrect Merkle root calculations when the transaction is a right leaf.
- Ensures compatibility with Bitcoin's Merkle tree structure and hashing rules.
- Improves code readability and maintainability by clearly separating the logic for left and right leaf nodes.

### DIFF
--- a/braidpool-primitives/src/utils/bitcoin.rs
+++ b/braidpool-primitives/src/utils/bitcoin.rs
@@ -10,23 +10,44 @@ type MerkleRoot = TxMerkleNode;
 
 pub struct MerklePathProof {
     pub transaction_hash: Txid,
+    pub is_right_leaf: bool,
     pub merkle_path: Vec<TxMerkleNode>,
 }
 
 impl MerklePathProof {
+    // All public functions go here!
     pub fn calculate_corresponding_merkle_root(&self) -> MerkleRoot {
-        let mut current_hash = self.transaction_hash.as_byte_array().clone();
-        for merkle_node in &self.merkle_path {
-            let merkle_node_as_bytes = merkle_node.as_byte_array();
-
-            let mut concatenated_hashes = Vec::new();
-            concatenated_hashes.extend_from_slice(&current_hash);
-            concatenated_hashes.extend_from_slice(merkle_node_as_bytes);
-
-            current_hash = Sha256d::hash(&concatenated_hashes).as_byte_array().clone();
+        // Handle left and right leaf situations!
+        let hashing_order = self.get_merkle_hashing_order();
+        let mut concatenated_hashes: Vec<u8> = Vec::new();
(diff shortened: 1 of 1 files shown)


Sansh2356 (approved):

TypiCal25:
Thankss!

braidpool-primitives/src/utils/bitcoin.rs
```diff
@@ -10,23 +10,44 @@ type MerkleRoot = TxMerkleNode;
 
 pub struct MerklePathProof {
     pub transaction_hash: Txid,
+    pub is_right_leaf: bool,
     pub merkle_path: Vec<TxMerkleNode>,
```
Aritra8438, line 14:
```suggestion
    pub is_right_leaf: bool,
    pub is_alone_leaf: bool,
    pub merkle_path: Vec<TxMerkleNode>,
```

I'm not sure if that will ever happen for `coinbase` or `payout` but we still need to handle one more case.
When there's an odd number of leaves, we need to handle the case for the last leaf where we do `HASH256(A + A)`

TypiCal25: Yeppp, but this would be already handled because the merkle path, will consist of that copy of the transaction, since the binary tree is always complete.

Aritra8438: Okk! Fine.

TypiCal25: perfect :)

PR #121, opened 2025-04-11
patch: add dag visualisation and fix vite build

its-me-abhishek:
- Added DAG Visualization via API into the UI
- Updated the dependencies to use Vite build successfully

https://github.com/user-attachments/assets/9f58535e-0a3f-4cd9-8bac-89e85de61973


priyashuu (dismissed):
LGTM 🚀🫡

priyashuu:
1. **Tooltip Positioning**:
   - Problem: Fixed position hides node info
   - Fix: Update to follow mouse:
     ```typescript
     .style('left', `${event.pageX + 10}px`)
     .style('top', `${event.pageY + 10}px`)
     ```

2. **Zoom State**:
   - Problem: `defaultZoom` may desync
   - Fix: Use D3 transform as single source:
     ```typescript
     const [zoomTransform, setZoomTransform] = useState(d3.zoomIdentity);
     ```

priyashuu:
Otherwise, it looks good to me 🫡🚀

its-me-abhishek:
the tooltip was dynamic earlier but it seems to mess everything up once there's a lot of data, by going out of the view, hence i have pinned it to the bottom-left, and will be visible only when you hover over it

priyashuu:
> the tooltip was dynamic earlier but it seems to mess everything up once there's a lot of data, by going out of the view, hence i have pinned it to the bottom-left, and will be visible only when you hover over it

Can we break some code into multiple files? It will be easier to manipulate changes in the future since we are using React. Otherwise, it looks good to me

rasta-dave:
> > the tooltip was dynamic earlier but it seems to mess everything up once there's a lot of data, by going out of the view, hence i have pinned it to the bottom-left, and will be visible only when you hover over it
> 
> Can we break some code into multiple files? It will be easier to manipulate changes in the future since we are using React. Otherwise, it looks good to me

Yeah we can totally modularize it further as we go

priyashuu:
> > > the tooltip was dynamic earlier but it seems to mess everything up once there's a lot of data, by going out of the view, hence i have pinned it to the bottom-left, and will be visible only when you hover over it
> > 
> > 
> > Can we break some code into multiple files? It will be easier to manipulate changes in the future since we are using React. Otherwise, it looks good to me
> 
> Yeah we can totally modularize it further as we go

True

PR #119, opened 2025-04-09
Update structs to reflect PPLNS and committed mempool

mcelrath:

PR #123, opened 2025-04-12
Updating the definitions for structs 

Sansh2356:
- Updating the `struct` definitions .
- Adding `Serialization` trait and unit test for checking serialization of bead .
- Removing the struct fields according to newer `bradipool_spec.md` PR #119 .

### DIFF
--- a/braidpool-primitives/src/bead/mod.rs
+++ b/braidpool-primitives/src/bead/mod.rs
@@ -1,60 +1,53 @@
+#![allow(unused)]
 // Standard Imports
+use ::serde::{Deserialize, Serialize};
 use std::collections::HashSet;
+use std::net::SocketAddr;
 
 // Bitcoin primitives
+use bitcoin::Address;
 use bitcoin::absolute::Time;
+use bitcoin::ecdsa::Signature;
+use bitcoin::p2p::Address as P2P_Address;
+use bitcoin::secp256k1::PublicKey;
 use bitcoin::transaction::TransactionExt;
-use bitcoin::{BlockHeader, CompactTarget, Transaction};
-
+use bitcoin::{BlockHeader, Transaction};
 // Custom Imports
 use crate::utils::BeadHash;
-use crate::utils::bitcoin::MerklePathProof;
+#[derive(Clone, Debug, Serialize)]
 
-// Type Aliases
-type TransactionWithMerklePath = (Transaction, MerklePathProof);
+pub struct CommittedMetadata {
+    // Committed Braidpool Metadata,
+    pub parents: HashSet<(BeadHash, Time)>,
+    pub transactions: Vec<Transaction>,
+    pub payout_address: Address,
+    //timestamp when the bead was created
+    pub observed_time_at_node: Time,
+    pub comm_pub_key: PublicKey,
+    pub miner_ip: SocketAddr,
+    pub extra_nonce: i32,
+}
+#[derive(Clone, Debug, Serialize)]
+
+pub struct UnCommittedMetadata {
+    //Uncomitted Metadata
+    //timestamp when the bead was broadcasted
+    pub broadcast_timestamp: Time,
+    pub signature: Signature,
+}
+#[derive(Clone, Debug, Serialize)]
 
 pub struct Bead {
     pub block_header: BlockHeader,
-    pub bead_hash: BeadHash,
-    pub coinbase_transaction: TransactionWithMerklePath,
-    pub payout_update_transaction: TransactionWithMerklePath,
-
-    // Committed Braidpool Metadata,
-    pub lesser_difficulty_target: CompactTarget,
-    pub parents: HashSet<(BeadHash, Time)>,
-    pub transactions: Vec<Transaction>,
-
-    pub observed_time_at_node: Time,
+    pub committed_metadata: CommittedMetadata,
+    pub uncommitted_metadata: UnCommittedMetadata,
 }
 
 impl Bead {
-    // All public definitions go in here!
-    #[inline]
-    pub fn is_transaction_included_in_block(
-        &self,
-        transaction_with_proof: &TransactionWithMerklePath,
-    ) -> bool {
-        transaction_with_proof
-            .1
-            .calculate_corresponding_merkle_root()
-            == self.block_header.merkle_root
-    }
-
     pub fn is_valid_bead(&self) -> bool {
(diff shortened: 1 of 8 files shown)


braidpool-primitives/src/bead/mod.rs
```diff
@@ -1,60 +1,53 @@
...
 
+pub struct Bead {
+    pub block_header: BlockHeader,
+    pub committed_metadata: CommittedMetadata,
+    pub uncommitted_metadata: UnCommittedMetadata,
+}
+
+impl Bead {
     pub fn is_valid_bead(&self) -> bool {
-        // Check whether the transactions are included in the block!
```
zaidmstrr, line 44: This function is always returning `true` without any condition. Is this a part of defined behaviour? And where are these removed checks handled in the codebase?

Sansh2356: This function will be either removed or updated it is not the purpose of this current PR @zaidmstrr  , another PR i am working upon as of now that will add and update these changes .

Aritra8438 (commented):
Minor suggestions from my side.

braidpool-primitives/src/bead/tests.rs
```diff
@@ -1 +1,77 @@
...
+    };
+    let test_bytes = [0u8; 32];
+    let test_bead = Bead {
+        block_header: BlockHeader {
+            version: BlockVersion::TWO,
+            prev_blockhash: BlockHash::from_byte_array(test_bytes),
+            bits: CompactTarget::from_consensus(32),
+            nonce: 1,
+            time: BlockTime::from_u32(8328429),
+            merkle_root: TxMerkleNode::from_byte_array(test_bytes),
```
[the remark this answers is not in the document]
zaidmstrr: Nice catch, and yes, we should follow the same ordering.

braidpool-primitives/src/bead/tests.rs
```diff
@@ -1 +1,77 @@
...
+        },
+        committed_metadata: test_committed_metadata,
+        uncommitted_metadata: test_uncommittedmetadata,
+    };
+    let deserialized_str = serde_json::to_string(&test_bead).unwrap();
+    assert_eq!(
+        deserialized_str,
+        r#"{"block_header":{"version":2,"prev_blockhash":"0000000000000000000000000000000000000000000000000000000000000000","merkle_root":"0000000000000000000000000000000000000000000000000000000000000000","time":8328429,"bits":32,"nonce":1},"committed_metadata":{"parents":[],"transactions":[],"payout_address":"32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf","observed_time_at_node":1653195600,"comm_pub_key":"02b98a7fb8cc007048625b6446ad49a1b3a722df8c1ca975b87160023e14d19097","miner_ip":"127.0.0.1:8080","extra_nonce":12},"uncommitted_metadata":{"broadcast_timestamp":1653195600,"signature":{"signature":"3046022100839c1fbc5304de944f697c9f4b1d01d1faeba32d751c0f7acb21ac8a0f436a72022100e89bd46bb3a5a62adc679f659b7ce876d83ee297c7a5587b2011c4fcc72eab45","sighash_type":"SIGHASH_ALL"}}}"#
+    );
```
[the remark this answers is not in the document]
zaidmstrr: I think we are not following the last two things `Transaction Count` `Transactions` in our bead data structure.

braidpool-primitives/src/bead/mod.rs
```diff
@@ -1,60 +1,53 @@
...
-    pub block_header: BlockHeader,
-    pub bead_hash: BeadHash,
-    pub coinbase_transaction: TransactionWithMerklePath,
-    pub payout_update_transaction: TransactionWithMerklePath,
+#[derive(Clone, Debug, Serialize)]
 
+pub struct CommittedMetadata {
     // Committed Braidpool Metadata,
-    pub lesser_difficulty_target: CompactTarget,
     pub parents: HashSet<(BeadHash, Time)>,
```
mcelrath, line 21: When we serialize a HashSet, can the order be depended upon?

I'm thinking maybe we move the timestamps into UncommittedMetadata -- my question is, if the serialization of `parents` is effectively a list, can we put a timestamp list in `UncommittedMetadata` such that the list of timestamps is in the same order as this `parents` serialization?

braidpool-primitives/src/bead/mod.rs
```diff
@@ -1,60 +1,53 @@
...
     pub transactions: Vec<Transaction>,
-
+    pub payout_address: Address,
+    //timestamp when the bead was created
     pub observed_time_at_node: Time,
+    pub comm_pub_key: PublicKey,
+    pub miner_ip: SocketAddr,
+    pub extra_nonce: i32,
```
mcelrath, line 28: We need to remove this, somehow. `extra_nonce` is required in order to compute the coinbase transaction, but then gets committed to twice (once in the coinbase and once in CommittedMetadata). However the [Mining Protocol](https://stratumprotocol.org/specification/05-Mining-Protocol/) makes this field available to mining devices in the coinbase for maniuplation. Mining devices will roll the extranonce, but not the commitment to the same in CommittedMetadata. So unfortunately putting `extra_nonce` here will break some mining devices.

What would the consequences be of putting this in `UncommittedMetadata`?

braidpool-primitives/src/bead/tests.rs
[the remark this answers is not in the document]
mcelrath: Yes indeed this needs to be exactly the same. I missed Tx count because I wrote the doc from memory...

mcelrath (changes requested):
I think we should remove the `MerklePathProof` stuff in `braidpool-primitives/src/utils/bitcoin/mod.rs` until such time as we need it.

braidpool-primitives/src/bead/mod.rs
Sansh2356: As discussed with @braidpool/sob-2025  and @mcelrath , `extra_nonce` is being moved to the uncommitted metadata struct , in order to avoid the issue of committing `extra_nonce` twice , that will eventually lead to making the valid `extra_nonce` being returned by the downstream node i.e. the miner invalid thus leading to different merkel node from the required one , causing invalidation of a valid weak-share.

zaidmstrr:
Concept ACK

mcelrath (approved):

PR #129, opened 2025-04-15
feat: Refactor Mined Shares Explorer with animated components and dat…

priyashuu:
…a visualization

- Added AnimatedStatCard component for displaying statistics with animations.
- Created DashboardHeader for navigation and header display.
- Implemented DateRangePicker for selecting time ranges.
- Developed FilterBar for filtering data based on time ranges.
- Introduced useChartData and useComparisonData hooks for data fetching and manipulation.
- Built MinedSharesExplorer component to aggregate all functionalities.
- Added MinerCompare component for comparing miner performance.
- Created MinerRow component for displaying individual miner details.
- Developed TransactionList component for listing transactions related to miners.
- Defined constants and types for miners, transactions, and time ranges.


priyashuu:
https://github.com/user-attachments/assets/585f048d-942a-4fbc-a9e0-c0da064ae2e1

rasta-dave:
It looks really good Priya! Nice structure on your code

rasta-dave (approved):

priyashuu:
image

mcelrath (changes requested):
On the Miner Stats/Miners tab:
1. Hovering over the transaction hash makes a popup that is unreadable due to transparency ("Transaction hash for block #1234"). Actually the transaction hash is getting redrawn on top of the popup. I would remove this popup actually.
2. I would rename this tab "Beads". The list of "miner1", "miner2" below that should be "Beads" and beads are identified by hashes.
3. Add a row: "Parents" which lists the hashes of parents, and links to them in the table.
4. Add header row item: "Difficulty"
5. The rows below each bead (currently called "miner1", "miner2"...) are transactions. Transactions don't have their own timestamp that's different from the bead's timestamp, and the column on the right "Transactions" doesn't make sense because transactions don't have transactions. Beads have transactions (but that's the header row).
6. The entry in the "Transactions" column should only appear in the header row and should be exactly the number of transactions below it. (So 3 for the "miner1" row, 2 for the "miner2" row, etc.
7. If you want to add more information about the transactions in columns, some interesting ideas would be: size in vbytes, fee paid in BTC, fee paid in sats/vbyte, number of inputs, number of outputs. Take a look at https://mempool.space for some ideas. (and that code is open source -- Check out https://mempool.space for some ideas)
8. The mempool.space code is open source, and uses the same license as Braidpool. Feel free to steal anything from that project! https://github.com/mempool/mempool/

In fact we might consider contributing Braidpool monitoring to the mempool.space project...

The "Blocks" tab is pretty straightforward and can be enhanced by code from mempool.space.

The "Trends" tab should probably show hashrate (need axis labels). I suggest breaking this into three tabs: "Hashrate" and "Latency" and "Transactions".
1. I suggest enhancing the simulator to report hashrate (lambda parameter) and a (latency parameter).
2. Both hashrate and latency should be real-time graphs, updated every time a bead comes in (every ~500ms). It's also jagged data, so consider adding a moving average and Bollinger bands, or other statistical indicators. We'll need to get this data from the simulator first.
3. "Transactions" should show statistics from the transactions in the Committed Mempool. We can re-use mempool.space code here, and we'll have to get transactions from `cmempoold` rather than `bitcoind`.

The "Blocks" tab is good. Steal some code from mempool.space here for more details.

I'm not sure what to do with the "Compare" tab. I don't think people need to or will want to "compare" miners. What was your thought on what this is supposed to be conveying?

priyashuu:
I added the Miner Compare section because choosing the right mining setup often requires comparing different miners based on their specifications, power consumption, efficiency, and profitability.

This feature is meant to help users—especially beginners—make informed decisions by providing a clear side-by-side comparison. It saves them the effort of researching and comparing details manually from different sources

priyashuu:
Maybe we can change the purpose by turning it into something more interactive, like comparing mining strategies, different cryptocurrencies, or even comparing data from different mining setups..

priyashuu:
https://github.com/user-attachments/assets/41b3e627-caa0-45e3-b48a-876bf66b8eb2

mcelrath:
I was mostly expecting the latest changes to display Beads here instead of miner1, miner2 as indicated in this comment: https://github.com/braidpool/braidpool/pull/129#pullrequestreview-2773191970

Am I looking in the wrong place? I can't see any visible changes since the last time I reviewed...

priyashuu:
The latest changes with Beads are actually in this PR: #134

PR #134, opened 2025-04-17
feat: Refactor Beads  explorer animated components and data

priyashuu:
image
fixes :https://github.com/braidpool/braidpool/pull/129


rasta-dave (changes requested):
The MinedSharesExplorer component looks great, but I'd recommend adding error handling for the case when network data fails to load. Given that this dashboard will be displaying real-time mining data, adding a simple error state with a retry option could improve user experience when connection issues occur. This would prevent users from seeing empty or partially loaded views

All in all it looks good!

PR #148, opened 2025-04-24
feat: Dashboard frontend revamp

its-me-abhishek:
- Increased node size for better visibility and aesthetics.
- Arrowheads added to links to indicate directionality.
- Highlighted the Hwpath
- Nodes are now plotted on both sides of the x-axis for better distribution.
- Updated to support the latest simulator_api
- Animated the links for the latest and oldest cohort changes


mcelrath (dismissed):
It looks like you're closing the WS connection after each bead. You should keep it open. Here's the simulator_api.py log:
```
INFO:__main__:Connection closed: 1005 - 
INFO:websockets.server:connection open
INFO:__main__:Client connected: ('127.0.0.1', 45616)
INFO:__main__:Connection closed: 1005 - 
INFO:websockets.server:connection open
INFO:__main__:Client connected: ('127.0.0.1', 45626)
INFO:__main__:Connection closed: 1005 - 
INFO:websockets.server:connection open
```

The arrow animation is being applied to the leftmost bead in the view also, when its parent moves out of view.

Large cohorts expand right from the point the first bead in the cohort was added. This can extend out of view for large cohorts.

I'm still seeing cohorts change color sometimes, especially with larger ones?

We still have lines overlapping beads in larger cohorts.

You need to take into account the actual width of the canvas. I usually use a browser window that's taller than it is wide, and in the default view the newest bead is off screen. Calculate an appropriate x position for the latest bead that is on-screen for the browser.

I just saw an example where one of the parent beads was more than 10 cohorts back. The API should have given you everything as one giant cohort -- this requires a significant redraw of the entire DAG.

You're numbering the beads sequentially from the leftmost bead, which among other things causes the numeric identifier of beads to change as they move left. Instead, use the actual bead hash returned by the API. You'll have to find a somewhat-readable way to display this. Probably drop the leading zeros and display only 4 hex digits, maybe 2 on top and 2 on bottom?

I'm going to merge this anyway so we can iterate on the design. I don't really like the pulsing arrows lol.

Can you fix the "connection closed" issue and the viewport issue to keep everything on the screen, and we can iterate on the rest?

Issue #147, opened 2025-04-23
Roadmap


Sansh2356:
The scope of this issue tracks the progress for the overall feature of `cmempoold` : 

- [x] Implementing the structs for the native data structures of braidpool such as `braid` and `bead` etc . (delivered in the PR #123 ) PR #227 . 
- [x] Adding `test_framework` for test coverage from start itself  (delivered in the PR #162 ) , rpc implementation PR #224 . 
- [x] `Serializability` and `deserializability` of the braid data structures (delivered in the PR #162 ).
- [ ] `XOnlyPubkey` or the `Taproot key` analysis based on quadratic residue .   
- [x] Updating the consensus functions for existing in the [`Braidpool example.ipynb`](https://github.com/braidpool/braidpool/blob/main/docs/braidpool_spec.md) . 
- [x] Integration of the primitive braidpool `p2p` functionalities 
  - [x] `intialized-braidpool` including the initialization of the `braid-node` , `cmempoold` proxy and `bitcoind` PR #214 #206 #202 #201 . 
  - [x ] `cmempoold` using the bitcoin mempool using our custom mempool PR #213 , #191 PR #229 . 
  - [x] `tokio-tracing` for better tracing #187 . 
- [x] Bead construction #160 .
- [x] Bead propagation and validation via [`SV2 mining protocol`.](https://github.com/stratum-mining/sv2-spec/blob/main/05-Mining-Protocol.md), though there was slight change as we opted for stratum v1 initially afterwards accompanying the stratum v2 primitive protocols for integration to be done along exisiting braid downstream channel such as the `mining_protocol` and `binary_sv2`  for reduced latency and secured channel.
- [x] Conversion of valid bead to mainnet block via PR #237  .
- [x] Submission of suitable values of `version`,`extranonce` and `nonce` by the ASICs for bead construction via PR #237 .
- [x] Bead accumulation for the formulation of a `Braid`. Addressed and merged in PR #207 .
- [ ] `PPLNS` or `UHPO` payment settlement etc along with `Eltoo` and `Schnorr` signature supporting multi-sig payout authentication mechanism. 
- [x] Complete `IPC` functionality and additional `cmempoold` functionality to be merged via PR #229 .
- [x] `Stratum` implementation for downstream connection completing the communication loop between the IPC and the downstream via PR #237 .
- [x] Bead propagation over to peers via PR #257 .  
- [ ] DB implementation for disk persistence see issue #262  PR #265 .
[...]

PR #160, opened 2025-04-30
Updating the consensus functions according to new struct definitions 

Sansh2356:
Continuation of PR #109.
Addresses #147.

### DIFF
--- a/braidpool-primitives/src/braid/mod.rs
+++ b/braidpool-primitives/src/braid/mod.rs
@@ -3,16 +3,16 @@ use ::serde::Serialize;
 use std::collections::{HashMap, HashSet};
 
 // Bitcoin Imports
-use bitcoin::CompactTarget;
+use bitcoin::{BlockHash, CompactTarget};
 
 // Custom Imports
 use crate::bead::Bead;
-use crate::utils::BeadHash;
+use crate::utils::{BeadHash, BeadLoadError};
 
 // Type Definitions
 #[derive(Clone, Debug, Serialize)]
 
-struct Cohort(HashSet<BeadHash>);
+pub(crate) struct Cohort(HashSet<BeadHash>);
 
 pub enum AddBeadStatus {
     DagAlreadyContainsBead,
@@ -26,14 +26,17 @@ type NumberOfBeadsUnorphaned = usize;
 #[derive(Clone, Debug, Serialize)]
 
 pub struct Braid {
-    beads: HashSet<BeadHash>,
-    tips: HashSet<BeadHash>,
-    cohorts: Vec<Cohort>,
+    pub(crate) beads: HashSet<BeadHash>,
+    pub(crate) tips: HashSet<BeadHash>,
+    pub(crate) cohorts: Vec<Cohort>,
 
-    orphan_beads: Vec<Bead>,
+    pub(crate) orphan_beads: Vec<Bead>,
 
     // Database related functions!
-    loaded_beads_in_memory: HashMap<BeadHash, Bead>,
+    pub(crate) loaded_beads_in_memory: HashMap<BeadHash, Bead>,
+
+    // Optimizations (not part of specification!)
+    pub(crate) genesis_beads: HashSet<BeadHash>,
 }
 
 impl Braid {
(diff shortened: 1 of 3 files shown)


braidpool-primitives/src/braid/mod.rs
```diff
@@ -26,14 +26,17 @@ type NumberOfBeadsUnorphaned = usize;
 #[derive(Clone, Debug, Serialize)]
 
 pub struct Braid {
-    beads: HashSet<BeadHash>,
-    tips: HashSet<BeadHash>,
-    cohorts: Vec<Cohort>,
+    pub(crate) beads: HashSet<BeadHash>,
```
mcelrath, line 29: Add a timestamp field here. (When this bead arrived according to the node's clock) Sender timestamps are in the bead. We will use this to characterize our latency measurements.

braidpool-primitives/src/braid/mod.rs
```diff
@@ -42,20 +45,41 @@ impl Braid {
...
             tips: genesis_beads.clone(),
-            cohorts: vec![Cohort(genesis_beads)],
+            cohorts: vec![Cohort(genesis_beads.clone())],
             orphan_beads: Vec::new(),
             loaded_beads_in_memory: HashMap::new(),
+            genesis_beads: genesis_beads,
+        }
+    }
+
+    pub fn load_beads_in_memory(&mut self, beads: Vec<Bead>) {
```
mcelrath, line 55: I think we should put Braid loading/saving in the Node instead of the Braid.

Let's try to keep the Bead and Braid data structures as "simple" data structures, with the Node doing most of the manipulation of them.

braidpool-primitives/src/braid/mod.rs
```diff
@@ -42,20 +45,41 @@ impl Braid {
...
+                }
+            };
+            let mut serialized_bytes = [0u8; 32];
+            hex::decode_to_slice(serialized_curr_bead_str, &mut serialized_bytes).unwrap();
+            let bead_hash = BlockHash::from_byte_array(serialized_bytes);
+            self.loaded_beads_in_memory.insert(bead_hash, bead);
         }
     }
 
     pub fn generate_from_previous_dag(previous_dag_braid: Braid) -> Self {
```
mcelrath, line 74: This is effectively a copy constructor...

How do you see this being used? I'm not sure we'll ever want to copy a braid...

braidpool-primitives/src/braid/mod.rs
```diff
@@ -68,14 +92,15 @@ impl Braid {
         //     return AddBeadStatus::DagAlreadyContainsBead;
         // }
 
-        if self.is_bead_orphaned(&bead) {
+        if bead.is_orphaned(self) {
```
mcelrath, line 95: Move to Node class.

braidpool-primitives/src/braid/mod.rs
```diff
@@ -85,6 +110,120 @@ impl Braid {
 }
 
 impl Braid {
+    // All pub(crate) functions go here!
+    #[inline]
+    pub(crate) fn load_bead_from_hash(&self, bead_hash: BeadHash) -> Result<&Bead, BeadLoadError> {
```
mcelrath, line 115: This should be done by Bead serialization, and the db work should go into the Node.

braidpool-primitives/src/braid/mod.rs
```diff
@@ -85,6 +110,120 @@ impl Braid {
...
+
+        match self.loaded_beads_in_memory.get(&bead_hash) {
+            Some(bead) => Ok(bead),
+            None => Err(BeadLoadError::BeadPruned),
+        }
+    }
+}
+
+impl Braid {
+    fn update_parents_of_bead(&self, child_bead: &Bead) {
```
mcelrath, line 128: We never need to update the parents of a bead.

When a Bead is created, its parents should be set in its constructor. If a new bead comes in, the Node should destroy the old Bead we were mining on, and create a new one (it will have different parents and possibly different transactions). So we basically have to recreate a whole new Bead from scratch instead of modifying an existing Bead.

braidpool-primitives/src/braid/mod.rs
```diff
@@ -85,6 +110,120 @@ impl Braid {
...
+        }
+        // Remove all cohorts after the found index
+        if let Some(idx) = remove_after {
+            self.cohorts.truncate(idx);
+        } else {
+            self.cohorts.clear();
+        }
+
+        // Construct a sub-braid from dangling and compute any new cohorts
+        // Here, we just create a new cohort with dangling beads
```
mcelrath, line 219: We need to call the `cohorts(dangling)` algorithm here. See node/src/braid.rs.

braidpool-primitives/src/utils/mod.rs
```diff
@@ -1,8 +1,41 @@
...
 
-// Type Definitions
+// External Type Aliases
 pub type BeadHash = BlockHash;
 pub type Byte = u8;
 pub type Bytes = Vec<Byte>;
+
+// Internal Type Aliases
+pub(crate) type ParentBeadHash = BeadHash;
+pub(crate) type ChildrenBeadHash = BeadHash;
```
mcelrath, line 16: Why two types here? Why not just use BeadHash?

PR #161, opened 2025-05-02
Beads explorer Modifications 

priyashuu:
Replaced "Difficulty" with "Work": Updated the terminology for clarity and accuracy.

Added Work Units: Displayed work in standardized units (e.g., TH, EH , Gh), making the data more informative.

Introduced Reward Field: Included miner rewards in each transaction entry or block summary.


mcelrath (changes requested):
image

See the above screenshot:
1. The "Analytics" button is off screen in portrait mode, and "Export Data" isn't visible.
2. The timestamp field is overlapping with the bead hash field
3. The size field is overlapping with the hash field.
4. The reward for a single bead will be roughly 1/1000 of that of a block. Can you put some random numbers here in the range of 1-3 mBTC? 

Looks good! I'm planning on showing this in my talk on Thursday. So if there are any other cool things to do before then it would be appreciated!

PR #162, opened 2025-05-03
Implement `encodable` trait for `Bead` structure

Aritra8438:
This PR add implementations of `Encodable` trait for custom structs, e.g., `bead`.

### DIFF
--- a/braidpool-primitives/src/bead/mod.rs
+++ b/braidpool-primitives/src/bead/mod.rs
@@ -1,4 +1,5 @@
 #![allow(unused)]
+use bitcoin::io::{self, Write};
 // Standard Imports
 use ::serde::{Deserialize, Serialize};
 use std::collections::HashSet;
@@ -7,6 +8,9 @@ use std::net::SocketAddr;
 // Bitcoin primitives
 use bitcoin::Address;
 use bitcoin::absolute::Time;
+use bitcoin::consensus::encode::Decodable;
+use bitcoin::consensus::encode::Encodable;
+use bitcoin::consensus::encode::serialize;
 use bitcoin::ecdsa::Signature;
 use bitcoin::p2p::Address as P2P_Address;
 use bitcoin::secp256k1::PublicKey;
@@ -14,19 +18,86 @@ use bitcoin::transaction::TransactionExt;
 use bitcoin::{BlockHeader, Transaction};
 // Custom Imports
 use crate::utils::BeadHash;
-#[derive(Clone, Debug, Serialize)]
+
+/// Wrapper type for HashSet<BeadHash> to allow custom trait implementations.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub struct BeadHashSet(pub HashSet<BeadHash>);
+
+impl Encodable for BeadHashSet {
+    fn consensus_encode<W: Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
+        let mut len = 0;
+        len += (self.0.len() as u64).consensus_encode(w)?;
+        for item in self.0.iter() {
+            len += item.consensus_encode(w)?;
+        }
+        Ok(len)
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub struct TimeHashSet(pub HashSet<Time>);
+
+impl Encodable for TimeHashSet {
+    fn consensus_encode<W: Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
+        let mut len = 0;
+        // Encode the length for deterministic encoding
+        len += (self.0.len() as u64).consensus_encode(w)?;
+        for time in &self.0 {
+            len += time.to_consensus_u32().consensus_encode(w)?;
+        }
+        Ok(len)
+    }
+}
+
+/// Wrapper type for SocketAddr to allow custom trait implementations.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct SerializableSocketAddr(pub SocketAddr);
+
+impl Encodable for SerializableSocketAddr {
+    fn consensus_encode<W: Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
+        let (address, port) = match self.0 {
+            SocketAddr::V4(addr) => (addr.ip().to_ipv6_mapped().segments(), addr.port()),
+            SocketAddr::V6(addr) => (addr.ip().segments(), addr.port()),
+        };
+        let mut len = 0;
+        len += address.consensus_encode(w)?;
+        len += port.consensus_encode(w)?;
+        Ok(len)
+    }
+}
+
+#[derive(Clone, Debug)]
 
 pub struct CommittedMetadata {
     // Committed Braidpool Metadata,
     pub transaction_cnt: u32,
     pub transactions: Vec<Transaction>,
-    pub parents: HashSet<BeadHash>,
-    pub payout_address: Address,
+    pub parents: BeadHashSet,
+    pub payout_address: P2P_Address,
     //timestamp when the bead was created
     pub observed_time_at_node: Time,
     pub comm_pub_key: PublicKey,
-    pub miner_ip: SocketAddr,
+    pub miner_ip: SerializableSocketAddr,
 }
+
+impl Encodable for CommittedMetadata {
+    fn consensus_encode<W: Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
+        let mut len = 0;
+        len += self.transaction_cnt.consensus_encode(w)?;
+        len += self.transactions.consensus_encode(w)?;
+        len += self.parents.consensus_encode(w)?;
+        len += self.payout_address.consensus_encode(w)?;
+        len += self
+            .observed_time_at_node
+            .to_consensus_u32()
+            .consensus_encode(w)?;
+        let pubkey_bytes = self.comm_pub_key.serialize();
+        len += pubkey_bytes.consensus_encode(w)?;
+        len += self.miner_ip.consensus_encode(w)?;
+        Ok(len)
+    }
+}
+
 #[derive(Clone, Debug, Serialize)]
 
 pub struct UnCommittedMetadata {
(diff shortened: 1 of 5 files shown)


braidpool-primitives/src/bead/mod.rs
```diff
@@ -1,34 +1,110 @@
...
+    }
+}
+
+impl Decodable for TimeVec {
+    fn consensus_decode<R: BufRead + ?Sized>(r: &mut R) -> Result<Self, Error> {
+        let len = u64::consensus_decode(r)?;
+        let mut vec = Vec::with_capacity(len as usize);
+        for _ in 0..len {
+            let time_u32 = u32::consensus_decode(r)?;
+            let time = Time::from_consensus(time_u32).unwrap();
```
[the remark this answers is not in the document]
Aritra8438: That thows the following error:
```
`?` couldn't convert the error to `bitcoin::consensus::Error`
the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
```

braidpool-primitives/src/bead/mod.rs
```diff
@@ -119,16 +248,9 @@ impl Bead {
...
-        let serialized_bead_str = serde_json::to_string(self).unwrap_or_else(|error| {
-            panic!(
-                "An error occurred while serializing the bead string: {:?}",
-                error
-            )
-        });
+        let serialized_bead = serialize(self);
         let mut serialized_bytes = [0u8; 32];
-        hex::decode_to_slice(serialized_bead_str, &mut serialized_bytes).unwrap();
+        hex::decode_to_slice(serialized_bead, &mut serialized_bytes).unwrap();
```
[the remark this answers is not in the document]
Aritra8438: @Sansh2356, after you review this, I'll modify this function `get_bead_hash`.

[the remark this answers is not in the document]
Sansh2356: Looks fine according to what we had discussed , just the order of the `timestamps` must be same to that of the order of `parents` received which will be same in this case also . On serialized portion and deserialize portion everything looks fine, I will add a `test_framework` and unit-tests for traits as mentioned by @mcelrath .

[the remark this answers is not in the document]
Aritra8438: > the order of the `timestamps` must be same to that of the order of `parents` received which will be same in this case also

Yes, we need to make sure of that.
Once you are done with the `test_framework`, we can start working on the `P2P` side.

braidpool-primitives/src/bead/mod.rs
```diff
@@ -1,34 +1,110 @@
...
+
+#[derive(Clone, Debug)]
 
 pub struct CommittedMetadata {
     // Committed Braidpool Metadata,
     pub transaction_cnt: u32,
     pub transactions: Vec<Transaction>,
-    pub parents: HashSet<BeadHash>,
-    pub payout_address: Address,
+    pub parents: Vec<BeadHash>,
```
mcelrath, line 60: `parents` is logically a set, and set operations are used in `cohorts()`. I suggest storing this as a set, but serializing it as a vector with a defined order (for instance, lexically sort the parents by their BeadHash).

braidpool-primitives/src/bead/mod.rs
```diff
@@ -1,34 +1,110 @@
...
+        }
+        Ok(TimeVec(vec))
+    }
+}
+
+#[derive(Clone, Debug)]
 
 pub struct CommittedMetadata {
     // Committed Braidpool Metadata,
     pub transaction_cnt: u32,
```
mcelrath, line 58: Is a count needed here? Doesn't the serialization of `Vec<Transaction>` include a count?

braidpool-primitives/src/bead/mod.rs
```diff
@@ -1,34 +1,110 @@
...
     pub transactions: Vec<Transaction>,
-    pub parents: HashSet<BeadHash>,
-    pub payout_address: Address,
+    pub parents: Vec<BeadHash>,
+    pub payout_address: P2P_Address,
     //timestamp when the bead was created
     pub observed_time_at_node: Time,
```
mcelrath, line 63: Can we rename this to `start_timestamp`? It's not an "observation". It's the timestamp put in by the node before the bead is handed off to mining hardware. We will use this and `UnCommittedMetadata.broadcast_timestamp` to get an estimate of how long it took the miner to solve this block. (this quantity is not used in consensus, but can be used for statistical analysis of the behavior of miners).

braidpool-primitives/src/bead/mod.rs
```diff
@@ -37,16 +113,69 @@ pub struct UnCommittedMetadata {
     pub extra_nonce: i32,
```
mcelrath, line 113: Please investigate the use of `extranonce1` and `extranonce2` in the StratumV2 protocol, and how we will use them, and whether we should have two of them.

braidpool-primitives/src/bead/mod.rs
```diff
@@ -1,34 +1,110 @@
...
     // Committed Braidpool Metadata,
     pub transaction_cnt: u32,
     pub transactions: Vec<Transaction>,
-    pub parents: HashSet<BeadHash>,
-    pub payout_address: Address,
+    pub parents: Vec<BeadHash>,
+    pub payout_address: P2P_Address,
     //timestamp when the bead was created
     pub observed_time_at_node: Time,
     pub comm_pub_key: PublicKey,
```
mcelrath, line 64: We need to specify what kind of pubkey this is and how it is encoded. I don't want to use arbitrary pubkeys allowed by bitcoin which have a large variation in their DER encoding and ECDSA/Schnorr usage. I'd like to use taproot-style pubkeys encoded in 64 bits using the quadratic residue algorithm. The point of this pubkey is that it will be used for signatures on the `UnCommittedMetadata` and for ECIES encryption between mining nodes.

Investigate the compatibility of Taproot keys with ECIES.

braidpool-primitives/src/utils/mod.rs
```diff
@@ -14,8 +14,8 @@ pub type Bytes = Vec<Byte>;
 // Internal Type Aliases
 pub(crate) type ParentBeadHash = BeadHash;
```
mcelrath, line 15: Why do we want 3 different type definitions here? A BeadHash is a BeadHash is a BeadHash... A single BeadHash can be both a parent and a child.

mcelrath (changes requested):
This looks really great guys, we're headed in the right direction.

Let's address these comments, add some tests, and get the Rust unit tests passing.

Merging this will be a big step into doing everything else. :-D

braidpool-primitives/src/bead/mod.rs
Sansh2356: ACK .

braidpool-primitives/src/bead/mod.rs
Sansh2356: I have done this earlier also will share with you on discord .

braidpool-primitives/src/utils/mod.rs
Sansh2356: ACK and done .

braidpool-primitives/src/bead/mod.rs
Sansh2356: ACK, To be dealt with in separate issue at #163 .

braidpool-primitives/src/braid/mod.rs
```diff
@@ -23,19 +17,14 @@ pub enum AddBeadStatus {
 
 // Type Aliases
 type NumberOfBeadsUnorphaned = usize;
```
mcelrath, line 19: What's this for? Why not just do Braid.orphan_beads.len()?

braidpool-primitives/src/braid/mod.rs
```diff
@@ -23,19 +17,14 @@ pub enum AddBeadStatus {
...
-#[derive(Clone, Debug, Serialize)]
+#[derive(Clone, Debug)]
 
 pub struct Braid {
     pub(crate) beads: HashSet<BeadHash>,
```
mcelrath, line 23: Why have `HashSet<BeadHash>` instead of `HashSet<Bead>`?

Essentially we've created a layer of indirection where we have to get the BeadHash and then look it up in `loaded_beads_in_memory`.

What's going to go wrong in Rust if we just use the Bead object directly?

This would mean that the same objects are in both `beads`, `tips`, `cohorts`, `orphan_beads` and `genesis_beads`. In C++ I'd just use references or pointers but Rust is annoying about ownership...

Can you make an argument for using BeadHash here instead of the Bead object itself?

Can we create some kind of lazy loading, where if we try to access a bead that is not in memory it will trigger a db call to load them?

braidpool-primitives/src/bead/mod.rs
```diff
@@ -102,34 +226,11 @@ impl Bead {
...
-                "An error occurred while serializing the bead string: {:?}",
-                error
-            )
-        });
-        let mut serialized_bytes = [0u8; 32];
-        hex::decode_to_slice(serialized_bead_str, &mut serialized_bytes).unwrap();
-        BlockHash::from_byte_array(serialized_bytes)
+        self.block_header.block_hash()
     }
 }
```
mcelrath, line 236: This would be a lot easier if we were carrying around Bead objects instead of BeadHash objects.

braidpool-primitives/src/bead/mod.rs
```diff
@@ -1,52 +1,182 @@
...
+        }
+        Ok(TimeVec(vec))
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
 
 pub struct CommittedMetadata {
-    // Committed Braidpool Metadata,
     pub transaction_cnt: u32,
```
mcelrath, line 52: I don't like transaction_cnt here. We have a length in `transactions` and there's no guarantee it's going to be the same as `transaction_cnt`. We should be serializing the length when we serialize `transactions` as part of serializing the vector (as it's done with `TimeVec`), not storing the length twice.

braidpool-primitives/src/utils/test_utils.rs
```diff
@@ -0,0 +1,204 @@
...
+        pub fn new() -> Self {
+            Self {
+                extra_nonce: 0,
+                broadcast_timestamp: None,
+                signature: None,
+                parent_bead_timestamps: None,
+            }
+        }
+
+        pub fn extra_nonce(mut self, nonce: i32) -> Self {
```
mcelrath, line 40: Why are there all these getter/setter functions? It's very verbose and useless code.

Can we do this without a bunch of getter/setters?

mcelrath (changes requested):

PR #164, opened 2025-05-12
docs: migrated from MkDocs to Vocs.dev for better UX and readability

punithbm:
I was going through the docs and found the content is very good but the readability and navigation a bit difficult with the current setup. So I migrated the documentation from MkDocs to [Vocs.dev](https://vocs.dev/), an open-source, lightweight tool optimized for better UX, clean layout, and faster access.

here's quick preview on how it looks, 

https://github.com/user-attachments/assets/c7b4e89b-7ff0-4d46-9a5e-514eaed54540


Chandher05:
This looks great! Easy to just go through it

PR #180, opened 2025-05-24
Libp2p bootstrap

ABD-AZE:
- changed enable the use of quic transport in libp2p to connect to a remote peer given its multiaddress. 
- The multiaddress of the seed node is to be passed to the cli arg --addnode
- The multiaddress on which the node would listen is to be passed to the cli arg --multiaddr
- The cli arg **bind** is of no use for running the code 
- Eg command for running the first node: 
```
cargo run -- --bind=localhost:8989 --bitcoin=0.0.0.0 --rpcport=8332 --rpcuser=xxxx --rpcpass=yyyy --zmqhashblockport=28332 --multiaddr=/ip4/127.0.0.1/udp/8000/quic-v1
```
- Eg. command for running the second node using the first node as the seed
```
cargo run -- --bind=localhost:9899 --addnode=/ip4/127.0.0.1/udp/8000/quic-v1 --bitcoin=0.0.0.0 --rpcport=8332 --rpcuser=xxxx --rpcpass=yyyy --zmqhashblockport=28332 --multiaddr=/ip4/127.0.0.1/udp/9000/quic-v1 
```


ABD-AZE:
> Will go through it soon. Just a quick review @ABD-AZE. Is the change in the ipynb file intended? If not, please undo the changes.

Removed the changes, the changes were due to running the examples file which has random output. My bad that I pushed it in hurry. The only change which remains is some version change which can be ignored or changed later if required.

mcelrath (changes requested):
As it stands, I can't figure out how to run this. It requires the `--multiaddr` argument to be passed, which I presume is the seed node, but this means it won't start without a seed node, which means it's impossible to start the seed node.

Please rename `--multiaddr` to `seed` and set its default to be `seed.braidpool.net` (I'll add this to DNS).

Then I think `--addnode` should take either a PeerID or a multiaddr.

ABD-AZE:
1) --multiaddr takes the multiaddress to which the node binds and listens on, it is the libp2p replacement for --bind.

2) --addnode takes the multiaddress of the seed node

3) As for running the code, I hope bitcoind is running with the specified configuration. If not it can be run with the following command: 
```
bitcoind -rpcport=8332 -rpcuser=xxxx -rpcpassword=yyyy -zmqpubhashblock=tcp://localhost:28332
```
After running bitcoind, the previous two commands in the PR description can be run in the given order.

PR #181, opened 2025-05-26
Ping behaviour

### LINKED ISSUE #176: Integrate libp2p: ping message (remove flexbuffers)
This should replace:
```
node/src/protocol.rs
node/src/connection.rs
```
with something more appropriate to libp2p. use the libp2p request_response framework to send a 'ping' to a peer.

Record the latency of the ping message in an appropriate node-wide data structure. (for later use in peer selection) This is probably a `HashMap<PeerID, time>`.

The only other thing in `node/src/protocol.rs` is a heartbeat message. Determine whether a heartbeat message is needed with libp2p. I assume this is handled under the hood?

ABD-AZE:
- Resolves #176 
- The connected nodes exchange a ping-pong message after every 15 seconds.
- The connection timeout limit between the nodes is set to u64::MAX to avoid the connection from getting closed.
```
.with_swarm_config(|cfg| cfg.with_idle_connection_timeout(Duration::from_secs(u64::MAX)))
```
- This PR build on top of #180
- For running the nodes refer #180


mcelrath (changes requested):
Same comments as #180 - I can't run it.

Please modify selftest.sh to spin up 2 (or more) nodes with one as a boot node and connect them to each other and send a ping.

ABD-AZE:
> Same comments as #180 - I can't run it.
> 
> Please modify selftest.sh to spin up 2 (or more) nodes with one as a boot node and connect them to each other and send a ping.
> 
> Since this includes the code of #180, can we close that one and merge #181 instead?

Modified selftest.sh to spin up 2 nodes. But both nodes use the same terminal for their outputs and both processes are killed immediately after exchanging one ping, thus it is better if you could use two different terminals to manually spin up the nodes. 
NOTE: Ensure that bitcoind is running with the desired configuration.

ABD-AZE:
> Same comments as #180 - I can't run it.
> 
> Please modify selftest.sh to spin up 2 (or more) nodes with one as a boot node and connect them to each other and send a ping.
> 
> Since this includes the code of #180, can we close that one and merge #181 instead?

I separated both PRs for ease of review, if you want we can directly merge this one.

PR #191, opened 2025-05-29
feature: Add Bitcoin IPC schema and implementation

zaidmstrr:
This PR implements a Rust-based schema and IPC implementation used by Cap'n Proto to talk with `bitcoin-node` through Unix socket.


mcelrath (approved):
Most of these `*.capnp` files are coming from bitcoin. Can we add bitcoin as a dependency, submodule, or use an environment variable to the path of the bitcoin source tree to obtain these files? I don't think we should check them into braidpool, so that we can absorb upstream changes.

Because cmempoold is consensus code, we need to ensure that the bitcoin dependency is a particular git commit, or particular versions with known-same mempool behavior.

Merging this because it's going into the branch `ipc/main` which I assume is a WIP. This needs to be rebased off dev, there are some upstream changes (braidpool-primitives is gone). Let's come back to the above questions when we merge into dev.

PR #204, opened 2025-06-04
persistent private key

### LINKED ISSUE #183: Use persistent private keys to derive the PeerId via public key
For the demonstration purposes in #180 , a fresh PeerID is generated for every run of the program. What we want is a deterministic PeerID for every system. This can be achieved by searching for ~/.braidpool/keystore for the private key (similar to how IPFS does it) and use it further. If the file doesn't contain (or doesn't exist) a valid private key, a new private key is generated and stored in the previously mentioned location.

ABD-AZE:
Resolves #183 
- The keys get stored in the file **keystore** inside the data directory (default data directory is "~/.braidpool/")
- Keys are stored in protobuf encoded format
- In order to run multiple instance of node on the same machine for testing in future comment the lines 64-79 and add the line:
``` 
 let keypair: Keypair = libp2p::identity::Keypair::generate_ed25519(); 
```

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -2,11 +2,15 @@ use clap::Parser;
 use futures::StreamExt;
 use libp2p::{
     core::multiaddr::Multiaddr,
-    identify, ping,
+    identify,
+    identity::Keypair,
+    ping,
     swarm::{NetworkBehaviour, SwarmEvent},
 };
 use std::error::Error;
 use std::fs;
+use std::os::unix::fs::PermissionsExt;
+use std::path::Path;
 use std::time::Duration;
 use tokio::sync::mpsc;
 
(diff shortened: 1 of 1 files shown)


node/src/main.rs
```diff
@@ -55,7 +58,27 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
     tokio::spawn(block_template::consumer(block_template_rx));
 
-    let mut swarm = libp2p::SwarmBuilder::with_new_identity()
+    let datadir_path = Path::new(&*datadir);
+    let keystore_path = datadir_path.join("keystore");
+
+    let keypair = match fs::read(&keystore_path) {
+        Ok(keypair) => {
+            log::info!("Loading existing keypair from keystore...");
+            libp2p::identity::Keypair::from_protobuf_encoding(&keypair).map_err(|e| {
```
mcelrath, line 83: Why protobuf? There are a number of common key storage encodings including PEM etc but I've never heard of anyone using protobuf for this. 

Will we ever have a need to store more than one key?

ABD-AZE: It is the default encoding scheme in libp2p::Identity

mcelrath (changes requested):
Overall good and works.

Can you set the mode of the `keystore` file to 0400 (only read permission for the user on this file):
```
-r-------- 1 mcelrath mcelrath 68 Jun  6 10:15 /home/mcelrath/.braidpool/keystore
```

node/src/main.rs
```diff
@@ -55,7 +59,35 @@ async fn main() -> Result<(), Box<dyn Error>> {
     tokio::spawn(zmq::zmq_hashblock_listener(zmq_url, rpc, block_template_tx));
     tokio::spawn(block_template::consumer(block_template_rx));
 
-    let mut swarm = libp2p::SwarmBuilder::with_new_identity()
+    let datadir_path = Path::new(&*datadir);
+    let keystore_path = datadir_path.join("keystore");
+    #[cfg(unix)]
+    {
+        if keystore_path.exists() {
+            let mut perms = fs::metadata(&keystore_path)?.permissions();
```
mcelrath: This is running *before* the key is created.

You might want to *check* the permissions here and issue a warning if they are wrong.

But is line should be in the Err handler below where the key is created and written.

mcelrath (dismissed):

mcelrath (approved):

Issue #193, opened 2025-05-30
Bitcoin health/stats tab


mcelrath:
Use Bitcoin's `getpeerinfo` and `getblockchaininfo` RPC calls (and any others you deem interesting) to get info about the Bitcoin node's health. Create a tab to display this. Run `bitcoin-qt` to get some more ideas for things that might be interesting to display.

priyashuu:
I would like to assign this to myself . Here’s my plan for the **Node Health** tab:

###  Data Sources (RPC Calls):

* `getblockchaininfo`: to get chain height, verification progress, and warnings.
* `getpeerinfo`: to list peers, ping times, client versions, and connection types.
* `getnetworkinfo`: to get node version, number of connections, and relay fee.
* `getnettotals`: to display bandwidth usage (total bytes sent/received).
* `getmempoolinfo`: for mempool transaction count and size.
* and many more in the future ig 

###  Features I’ll Include:

* Node sync status (e.g., current block vs. verification progress).
* Number of connected peers (inbound vs. outbound).
* Table of top peers with ping times, client versions, and ban score.
* Network totals: bytes sent/received.
* Mempool stats: size and transaction count.
* Node warnings or alerts (if any).
* Auto-refresh every 30–60 seconds for real-time health monitoring.
* and many more in the future ig 

Let me know if you'd like any additional metrics or ideas included!

PR #205, opened 2025-06-05
Node health

priyashuu:
feat: Add Node Health Dashboard with network, mempool, and bandwidth statistics
- Updated package.json to include new dependencies: cors, dotenv, express, and js-cookie.
- Implemented NodeHealth component with tabs for Blockchain, Peers, Network, Mempool, and Bandwidth.
- Created BandwidthPanel, MempoolPanel, NetworkPanel, and Peers components to display respective data.
- Added API calls to fetch blockchain, peer, network, mempool, and net totals information.
- Introduced RawJsonViewer for displaying raw data in JSON format.
- Set up a proxy server to handle API requests to the Braidpool backend.
- issue: #193 
image
image
image
image
image


its-me-abhishek:
- One should not place any other file except for tests in the `__tests__` directory for example:  `dashboard/src/components/NodeHealth/__tests__/server.ts`, etc. Move all these files to the parent directory
- You seem to have created some API functions in  `dashboard/src/components/NodeHealth/__tests__/` but we already have an API residing in `dashboard/api`. It would be beneficial to extend that instead of creating another API as this creates more dependencies in the dashboard than required and convoluted code

priyashuu:
> * One should not place any other file except for tests in the `__tests__` directory for example:  `dashboard/src/components/NodeHealth/__tests__/server.ts`, etc. Move all these files to the parent directory
> * You seem to have created some API functions in  `dashboard/src/components/NodeHealth/__tests__/` but we already have an API residing in `dashboard/api`. It would be beneficial to extend that instead of creating another API as this creates more dependencies in the dashboard than required and convoluted code
u made the api directory for the bitcoin/stats am i right ?

priyashuu:
you made the api directory for the bitcoin/stats am i right ?

priyashuu:
If yes, then I’ll rearrange the file structure according to yours after the bitcoin/stats merge :)

PR #207, opened 2025-06-09
Replace `beadhash` with `beads` and `indices`

Aritra8438:
This PR will do the `rust side` implementation of consensus functionalities inherit to `braidpool` the `braid functions` in the rust.


Aritra8438:
This is a draft, I need to modify the `extend` function.

node/src/braid/mod.rs
```diff
@@ -119,31 +166,618 @@ impl Braid {
...
+        return genesis_bead_indices;
+    }
+
+    fn tips(&mut self, parents: HashMap<usize, HashSet<usize>>) -> HashSet<usize> {
+        let mut bead_indices_mapping: HashMap<usize, usize> = HashMap::new();
+        let mut tips_indices: HashSet<usize> = HashSet::new();
+        for parent_bead_idx in parents {
+            bead_indices_mapping.insert(parent_bead_idx.0, 0);
+        }
+        let current_beads = self.beads.clone();
```
mcelrath, line 197: This is going to be a very expensive operation for a large braid. Can't we iterate over self.beads without cloning it?

node/src/braid/tests.rs
```diff
@@ -1 +1,323 @@
...
+    let test_bead = TestBeadBuilder::new()
+        .block_header(test_block_header)
+        .committed_metadata(committed_metadata)
+        .uncommitted_metadata(uncommitted_metadata)
+        .build();
+    test_bead
+}
+
+#[test]
+pub fn test_extend_functionality() {
```
mcelrath, line 199: This is very nice.

mcelrath (dismissed):
Since this is a complete port of the braid code into a Braid class, we will need to also import all the tests in braid_functions/tests.rs, to ensure that the port is correct.

I had been thinking that you were going to leave braid_functions/mod.rs as it is and import and use it. The reason to do that is that the tests are complex and we can leave them as they are, just passing around HashMaps instead of Bead objects.

I think the direction you've gone here is fine though, but in the process, I'd like to import all tests in braid_functions/tests.rs so that they are applied to this new Braid class, and delete braid_functions/ if the port is successful and complete.

Your new test of extend() is very good, and I would like to see it tested against all the braids in tests/braids/*.json.

node/src/braid/mod.rs
Sansh2356: This can be taken care of @mcelrath ACK .

node/src/braid/mod.rs
```diff
@@ -119,31 +166,647 @@ impl Braid {
...
+                    .insert(current_bead_idx);
+            }
+        }
+        return bead_children_mapping;
     }
 
-    fn generate_tip_cohorts(&self) -> Vec<Cohort> {
-        let mut cohorts = Vec::new();
-        let tips = self.tips.clone();
+    fn generation(
```
zaidmstrr, line 243: It is good to write what the function does at the top of every possible function (in comments).

node/src/braid/mod.rs
```diff
@@ -119,31 +166,647 @@ impl Braid {
...
+            if let Some(child_beads) = bead_children_mapping.get(&bead) {
+                for child_bead in child_beads {
+                    children_set.insert(*child_bead);
+                }
+            }
+        }
+        return children_set;
+    }
+
+    fn all_ancestors<'a>(
```
zaidmstrr: Same applies for here

node/src/braid/mod.rs
```diff
@@ -119,31 +166,647 @@ impl Braid {
...
+                            dequeue.push_back((*parent, false));
+                        }
+                    }
+                }
+            }
+        }
+
+        return ancestors;
+    }
+    fn get_all_ancestors<'a>(
```
zaidmstrr, line 336: This function name is confusing, as a similar function name already exists `all_ancestors`.

zaidmstrr (commented):
I haven't checked for the program logic. But these are some small recommendations for the program's clarity and maintainability. I have seen many functions that don't specify their use cases in the comments above, so try to write comments for those functions, as this is the consensus code; this needs to be clear to anyone who looks at this code. You can skip those functions which represent their working in their name, but if those functions are also consensus critical, then it's good to write the comments above. 

Related commits can also be squashed together to make commit history clean.

node/src/braid/mod.rs
Sansh2356: @zaidmstrr agreed will do so .

Sansh2356:
> I haven't checked for the program logic. But these are some small recommendations for the program's clarity and maintainability. I have seen many functions that don't specify their use cases in the comments above, so try to write comments for those functions, as this is the consensus code; this needs to be clear to anyone who looks at this code. You can skip those functions which represent their working in their name, but if those functions are also consensus critical, then it's good to write the comments above.
> 
> Related commits can also be squashed together to make commit history clean.

Yes will do .

mcelrath (approved):
Tested, works.

PR #192, opened 2025-05-30
Add Bitcoin stats and refactor the codebase

its-me-abhishek:
- Added a view for bitcoin statistics graphs
- Added the API to fetch the stats via WebSockets
- Replaced MUI with Tailwind in some major components 
- Refactored the codebase, migrating props to Types.ts for every component

PR #208, opened 2025-06-11
Beads update

priyashuu:
BeadsTab tested and updated:

- Removed DateRangePicker  RewardMilestones , filterbar

- Cleaned up state and removed unused refs in MinedSharesExplorer

- Moved utility functions to a single utils.ts file

- Updated constants and types to include fee-related fields

- All parts tested except transactions

AdvancedChart and HashrateTab:

- Improved hover effects and dynamic data display
- Fixed class names and removed unnecessary comments
- Adjusted layout styles


dashboard/src/components/BeadsTab/BlockVisulisation/Block.tsx
```diff
@@ -10,7 +10,7 @@ export default function EnhancedBlocksTab({
 }) {
   const [isLoaded, setIsLoaded] = useState(false);
   const [animateBlocks, setAnimateBlocks] = useState(false);
-  const [viewMode, setViewMode] = useState('list');
```
its-me-abhishek, line 13: why did you remove this setter? only getter just means a constant value, either shall add the setter back, or just turn it into a constant

dashboard/src/components/BeadsTab/Reward/RewardsSection.tsx
```diff
@@ -1,418 +1,216 @@
...
-          '0.01 BTC Earned',
-        ],
-      };
-
-      setRewardData(data);
-      setIsLoaded(true);
-    }, 1500);
-
-    return () => clearTimeout(timer);
+    let firstLoad = true;
```
its-me-abhishek: firstLoad will always be true on first render because useEffect(() => {}, []) only runs once on mount. So firstLoad adds no value and causes confusion it will always execute that if (firstLoad) block once and then disappear

dashboard/src/components/BeadsTab/AdvancedChart.tsx
```diff
@@ -1,13 +1,16 @@
...
+  Tooltip,
+  ResponsiveContainer,
+  Area,
+} from 'recharts';
+import { Maximize2, RefreshCw, Download } from 'lucide-react';
+import { useState } from 'react';
+import { motion } from 'framer-motion';
 type TrendType = 'up' | 'down' | 'neutral';
 
 interface ChartDataPoint {
```
its-me-abhishek: formattedDate? is unused, and these interfaces should be moved to Types.ts of the directory

dashboard/src/components/BeadsTab/AnimatedStatCard.tsx
```diff
@@ -9,148 +7,57 @@ interface AnimatedStatCardProps {
...
-  };
+  const isPositive = change.startsWith('+');
+  const trendColor = isPositive
+    ? 'text-emerald-400'
+    : change.startsWith('-')
+      ? 'text-red-400'
+      : 'text-gray-400';
 
   return (
     <motion.div
```
its-me-abhishek: please make sure that the styles are migrated to tailwind if possible

dashboard/src/components/BeadsTab/TransactionList.tsx
```diff
@@ -2,8 +2,7 @@ import { useState } from 'react';
 import { motion } from 'framer-motion';
 import { Activity } from 'lucide-react';
 import type { Transaction } from './lib/types';
-
```
its-me-abhishek, line 5: This TransactionList is the same as `transation.tsx` I think you forgot to remove that one

dashboard/src/components/BeadsTab/api/nodeApi.ts
```diff
@@ -0,0 +1,29 @@
+import axios from 'axios';
```
its-me-abhishek: either both of the API files should be TS or JS, seems to be a typo

dashboard/src/components/BeadsTab/MinedSharesExplorer.tsx
```diff
@@ -1,153 +1,100 @@
...
-  // Handle parent bead click
   const handleParentClick = (parentHash: string) => {
-    const bead = BEADS.find((b) => b.name === parentHash);
-    if (bead) toggleBead(bead.id);
-  };
-
-  const handleRetry = () => {
-    setError(null);
-    refetch();
+    // Handle parent click logic here
```
its-me-abhishek: seems to be unimplemented, please look into it

dashboard/src/components/BeadsTab/api/nodeApi.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const BASE_URL = 'http://localhost:3001/api/node';
+
+const fetchNodeData = async (method: string) => {
```
its-me-abhishek: this method is unused, can be removed

its-me-abhishek (changes requested):
Please check these out, will report more if I encounter anything else

its-me-abhishek:
- You seem to have removed some of the animations and colors, but the colors, look and feel on different views are still quite different, please look into it, as tests shall be written only when the UI has migrated to the final form (similar to the Dashboard, and other Pages)
 
For example:
image
image
both look quite different

- The BeadTabs/api folder does not contain a package.json, so I am unable to run the API for rewards fetching, did you test this yet?

- Vite Build still throws multiple errors: 
```
npm run build

> dashboard@0.1.0 build
> tsc && vite build

src/components/BeadsTab/Trends/TransactionsTab.tsx:48:11 - error TS2322: Type '{ title: string; value: string; change: string; icon: Element; color: string; delay: number; }' is not assignable to type 'IntrinsicAttributes & AnimatedStatCardProps'.
  Property 'color' does not exist on type 'IntrinsicAttributes & AnimatedStatCardProps'.

48           color="emerald"
             ~~~~~

src/components/BeadsTab/Trends/TransactionsTab.tsx:56:11 - error TS2322: Type '{ title: string; value: string; change: string; icon: Element; color: string; delay: number; }' is not assignable to type 'IntrinsicAttributes & AnimatedStatCardProps'.
  Property 'color' does not exist on type 'IntrinsicAttributes & AnimatedStatCardProps'.

56           color="purple"
             ~~~~~

src/components/BeadsTab/Trends/TransactionsTab.tsx:64:11 - error TS2322: Type '{ title: string; value: string; change: string; icon: Element; color: string; delay: number; }' is not assignable to type 'IntrinsicAttributes & AnimatedStatCardProps'.
  Property 'color' does not exist on type 'IntrinsicAttributes & AnimatedStatCardProps'.

64           color="blue"
             ~~~~~

src/components/ShareDetails/ShareDetailsDemo.tsx:3:26 - error TS2307: Cannot find module './ShareDetails/ShareDetails' or its corresponding type declarations.

3 import ShareDetails from './ShareDetails/ShareDetails';
                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/components/ShareDetails/ShareDetailsDemo.tsx:4:39 - error TS2307: Cannot find module '../data/mockBeads' or its corresponding type declarations.

4 import { mockBeads, printDebug } from '../data/mockBeads';
                                        ~~~~~~~~~~~~~~~~~~~

src/components/ShareDetails/ShareDetailsDemo.tsx:41:22 - error TS18046: 'bead' is of type 'unknown'.

41       ([_, bead]) => bead.beadHash === beadHash
                        ~~~~

src/components/common/ActionIconButton.tsx:28:9 - error TS2769: No overload matches this call.
  The last overload gave the following error.
    Object literal may only specify known properties, and 'className' does not exist in type 'Partial<unknown> & Attributes'.

28         className: 'w-[18px] h-[18px] text-inherit',
           ~~~~~~~~~

  node_modules/@types/react/index.d.ts:523:14
    523     function cloneElement<P>(
                     ~~~~~~~~~~~~
    The last overload is declared here.

src/utils/MinerCompare.tsx:7:23 - error TS2307: Cannot find module './lib/constants' or its corresponding type declarations.

7 import { BEADS } from './lib/constants';
                        ~~~~~~~~~~~~~~~~~

src/utils/MinerCompare.tsx:8:35 - error TS2307: Cannot find module './Hooks/useComparisonData' or its corresponding type declarations.

8 import { useComparisonData } from './Hooks/useComparisonData';
                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/utils/MinerCompare.tsx:9:27 - error TS2307: Cannot find module './AdvancedChart' or its corresponding type declarations.

9 import AdvancedChart from './AdvancedChart';
                            ~~~~~~~~~~~~~~~~~

src/utils/MinerCompare.tsx:39:25 - error TS7006: Parameter 'bead' implicitly has an 'any' type.

39             {BEADS.map((bead) => (
                           ~~~~

src/utils/MinerCompare.tsx:64:29 - error TS7006: Parameter 'm' implicitly has an 'any' type.

64               ? BEADS.find((m) => m.id === selectedMiner)?.name || 'Comparison'
                               ~

tailwind.config.ts:4:3 - error TS2322: Type '["class"]' is not assignable to type 'DarkModeStrategy | undefined'.
  Type '["class"]' is not assignable to type '["class", string]'.
    Source has 1 element(s) but target requires 2.

4   darkMode: ['class'],
    ~~~~~~~~

  node_modules/tailwindcss/dist/types-B254mqw1.d.mts:83:5
    83     darkMode?: DarkModeStrategy;
           ~~~~~~~~
    The expected type comes from property 'darkMode' which is declared here on type 'UserConfig'


Found 13 errors in 5 files.
```

priyashuu:
hey @its-me-abhishek  
All the changes you requested have been resolved:

 No more build errors as of now
image


 No extra dependencies added , just upgraded the existing ones

UI has been synchronized as per your suggestions
image


Let me know if there’s anything else you’d like me to update :)

its-me-abhishek:
The build works fine, but it seems you had a misunderstanding regarding the UI updates. The screenshot I shared earlier was just an example, the other lists/tabs/glassy views still need to be addressed. Please review the entire dashboard and fix the issues. I have a non exhaustive list of those:
- Fonts are not consistent across components.
- The sticky header extends too far.
- The bottom border of the sticky header is overly extended.
- In the Beads tab, the highlight color doesn't go away after selecting a bead.
- The background color of this page differs from other pages.
- Row heights are not optimal.
- There are still animated components in the dashboard, these should ideally use CSS-only animations to reduce build size.
- EDIT: PR #209 already moved the API in the dashboard directory, outside the components, so please align with that structure.

https://github.com/user-attachments/assets/ca2e152a-c0bf-449b-99a7-8bc67475e751

priyashuu:
https://github.com/user-attachments/assets/4cc762d8-0348-47fc-91b0-40e9adf378ee

dashboard/tsconfig.tsbuildinfo
```diff
@@ -0,0 +1 @@
...0,1],[45,1],[46,1],[47,1],[48,1],[8,1],[54,1],[51,1],[52,1],[53,1],[55,1],[9,1],[56,1],[57,1],[58,1],[60,1],[59,1],[61,1],[62,1],[10,1],[63,1],[64,1],[65,1],[11,1],[66,1],[67,1],[68,1],[69,1],[70,1],[1,1],[71,1],[72,1],[12,1],[76,1],[74,1],[78,1],[73,1],[77,1],[75,1],[978,600],[994,601],[91,602],[87,603],[86,1],[88,604],[89,1],[90,605],[1067,606],[1066,607],[1047,608],[970,609],[974,609],[975,610],[972,611],[969,1],[971,612],[968,613],[114,614],[115,615],[113,616],[112,1],[119,617],[1056,618],[930,619],[1052,620],[1055,621],[973,622],[1048,623],[1049,623],[1050,624],[1051,625],[1058,626],[1057,627],[120,1],[121,1],[966,628],[122,1],[915,23],[909,629],[916,630],[1072,631],[908,632],[124,23],[1059,633],[1080,634],[964,635],[963,636],[965,637],[123,1],[1065,638],[922,639],[126,640],[129,1],[92,641],[1068,642],[93,1],[108,1],[127,1],[125,1],[128,1],[130,643],[131,1],[85,644]],"affectedFilesPendingEmit":[1054,1088,1087,1085,1086,1084,1082,1083,1067,1066,1047,970,974,975,972,969,971,968,114,115,113,112,119,1056,930,1052,1055,973,1048,1049,1050,1051,1058,1057,120,121,966,122,915,909,916,1072,908,124,1059,1080,964,963,965,123,1065,922,126,1068,108,127,125,128,130,131,85],"version":"5.8.3"}
```
its-me-abhishek: this file should never be pushed to a repo, its just a build file and leads to conflicts

dashboard/package.json
```diff
@@ -22,17 +22,11 @@
     "axios": "^1.8.4",
     "d3": "^7.9.0",
     "dayjs": "^1.11.13",
-    "framer-motion": "^12.6.5",
```
its-me-abhishek, line 25: why is axios added to a frontend? imo it should be removed EDIT: if unused. radix-ui icons are still there as a dependency, should be removed as discussed on discord

dashboard/.gitignore
```diff
@@ -17,6 +17,7 @@
 .env.development.local
 .env.test.local
 .env.production.local
+.env
```
its-me-abhishek: we need not have any .env file in the frontend though

its-me-abhishek (changes requested):
Please fix, will add more changes if I find any issues. Also, in Dashboard.tsx:391, You are simply returning the -       case Page.MINER_STATS:
        return <MinedSharesExplorer />;, try to wrap it inside a Card component like other components for the time being, for smoother display UI

mcelrath (changes requested):
Tested and looks good. `react-icons` is missing as a dependency.

Please add that and I'll merge.

its-me-abhishek:
These are some Frontend issues I have encountered so far, will post a review about backend and about frontend (after I go through the code): 

1. X-axis should have shown timestamps instead of dates on each graph
Screenshot 2025-06-19 162037

2. Is this data static? either ways it is wrongly mapped I assume. Also, the bg-color of this graph does not match others. 
Screenshot 2025-06-19 162229

3. BTW did you implement the download function? I got a chart.svg for all graphs, and got this image every time
Screenshot 2025-06-19 162240

4. Should have some explanation imo, what does Primary stand for (might be a small sentence)
Screenshot 2025-06-19 162305

5. Overflow when I opened console for Websocket errors
Screenshot 2025-06-19 162332

6. Race condition, will fail in FireFox, similar to #192 
Screenshot 2025-06-19 162357

7. Please lookout if its a bug in Footer or BeadsTab, but they get scrollable whenever I open console
Screenshot 2025-06-19 162421

8. Overflow, on left, also I guess 1e+2 should be some whole no, not sure on this. About the overflow, I had this issue in BTC prices page, should be checked, and width is needed to be updated accordingly
Screenshot 2025-06-19 162501

dashboard/tsconfig.json
```diff
@@ -33,8 +33,12 @@
     "**/*.ts",
     "**/*.tsx",
     ".next/types/**/*.ts",
-    "src"
+    "src",
+    "api/api",
```
its-me-abhishek: NodeAPI, Shall be removed as they dont exist anymore

dashboard/src/components/Dashboard/Dashboard.tsx
```diff
@@ -382,14 +382,22 @@ const Dashboard = () => {
             </Card>
           </Box>
         );
+      case Page.MINER_STATS:
+        return (
+          <Box>
```
its-me-abhishek: This does not have that sx element, as compared to others

dashboard/api/utils/BeadsTab/fetchHashrate.js
```diff
@@ -0,0 +1,51 @@
...
+    const timestamp = new Date().toISOString();
+
+    statsHistory.push({
+      value: hashrate,
+      label: new Date(timestamp).toLocaleString(),
+      date: timestamp,
+      latency,
+    });
+
+    if (statsHistory.length > 100) statsHistory.shift();
```
its-me-abhishek: This is an O(n) operation on the backend, shall be handled on Frontend easily

dashboard/api/utils/BeadsTab/fetchHashrate.js
```diff
@@ -0,0 +1,51 @@
...
+  try {
+    const startTime = Date.now();
+
+    const [hashrate, difficulty] = await Promise.all([
+      rpcWithEnv({ method: 'getnetworkhashps' }),
+      rpcWithEnv({ method: 'getdifficulty' }),
+    ]);
+
+    const latency = Date.now() - startTime;
+    const timestamp = new Date().toISOString();
```
its-me-abhishek: is this necessary to Parse This?

dashboard/api/utils/BeadsTab/fetchLatency.js
```diff
@@ -0,0 +1,48 @@
...
+
+    const avgPing = pings.length
+      ? pings.reduce((a, b) => a + b, 0) / pings.length
+      : 0;
+    const peakLatency = pings.length ? Math.max(...pings) * 1000 : 0;
+    const peerCount = pings.length;
+    const latency = Math.round(avgPing * 1000);
+    const timestamp = new Date().toISOString();
+
+    latencyHistory.push({
```
its-me-abhishek: again, imo, shall be handled on the frontend

dashboard/api/utils/BeadsTab/fetchLatency.js
```diff
@@ -0,0 +1,48 @@
...
+      date: timestamp,
+    });
+
+    if (latencyHistory.length > 100) latencyHistory.shift();
+
+    const payload = {
+      type: 'latency_update',
+      data: {
+        chartData: latencyHistory,
+        averageLatency: latency.toFixed(2),
```
its-me-abhishek: Latency was Rounded earlier, I dont think it will have Decimals, why was toFixed used? or was it an error earlier? one of these should go

dashboard/api/utils/BeadsTab/fetchRpc.js
```diff
@@ -0,0 +1,21 @@
+import axios from 'axios';
+
+export async function callRpc({ url, user, pass, method, params = [] }) {
+  const payload = {
```
its-me-abhishek: RPC connection can fail sometimes, should have used some Timeout to retry after X seconds

dashboard/api/utils/BeadsTab/fetchrewards.js
```diff
@@ -0,0 +1,65 @@
...
+
+    const rewardHistory = [];
+    let height = 0;
+
+    const totalRewards = blockCount * blockReward;
+    const rewardRate = blockReward * 144; // Block reward per day
+
+    console.log(`Generating reward history up to block ${currentHeight}`);
+
+    while (height <= currentHeight && blockReward >= 0.00000001) {
```
its-me-abhishek: Can be improved i guess? or maybe handled on frontend only.. these things can slowdown the backend as it keeps on running

dashboard/api/server.js
```diff
@@ -1,16 +1,23 @@
 import { WebSocketServer } from 'ws';
 import dotenv from 'dotenv';
-import fetchBitcoinPrices from './utils/fetchBitcoinPrices.js';
-import fetchGlobalCryptoData from './utils/fetchGlobalData.js';
+import fetchBitcoinPrices from './utils/BitcoinPrice/fetchBitcoinPrices.js';
+import fetchGlobalCryptoData from './utils/BitcoinPrice/fetchGlobalData.js';
+import { fetchHashrateStats } from './utils/BeadsTab/fetchHashrate.js';
```
its-me-abhishek: Directory name should be changed, Frontend and Backend should not be much related

dashboard/src/components/BeadsTab/Trends/LatencyTab.tsx
```diff
@@ -1,36 +1,88 @@
...
+import { LatencyPayload } from '../lib/types';
 
 export default function LatencyTab({
-  chartData,
   isChartLoading,
   chartHovered,
   setChartHovered,
   timeRange,
 }: any) {
+  const [chartData, setChartData] = useState<any[]>([]);
```
its-me-abhishek: maybe shall have used props

dashboard/src/components/BeadsTab/BeadRow.tsx
```diff
@@ -1,41 +1,15 @@
 import React, { useState } from 'react';
-import { motion, AnimatePresence } from 'framer-motion';
-import { ChevronDown, Zap } from 'lucide-react';
+import { ChevronDown } from 'lucide-react';
 import TransactionList from './TransactionList';
-import { shortenHash } from './lib/utils/shortenHash';
-import type { Bead, Transaction } from './lib/types';
+import { shortenHash } from './lib/utils/utils';
```
its-me-abhishek: shortenHash and formatWork come from the same utils folder, imported twice from the same file

dashboard/src/components/BeadsTab/BeadRow.tsx
```diff
@@ -49,102 +23,59 @@ export default function BeadRow({
...
-            <motion.div
-              initial={{ opacity: 0, scale: 0 }}
-              animate={{ opacity: 1, scale: 1 }}
-              className="ml-2"
-            >
-              <Zap className="h-4 w-4 text-yellow-400" />
-            </motion.div>
-          )}
+          </span>
+          {isExpanded && <div className="ml-2"></div>}
```
its-me-abhishek: This empty div inside the bead name conditional rendering serves no purpose and should be removed else some placeholder be added

dashboard/src/components/BeadsTab/MinedSharesExplorer.tsx
```diff
@@ -1,153 +1,101 @@
...
-  const [error, setError] = useState<string | null>(null);
-
-  // Data fetching
-  const {
-    data: chartData,
-    isLoading: isChartLoading,
-    error: chartError,
-    refetch,
-  } = useChartData(timeRange);
+  const [timeRange] = useState('month');
```
its-me-abhishek: why is a constant being used in a state? similar was flagged earlier and that has not been updated either

dashboard/src/components/BeadsTab/MinedSharesExplorer.tsx
```diff
@@ -1,153 +1,101 @@
...
+                  <div className="p-8">
+                    <div className="h-12 bg-gray-800/50 rounded-md animate-pulse mb-4"></div>
+                    <div className="h-12 bg-gray-800/50 rounded-md animate-pulse"></div>
+                  </div>
+                ) : (
+                  BEADS.map((bead) => (
+                    <BeadRow
+                      key={bead.id}
+                      bead={bead}
+                      isExpanded={!!expandedBeads[bead.id as BeadId]}
```
its-me-abhishek: you do bead.id as BeadId multiple times, but there’s no runtime guarantee the IDs conform. htis could lead to runtime errors or type mismatches later on if changes are introduced

dashboard/src/components/BeadsTab/MinedSharesExplorer.tsx
```diff
@@ -1,153 +1,101 @@
...
-    if (bead) toggleBead(bead.id);
-  };
-
-  const handleRetry = () => {
-    setError(null);
-    refetch();
+    navigator.clipboard
+      .writeText(parentHash)
+      .then(() => {
+        console.log(`Parent hash copied to clipboard:\n${parentHash}`);
```
its-me-abhishek: consider adding UI feedback (toast, tooltip, a green text maybe) to inform users their action succeeded or failed instead of just console.log.

its-me-abhishek (changes requested):
Please fix, will add more changes if I find any issues.

dashboard/api/server.js
```diff
@@ -47,11 +54,10 @@ async function sendDataToClients() {
 
 // Fetch and send every 10 seconds
 setInterval(sendDataToClients, 1000);
+setInterval(() => fetchHashrateStats(wss), 1000);
```
its-me-abhishek: I think these can be put inside a same setInterval as well? also that comment seems to be wrong, 1000 is in ms (probably), so it sends per second and not per 10 second... needs to be fixed

dashboard/src/components/BeadsTab/AdvancedChart.tsx
```diff
@@ -1,282 +1,75 @@
...
-    } else {
-      chartControls.start({
-        pathLength: 0,
-        transition: { duration: 0.5 },
-      });
-    }
-  }, [isInView, chartControls]);
-
-  // Refresh animation
   const handleRefresh = () => {
```
its-me-abhishek: this function only refreshes the animation(shows a loader) and not graph i guess?

its-me-abhishek:
For the frontend issues, some of the issues from the [previous comment](https://github.com/braidpool/braidpool/pull/208#issuecomment-2987689974) persist

1. Maybe a key is missing in a map shall be easy to fix Screenshot 2025-06-20 220747

2. The date is the same as prev, but time is updated (and appears same for me) - is it randomly generated?
Screenshot 2025-06-20 221313

3. The overflow again persists, moving divs won't work, probably just put that inside another flex
Screenshot 2025-06-20 221406

4. Race condition was not fixed as well
Screenshot 2025-06-20 221428

5. This shows invalid for some reason
Screenshot 2025-06-20 221507

6. These buttons can be removed for now as they don't do anything I guess (or maybe not functioning), as they might need to be tested unnecessarily
Screenshot 2025-06-20 221521

7. The X-axis look very weird, they donot update over ms so accuracy is not an issue, yet I get randomly multiple instances of the same seconds' plots.. how is this plotted?

https://github.com/user-attachments/assets/f3bab584-c517-403b-9b6a-129070b7b001

Screenshot 2025-06-20 221821
Screenshot 2025-06-20 221838
Screenshot 2025-06-20 222126
Screenshot 2025-06-20 222140

8. The bg color needs to be fixed
Screenshot 2025-06-20 222152

9.  https://github.com/user-attachments/assets/27040c5a-fee2-436b-af03-0fad3df78737

I think they (graphs and those 2 blocks) should not appear together on clicking history? only graph should appear


10. The svg works. But the refresh button still does not work, and the graphs seem to persist the black bg after downloading an svg


https://github.com/user-attachments/assets/b9741cf0-fe25-49f0-b94f-4fe08bd35601

dashboard/src/components/Dashboard/Dashboard.tsx
```diff
@@ -382,14 +382,22 @@ const Dashboard = () => {
             </Card>
           </Box>
         );
+      case Page.MINER_STATS:
+        return (
+          <Box sx={{ p: 1 }}>
+            <Card>
```
its-me-abhishek: why did you not add a title in this like others and remove that header from MinedSharesExplorer? i think that would have made the UI a lot more consistent

dashboard/api/ws/handleWebSocketConnection.js
```diff
@@ -0,0 +1,29 @@
...
+  ws.on('message', async (message) => {
+    try {
+      const data = JSON.parse(message);
+
+      if (data.type === 'rpc_call') {
+        const result = await callRpc({
+          url: process.env.BRAIDPOOL_URL,
+          user: process.env.RPC_USER,
+          pass: process.env.RPC_PASS,
+          method: data.method,
```
its-me-abhishek, line 57: vulnerable imo, any method can be passed into this

dashboard/api/server.js
```diff
@@ -45,13 +52,16 @@ async function sendDataToClients() {
   }
 }
 
-// Fetch and send every 10 seconds
-setInterval(sendDataToClients, 1000);
+// Fetch and broadcast all data every second
+setInterval(() => {
+  Promise.all([
```
its-me-abhishek: Promise.all should not be used. If one fails, the whole batch will fail, making the complete dashboard suffer

dashboard/src/components/BeadsTab/Trends/HashrateTab.tsx
```diff
@@ -1,65 +1,132 @@
...
 
 export default function HashrateTab({
-  chartData,
   isChartLoading,
   chartHovered,
   setChartHovered,
   timeRange,
 }: any) {
+  const [latestStat, setLatestStat] = useState<any>(null);
+  const [chartData, setChartData] = useState<any[]>([]);
```
its-me-abhishek: should have used props here as well, type-safe code is required in case one uses TS, Latency Tab also uses `any` in some props, needs to be type checked

dashboard/src/components/BeadsTab/Trends/LatencyTab.tsx
```diff
@@ -19,47 +81,48 @@ export default function LatencyTab({
...
+          tooltipFormatter={(value, name) => [`${value} ms`, name]}
         />
       </div>
 
       <div className="grid grid-cols-1 md:grid-cols-3 gap-4">
         <AnimatedStatCard
           title="Average Latency"
-          value="118ms"
+          value={averageLatency}
           change="-3%"
```
its-me-abhishek: changes, delays, etc. should not be hardcoded. might have to add some function to calculate them if possible

its-me-abhishek (commented):
Please fix these issues, and test on your end, will add more changes if I find any

its-me-abhishek:
1. The socket doesn't work at first and throws a console error again

Screenshot 2025-06-23 165214

2. You've created a separate hook for the socket does it really help? It takes similar LOC as using WebSocket directly so becomes kind of a wrapper of a wrapper. Why not just use the WebSocket directly? a centralised one might just make things a bit too convoluted.

3. The refresh button is not appearing anymore was it removed?

4. Changes (+ve, -ve) on stats do not appear anymore as well, is it a bug?

5. [It's still vulnerable and unfixed](https://github.com/braidpool/braidpool/pull/208#discussion_r2159891735)

its-me-abhishek:
6. The rewards tab still has an overflow issue. The other tabs also have an overflow issue.
Screenshot 2025-06-23 170505
Screenshot 2025-06-23 165712


7. Graphs can get mixed up if many are created; consider adding the graph name, x-y axes, or date+graph name to help identify them.

8. "Peak hashrate (24h)" doesn't really make sense without historical data and is not useful in its current form. I think peak 24h should mean peak since past 24 hrs in a conventional manner, currently it requires the user to stay on the dash to check for the 24h peak, making it more like a session based peak hashrate

```
  if (hashrate > peakHashrate) {
    peakHashrate = hashrate;
  }
```

9. "Copied" works on parents but not on hash.
Screenshot 2025-06-23 172600

10. Bug in the component here, two timestamps with same value on x-axis
Screenshot 2025-06-23 165519

11. 10 Seconds update is too late or high (maybe, imo websockets should be kind of fast, else it will be similar to REST)

12. Seems like you misinterpreted the 7th comment [here](https://github.com/braidpool/braidpool/pull/208#issuecomment-2992327977)

image

priyashuu:
> 1. The socket doesn't work at first and throws a console error again
> 
> Screenshot 2025-06-23 165214
> 
> 2. You've created a separate hook for the socket does it really help? It takes similar LOC as using WebSocket directly so becomes kind of a wrapper of a wrapper. Why not just use the WebSocket directly? a centralised one might just make things a bit too convoluted.
> 3. The refresh button is not appearing anymore was it removed?
> 4. Changes (+ve, -ve) on stats do not appear anymore as well, is it a bug?
> 5. [It's still vulnerable and unfixed](https://github.com/braidpool/braidpool/pull/208#discussion_r2159891735
I'm not experiencing any errors. How did you encounter the error you mentioned?
Screenshot 2025-06-23 184625

priyashuu:
.Yes, I removed the refresh svg since it was already updating every 10 seconds. I also removed the stats to enhance simplicity.

its-me-abhishek:
I have faced this in both Brave and Firefox, its similar to what we faced in #192  maybe
image
image

dashboard/api/utils/fetchHashrate.js
```diff
@@ -0,0 +1,42 @@
+import { rpcWithEnv } from './rpcWithEnv.js';
```
its-me-abhishek, line 1: - you call Date.now() multiple times within milliseconds. you can reuse a single value unless you need them to differ.
- magic numbers should be clarified

dashboard/api/utils/fetchblockdetails.js
```diff
@@ -0,0 +1,97 @@
...
+    const blockData = await rpcWithEnv({
+      method: 'getblock',
+      params: [blockHash, 2],
+    });
+
+    const coinbaseTx = blockData.tx[0];
+    const rewardBTC = coinbaseTx.vout.reduce((acc, out) => acc + out.value, 0);
+
+    const transactions = blockData.tx.slice(1).map((tx, index) => {
+      const feeBTC = tx.fee !== undefined ? tx.fee : 0.0001;
```
its-me-abhishek: why is it 0.0001?

dashboard/api/ws/handleWebSocketConnection.js
its-me-abhishek: seems to be unfixed yet

dashboard/api/server.js
```diff
@@ -45,13 +52,19 @@ async function sendDataToClients() {
...
-
-// WebSocket connection handler
-wss.on('connection', (ws) => {
-  console.log('Client connected');
-  ws.send(JSON.stringify({ type: 'connection', status: 'connected' }));
-});
+// Fetch and broadcast all data every 10 seconds (more reasonable interval)
+setInterval(async () => {
+  try {
+    await Promise.allSettled([
```
its-me-abhishek: still this will probably throw error for the first one only, and not if any other fails.

its-me-abhishek (commented):
have reviewed the backend, will review the frontend once the websocket gets fixed
EDIT: typo

dashboard/api/server.js
priyashuu: I have removed the `Promise.allSettled`.

dashboard/api/server.js
priyashuu: In Bitcoin's early days, a common default transaction fee was 0.0001 BTC, which is about 10,000 satoshis. However, I have updated the default value to "Unknown."

its-me-abhishek:
For Frontend:
The transaction tab does not work anymore 
Screenshot 2025-06-25 105208

majority of the frontend issues flagged in previous comments still stand the same as unfixed. These are some that I have observed so Far. you'd have to go through the thread to find them though - 
Screenshot 2025-06-25 104942
Screenshot 2025-06-25 104951
Screenshot 2025-06-25 105009
Screenshot 2025-06-25 105051
Screenshot 2025-06-25 105105
Screenshot 2025-06-25 105348

dashboard/api/utils/fetchHashrate.js
```diff
@@ -0,0 +1,42 @@
...
+    const latency = now - startTime;
+    const timestamp = now;
+
+    const payload = {
+      type: 'hashrate_data',
+      data: {
+        hashrate: hashrateEH,
+        timestamp: timestamp,
+        networkDifficulty: lastDifficulty,
+        latency,
```
its-me-abhishek: latency being passed in Hashrate? is it the interval between updates? 
```
    const latency = now - startTime;
```
also it is always = 0

this is the console.log of the payload-> 
```{
  type: 'hashrate_data',
  data: {
    hashrate: 1002.800525137721,
    timestamp: 1750934924695,
    networkDifficulty: 126411437451912.2,
    latency: 0
  }
}```

File: dashboard/api/utils/fetchblockdetails.js [hunk 38]
```diff
@@ -0,0 +1,99 @@
+import WebSocket from 'ws';
+import { rpcWithEnv } from './rpcWithEnv.js';
+
+let lastBlockHash = null;
```
its-me-abhishek, line 4: lastBlockHash is a shared and mutable global, maybe can cause race conditions in environments with concurrent invocations, shall be moved to the function if it works there without breaking things

dashboard/api/ws/handleWebSocketConnection.js
```diff
@@ -0,0 +1,63 @@
...
+  'getblock',
+  'getblockhash',
+  'getdifficulty',
+  'getnetworkhashps',
+  'getmempoolinfo',
+  'getpeerinfo',
+  'getblockchaininfo',
+]);
+
+export async function handleWebSocketConnection(ws, wss) {
```
its-me-abhishek: wss declared but never read

dashboard/api/server.js
```diff
@@ -45,13 +52,25 @@ async function sendDataToClients() {
...
-wss.on('connection', (ws) => {
-  console.log('Client connected');
-  ws.send(JSON.stringify({ type: 'connection', status: 'connected' }));
-});
+  fetchReward(wss).catch((err) =>
+    console.error('[Server] fetchReward failed:', err)
+  );
+}, 10000); // 10-second interval for better performance
 
 console.log('WebSocket server running on ws://localhost:5000');
```
its-me-abhishek: the log is incorrect, should have used PORT

dashboard/src/components/BeadsTab/Reward/generateRewardHistory.tsx
```diff
@@ -0,0 +1,22 @@
+export function generateRewardHistory(
+  blockCount: number,
+  currentReward: number = 3.125
```
its-me-abhishek, line 3: here also, unused, and hardcoded var

dashboard/src/components/BeadsTab/Trends/LatencyTab.tsx
```diff
@@ -1,66 +1,179 @@
...
-  timeRange,
-}: any) {
+import { LatencyData } from '../lib/types';
+
+const MAX_LATENCY_HISTORY = 100;
+
+export default function LatencyTab({ timeRange }: { timeRange: string }) {
+  const [latencyData, setLatencyData] = useState<LatencyData>({
+    chartData: [],
+    averageLatency: 'Laoding',
```
its-me-abhishek: typo

dashboard/src/components/BeadsTab/lib/utils/useCopyToClipboard.ts
```diff
@@ -0,0 +1,14 @@
+import { useState } from 'react';
+
+export default function useCopyToClipboard(timeout = 1500) {
```
its-me-abhishek: should be moved to utils itself

dashboard/src/components/BeadsTab/BeadRewardTooltip.tsx
```diff
@@ -20,40 +14,38 @@ export function BeadRewardTooltip({ reward }: BeadRewardTooltipProps) {
```
its-me-abhishek: why is the USD Value hardcoded??

dashboard/src/components/BeadsTab/BeadRewardTooltip.tsx
```diff
@@ -20,40 +14,38 @@ export function BeadRewardTooltip({ reward }: BeadRewardTooltipProps) {
...
-      </AnimatePresence>
+      </div>
+
+      {showTooltip && (
+        <div
+          className={`
+            absolute z-50 top-full left-1/2 transform -translate-x-1/2 translate-y-2 w-48 
+            bg-gray-900/95 border border-gray-700 rounded-lg shadow-xl p-3
+            transition-all duration-200
+            ${showTooltip ? 'opacity-100 scale-100' : 'opacity-0 scale-90 -translate-y-1'}
```
its-me-abhishek: You're already checking showTooltip in the conditional render ({showTooltip && (...)}). So this:
is unreachable else-wise, since showTooltip === true.

dashboard/src/components/BeadsTab/BeadRow.tsx
```diff
@@ -48,103 +20,60 @@ export default function BeadRow({
...
-
-          <motion.span
-            className="text-blue-100 font-medium font-mono text-sm sm:text-base truncate md:truncate-0"
-            animate={{
-              color: isExpanded ? '#93c5fd' : '#e0e7ff',
-            }}
-            transition={{ duration: 0.3 }}
+          </div>
+          <span
+            className={`text-sm sm:text-base truncate md:truncate-0 ${
```
its-me-abhishek: there is no truncate-0 in Tailwind though

dashboard/src/components/BeadsTab/TransactionList.tsx
```diff
@@ -1,121 +1,81 @@
...
-      <div className="grid sm:grid grid-cols-6 text-xs text-blue-300 font-semibold mb-2 px-2">
-        <div>Hash</div>
-        <div>Size (vbytes)</div>
-        <div>Fee (BTC)</div>
-        <div>Fee Rate (sats/vB)</div>
-        <div>Inputs</div>
-        <div>Outputs</div>
+        Showing {limitedTransactions.length} of {transactions.length}{' '}
+        Transactions
+        {hasMoreTransactions && (
```
its-me-abhishek, line 21: this is not doing anything with it besides a message, which can be replaced with the length var itself i suppose

its-me-abhishek (commented):
Please fix these issues

its-me-abhishek:
For frontend, the issues Still persist:

1. Beads tab
Screenshot 2025-06-26 160313
Screenshot 2025-06-26 161333

2. Does not throw an error now but still does not work 
Screenshot 2025-06-26 160750

3. The graphs still have that issue as discussed on Discord 

https://github.com/user-attachments/assets/7d229a8a-2ac1-483e-b966-267cc53edec6

https://github.com/user-attachments/assets/8037cf29-acb1-41fb-96d4-ea0daca33dce

dashboard/api/utils/fetchblockdetails.js
```diff
@@ -0,0 +1,98 @@
...
+    const startTime = Date.now();
+    const blockchainInfo = await rpcWithEnv({ method: 'getblockchaininfo' });
+    const latestHeight = blockchainInfo.blocks;
+
+    const blockHash = await rpcWithEnv({
+      method: 'getblockhash',
+      params: [latestHeight],
+    });
+
+    if (blockHash === lastBlockHash) {
```
its-me-abhishek: won't this be useless if lastBlockHash is always null

dashboard/api/utils/fetchblockdetails.js
```diff
@@ -0,0 +1,98 @@
...
+      method: 'getblock',
+      params: [blockHash, 2],
+    });
+
+    const coinbaseTx = blockData.tx[0];
+    const rewardBTC = coinbaseTx.vout.reduce((acc, out) => acc + out.value, 0);
+
+    const transactions = blockData.tx.slice(1).map((tx, index) => {
+      const feeBTC = tx.fee !== undefined ? tx.fee : 'Unknown';
+      const size = tx.size || tx.weight || 225;
```
its-me-abhishek: tx.weight isn't equivalent to size though, and is 225 a good fallback?

dashboard/api/utils/fetchblockdetails.js
```diff
@@ -0,0 +1,98 @@
...
+      const feeRate = size > 0 ? (feeBTC * 1e8) / size : 0;
```
its-me-abhishek, line 41: feeBTC might be a string ('Unknown'), which could break the feeRate calc

dashboard/api/utils/fetchblockdetails.js
```diff
@@ -0,0 +1,98 @@
...
+        difficulty: blockData.difficulty,
+        txCount: blockData.tx.length,
+        reward: rewardBTC,
+        parent: blockData.previousblockhash,
+        transactions,
+      },
+    };
+
+    const totalFees = transactions.reduce((acc, tx) => acc + tx.fee, 0);
+    const avgFeeRate =
```
its-me-abhishek, line 74: You compute totalFees, avgFeeRate, and avgTxSize by iterating the array three times, could be put inside a single loop for less complexitty if possible

dashboard/api/utils/fetchblockdetails.js
```diff
@@ -0,0 +1,98 @@
...
+    const avgTxSize =
+      transactions.length > 0
+        ? transactions.reduce((acc, tx) => acc + tx.size, 0) /
+          transactions.length
+        : 0;
+
+    const statsPayload = {
+      type: 'transaction_stats',
+      data: {
+        mempoolSize: blockData.tx.length - 1,
```
its-me-abhishek: mempoolSize and txRate are given the same value `blockData.tx.length`. i think this is incorrect, mempool size should include all txns, and should be fetched from getmempoolinfo instead

dashboard/src/components/BeadsTab/lib/utils/blockProcessor.ts
```diff
@@ -0,0 +1,31 @@
...
+    height,
+    difficulty,
+    txCount,
+    reward,
+    parent,
+    transactions,
+  } = data;
+
+  const work = `${(difficulty / 1e6).toFixed(2)} EH`;
+  const formattedTransactions = transactions.map((tx: any) => ({
```
its-me-abhishek: might use props if the structure is finalised

dashboard/src/components/BeadsTab/MinedSharesExplorer.tsx
```diff
@@ -1,153 +1,190 @@
...
-                    ))
-                  )}
-                </div>
+                    <div className="h-12 bg-gray-800/50 rounded-md animate-pulse mb-4"></div>
+                    <div className="h-12 bg-gray-800/50 rounded-md animate-pulse"></div>
+                  </div>
+                ) : (
+                  liveBeads.map((bead) => (
+                    <BeadRow
+                      isActive
```
its-me-abhishek: You're probably passing isActive as true to all BeadRow components. This seems incorrect. It should probably be isActive={activeBead === bead.id} if activeBead is used for styling the currently expanded/selected bead. If isActive serves another purpose, please clarify its intent.

dashboard/src/components/BeadsTab/MinedSharesExplorer.tsx
```diff
@@ -1,153 +1,190 @@
...
-import EnhancedBlocksTab from './BlockVisulisation/Block';
 import BeadRow from './BeadRow';
-import { BEADS, TRANSACTIONS } from './lib/constants';
-import { useChartData } from './Hooks/useChartData';
 import { TrendsTab } from './Trends/TrendsTab';
 import { RewardsDashboard } from './Reward/RewardsSection';
+import { Transaction, Bead } from './lib/types';
+import { processBlockData } from './lib/utils/blockProcessor';
+
+type BeadId = string;
```
its-me-abhishek: types should be in types.ts

dashboard/src/components/BeadsTab/TransactionList.tsx
its-me-abhishek: seems to be unfixed

dashboard/src/components/BeadsTab/MinedSharesExplorer.tsx
```diff
@@ -1,153 +1,190 @@
-import React, { useState, useRef, useEffect } from 'react';
-import { useTransform, useScroll } from 'framer-motion';
-
+import React, { useState, useEffect, useRef } from 'react';
```
its-me-abhishek, line 1: Not sure on this but: if there are too many beads one would have to scroll a lot, maybe a paginator would work

its-me-abhishek (commented):
Please fix these

its-me-abhishek:
On frontend, this issue still persists, 
image

dashboard/src/components/BeadsTab/TransactionList.tsx
priyashuu: Thanks for the suggestion! You're right that if it were only for the message, calculating Math.min(transactions.length, 10) directly would make sense. But since we're also slicing the transactions array for rendering the rows themselves, having limitedTransactions as a variable keeps it clean and avoids recalculating the slice multiple times.

dashboard/api/utils/fetchblockdetails.js
priyashuu: 225 vBytes is an average estimate for a typical Bitcoin transaction , only an approxmiation

dashboard/src/components/BeadsTab/MinedSharesExplorer.tsx
priyashuu: added a paginator

dashboard/src/components/BeadsTab/lib/utils/blockProcessor.ts
priyashuu: Not yet. I believe we need to add more fields in the future, such as miner information, health details, and additional fee information. It is not finalized yet; I will add more info  after this PR.

PR #213, opened 2025-06-18
feat: Add IPC implementation to fetch block template and notifications

zaidmstrr:
This PR adds a new way to call the `getblocktemplate` function using IPC and listens for block notifications. This also uses different additional features such as:

- Priority-based request processing
- Event-driven approach to handle the requests
- Metrics and monitoring – to check the queue sizes and overall health of the network
- Error handling (restarting the connection, provides debug errors in case of failure)

#### To Test:
- Navigate to the root path `/`  or inside `/node`
- Run `cargo run -- --ipc --ipc-socket=/tmp/bitcoin-ipc.sock ` (where the socket path is the same as you opted for while running bitcoin-node)

I preserved the existing ZMQ functionality for now.
This PR needs critical reviews and evaluation. The higher the reviewers, the higher the chances of catching the bug.

### DIFF
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -43,4 +43,12 @@ pub struct Cli {
     /// Use this port for bitcoin ZMQ
     #[arg(long, default_value = "28332")]
     pub zmqhashblockport: u16,
+
+    /// Use IPC Bitcoin Core communication
+    #[arg(long, default_value = "false")]
+    pub ipc: bool,
+
+    /// Path to Bitcoin Core IPC socket (used when --ipc is enabled)
+    #[arg(long, default_value = "/tmp/bitcoin-ipc.sock")]
+    pub ipc_socket: String,
 }
--- a/node/src/ipc.rs
+++ b/node/src/ipc.rs
@@ -1,0 +1,341 @@
+//! Listens for block notifications and fetches new block templates via IPC
+use tokio::sync::mpsc::Sender;
+use tokio::sync::oneshot;
+pub mod client;
+pub use client::{bytes_to_hex, BitcoinNotification, RequestPriority, SharedBitcoinClient};
+
+/// Main IPC block listener that maintains connection to Bitcoin Core and forwards block templates
+///
+/// This function implements a robust connection loop that:
+/// * Automatically reconnects on connection failures
+/// * Fetches initial block template on startup (if node is synced)
+/// * Listens for new block notifications and fetches fresh templates
+/// * Provides health monitoring and connection statistics
+/// * Handles graceful degradation when Bitcoin Core is not fully synced
+pub async fn ipc_block_listener(
+    ipc_socket_path: String,
+    block_template_tx: Sender<Vec<u8>>,
+) -> Result<(), Box<dyn std::error::Error>> {
+    log::info!("Starting IPC block listener on: {}", ipc_socket_path);
+    let local = tokio::task::LocalSet::new();
+    local.run_until(async move {
+        let mut health_check_interval = tokio::time::interval(tokio::time::Duration::from_secs(30));
+        let mut detailed_stats_interval = tokio::time::interval(
+            tokio::time::Duration::from_secs(100)
+        );
+        loop {
+            let mut shared_client = match SharedBitcoinClient::new(&ipc_socket_path).await {
+                Ok(client) => {
+                    log::info!("IPC connection established");
+                    client
+                }
+                Err(e) => {
+                    log::error!("Failed to connect to IPC socket: {}", e);
+                    log::info!("Retrying connection in 10 seconds...");
+                    tokio::time::sleep(tokio::time::Duration::from_secs(10)).await;
+                    continue;
+                }
+            };
+
+            let initial_sync_check = match
+                shared_client.is_recently_synced(Some(RequestPriority::High)).await
+            {
+                Ok(is_synced) => {
+                    if !is_synced {
+                        log::warn!("Node is not synced - waiting for sync to complete");
+                    } else {
+                        log::info!("Node is synced and ready to be used");
+                    }
+                    is_synced
+                }
+                Err(e) => {
+                    log::error!("Initial sync check failed: {}", e);
+                    if is_connection_error(&e) {
+                        log::error!("Connection lost during initial sync check - reconnecting");
+                        continue; // Restart the main connection loop
+                    }
+                    log::warn!("Continuing without sync check");
+                    false
+                }
+            };
+
+            // Only try to get initial template if node is synced
+            if initial_sync_check {
+                match get_template_with_retry(
+                    &mut shared_client,
+                    3,
+                    RequestPriority::High,
+                    "initial template"
+                ).await {
+                    Ok(template) => {
+                        log::info!("Got initial block template: {} bytes", template.len());
+                        if let Err(e) = block_template_tx.send(template).await {
+                            log::error!("Failed to send initial template: {}", e);
+                            continue;
+                        }
+                    }
+                    Err(e) => {
+                        log::error!("Failed to get initial template: {}", e);
+                        if is_connection_error(&e) {
+                            log::error!("Connection lost getting initial template - reconnecting");
+                            continue;
+                        }
+                        // Continue anyway - we'll get templates on block changes
+                    }
+                }
+            }
+
+            let mut notification_receiver = match shared_client.take_notification_receiver() {
+                Some(receiver) => receiver,
+                None => {
+                    log::error!("Failed to get notification receiver - reconnecting");
+                    tokio::time::sleep(tokio::time::Duration::from_secs(5)).await;
+                    continue;
+                }
+            };
+
+            log::info!("listening for block notifications...");
+
+            // Listen for block connect notifications only
+            let should_reconnect = loop {
+                tokio::select! {
+                    notification = notification_receiver.recv() => {
+                        match notification {
+                            Some(BitcoinNotification::BlockConnected { height, hash, .. }) => {
+                                let mut hash_reversed = hash.clone();
+                                hash_reversed.reverse();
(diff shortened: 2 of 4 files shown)


zaidmstrr:
Simplified the retry logic and reduced this to under 1.5 seconds at max (worst case)

Sansh2356:
Went through it completely, and most implementation follows standard implementation for `capnproto` communication aptly followed provided according to the `capnpc` rust documentation so on the IPC portion there is no issue . On the notification purpose is also nicely done on the basis of bitcoin-core Notification interface defined under Chain. Capnp which implements all the communication endpoints with the `bitcoind` just on the part of processing of request via priority is considered i do not think it is required i find it to be good in terms of implementation side but the processing itself is blocking and since each request will be served once only and multiple requests can't be sent by the sender/miner if he is planning to run multiple downstream nodes then also i think it is not required . Otherwise tested it completely and from my side it can be merged .

Sansh2356:
> Went through it completely, and most implementation follows standard implementation for `capnproto` communication aptly followed provided according to the `capnpc` rust documentation so on the IPC portion there is no issue . On the notification purpose is also nicely done on the basis of bitcoin-core Notification interface defined under Chain. Capnp which implements all the communication endpoints with the `bitcoind` just on the part of processing of request via priority is considered i do not think it is required i find it to be good in terms of implementation side but the processing itself is blocking and since each request will be served once only and multiple requests can't be sent by the sender/miner if he is planning to run multiple downstream nodes then also i think it is not required . Otherwise tested it completely and from my side it can be merged .

Continuing onto this it firstly requires to be merged/rebased with `dev` just for the sake to test the overall structure. If @zaidmstrr you will allow me to integrate i will be willing to take this up so that you can continue on the template work itself .

zaidmstrr:
>  Capnp which implements all the communication endpoints with the `bitcoind` just on the part of processing of request via priority is considered i do not think it is required i find it to be good in terms of implementation side but the processing itself is blocking and since each request will be served once only and multiple requests can't be sent by the sender/miner if he is planning to run multiple downstream nodes then also i think it is not required . 

The priority queue was implemented for a reason and can be used for critical requests, like when a miner submits a full block submission. We need to process this query on an urgent basis; otherwise, we might lose the block reward or the block may get orphaned. 

> i will be willing to take this up so that you can continue on the template work itself .

First propose your changes in the public channel before working on something new. So, others can suggest or review.

Sansh2356:
> > Capnp which implements all the communication endpoints with the `bitcoind` just on the part of processing of request via priority is considered i do not think it is required i find it to be good in terms of implementation side but the processing itself is blocking and since each request will be served once only and multiple requests can't be sent by the sender/miner if he is planning to run multiple downstream nodes then also i think it is not required .
> 
> The priority queue was implemented for a reason and can be used for critical requests, like when a miner submits a full block submission. We need to process this query on an urgent basis; otherwise, we might lose the block reward or the block may get orphaned.
> 
> > i will be willing to take this up so that you can continue on the template work itself .
> 
> First propose your changes in the public channel before working on something new. So, others can suggest or review.

For the willing to take up portion it was meant solely for taking up the integration of your branch into dev branch I hope I have cleared the confusion regarding it .

Sansh2356:
> > Capnp which implements all the communication endpoints with the `bitcoind` just on the part of processing of request via priority is considered i do not think it is required i find it to be good in terms of implementation side but the processing itself is blocking and since each request will be served once only and multiple requests can't be sent by the sender/miner if he is planning to run multiple downstream nodes then also i think it is not required .
> 
> The priority queue was implemented for a reason and can be used for critical requests, like when a miner submits a full block submission. We need to process this query on an urgent basis; otherwise, we might lose the block reward or the block may get orphaned.
> 
> > i will be willing to take this up so that you can continue on the template work itself .
> 
> First propose your changes in the public channel before working on something new. So, others can suggest or review.

On th priority queue part it can lead to starvation since the priorities are currently static and secondly since except the submit block request all the other have similar priority so they can be processed in fifo manner only in a single queue instead of priority structure.

zaidmstrr:
> On th priority queue part it can lead to starvation since the priorities are currently static and secondly since except the submit block request all the other have similar priority so they can be processed in fifo manner only in a single queue instead of priority structure.

How will this occur in starvation when most of the processes have the same priority except the submit block? And the point regarding the submit block is most crucial for us; that's why we are building this project. If a miner founds a block and we delay the processing, and in the meantime, some other pool finds a block and sends it first, then we will lose that block. We don't want to do this in our case, so it's our highest priority task. Thus, indeed, a priority queue is needed here. 

Secondly, we are processing requests at a much higher rate than any other pool. For example, a bead can come every 200 ms, the notification handler, and the getblocktemplate RPC, which is resource-intensive.


> For the willing to take up portion it was meant solely for taking up the integration of your branch into dev branch I hope I have cleared the confusion regarding it .

Okay then, you didn't describe this context earlier. You can take this up for integration.

Sansh2356:
> > On th priority queue part it can lead to starvation since the priorities are currently static and secondly since except the submit block request all the other have similar priority so they can be processed in fifo manner only in a single queue instead of priority structure.
> 
> How will this occur in starvation when most of the processes have the same priority except the submit block? And the point regarding the submit block is most crucial for us; that's why we are building this project. If a miner founds a block and we delay the processing, and in the meantime, some other pool finds a block and sends it first, then we will lose that block. We don't want to do this in our case, so it's our highest priority task. Thus, indeed, a priority queue is needed here.
> 
> Secondly, we are processing requests at a much higher rate than any other pool. For example, a bead can come every 200 ms, the notification handler, and the getblocktemplate RPC, which is resource-intensive.
> 
> > For the willing to take up portion it was meant solely for taking up the integration of your branch into dev branch I hope I have cleared the confusion regarding it .
> 
> Okay then, you didn't describe this context earlier. You can take this up for integration.

I agree with `priority_queue` on the part of `submitblock` wrt to other requests but supposedly except the submitblock all the other requests can be kept in a single queue there is no requirement of further division of queues on the basis of priorities among those requests . Regarding the context for the integration part i had provided the context in the comment itself though hopefully it avoids the confusion now .

zaidmstrr:
It is still needed we are in the early phase; we dont know which type of requests we need to call in the future. For example, currently I'm using basic polling to check the Bitcoin socket connection, and this is the low-priority task. And having those queue categories doesn't impact the performance of the program in any way.

node/src/cli.rs
```diff
@@ -43,4 +43,12 @@ pub struct Cli {
     /// Use this port for bitcoin ZMQ
     #[arg(long, default_value = "28332")]
     pub zmqhashblockport: u16,
+
+    /// Use IPC Bitcoin Core communication
+    #[arg(long, default_value = "false")]
```
mcelrath, line 48: Make this a boolean flag (present or not) instead of accepting a boolean value.

node/src/main.rs
```diff
@@ -104,6 +164,19 @@ async fn main() -> Result<(), Box<dyn Error>> {
     }
 }
 
+async fn ipc_template_consumer(mut template_rx: mpsc::Receiver<Vec<u8>>) {
```
mcelrath, line 167: We also have block_template::consumer. Do we need both? Both of them are stubs at present.

node/src/ipc.rs
```diff
@@ -0,0 +1,341 @@
...
+                    if !is_synced {
+                        log::warn!("Node is not synced - waiting for sync to complete");
+                    } else {
+                        log::info!("Node is synced and ready to be used");
+                    }
+                    is_synced
+                }
+                Err(e) => {
+                    log::error!("Initial sync check failed: {}", e);
+                    if is_connection_error(&e) {
```
mcelrath, line 53: This probably needs an exponential backoff and enumeration of different kinds of connection issues. Not everything requires a reconnect.

node/src/ipc.rs
```diff
@@ -0,0 +1,341 @@
...
+/// * `max_attempts` - Maximum retry attempts (typically 2-3 for fast response)
+/// * `priority` - Request priority affecting queue position
+/// * `context` - Context for logging
+async fn get_template_with_retry(
+    client: &mut SharedBitcoinClient,
+    max_attempts: u32,
+    priority: RequestPriority,
+    context: &str,
+) -> Result<Vec<u8>, Box<dyn std::error::Error>> {
+    const MIN_TEMPLATE_SIZE: usize = 512;
```
mcelrath, line 232: Can you justify this number?

node/src/ipc/client.rs
```diff
@@ -0,0 +1,990 @@
...
+        BlockConnectedParams, BlockConnectedResults, ChainStateFlushedParams,
+        ChainStateFlushedResults, DestroyParams, DestroyResults, UpdatedBlockTipParams,
+        UpdatedBlockTipResults,
+    },
+};
+
+use crate::init_capnp::init::Client as InitClient;
+use crate::proxy_capnp::thread::Client as ThreadClient;
+
+#[allow(dead_code)]
```
mcelrath, line 27: Should be removed, it's used.

mcelrath (approved):
We need to merge ipc/main into dev next.

node/src/ipc.rs
zaidmstrr: There was no hardcore reason to choose this number but practically, if we call the full block template, the data we will get is always more than this threshold. Where a block header is 80 bytes, a coinbase is 180-300 bytes, other metadata (noncerange, capabilities, etc.) is around 150-200 bytes, and block transactions.

node/src/main.rs
zaidmstrr: I just added this to fetch the results of the block template using the IPC method. We can allow this under dead code, I think. And we don't need both; it depends upon which method we are using to communicate with bitcoin; if IPC then `ipc_template_consumer` or if RPC then `block_template::consumer`

PR #214, opened 2025-06-18
bead_sync protocol

ABD-AZE:
`Request response` functionality for the purpose of `IBD`(inital beed download) along with that additionally for P2P queries .

### DIFF
--- a/node/src/bead/mod.rs
+++ b/node/src/bead/mod.rs
@@ -1,10 +1,16 @@
 use crate::committed_metadata::CommittedMetadata;
 use crate::uncommitted_metadata::UnCommittedMetadata;
+use crate::utils::{hashset_to_vec_deterministic, BeadHash};
+use async_trait::async_trait;
 use bitcoin::consensus::encode::Decodable;
 use bitcoin::consensus::encode::Encodable;
 use bitcoin::consensus::Error;
 use bitcoin::io::{self, BufRead, Write};
 use bitcoin::BlockHeader;
+use libp2p::futures::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
+use libp2p::request_response::Codec;
+use libp2p::StreamProtocol;
+use std::collections::HashSet;
 
 #[derive(Clone, Debug, PartialEq)]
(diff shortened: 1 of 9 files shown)


node/src/bead/mod.rs
```diff
@@ -36,5 +42,233 @@ impl Decodable for Bead {
...
+
+impl Decodable for BeadRequest {
+    fn consensus_decode<D: BufRead + ?Sized>(d: &mut D) -> Result<Self, Error> {
+        let request_type = u8::consensus_decode(d)?;
+        match request_type {
+            0 => {
+                let count = u32::consensus_decode(d)?;
+                let mut hashes = HashSet::new();
+                for _ in 0..count {
+                    let hash = BeadHash::consensus_decode(d)?;
```
Sansh2356, line 99: The PR #207 is introducing the `index` based mapping to deal with instead of `beadhashes` present in the `braid` struct so this will be further reverted to `usize::consensus_decode` . But that can be done afterwards since 207 will require a longer time for fully testing to take place .

node/src/behaviour/tests.rs
```diff
@@ -0,0 +1,282 @@
...
+                                // Handle bead response
+                                request_response::Message::Response {
+                                    request_id: _,
+                                    response,
+                                } => {
+                                    if let BeadResponse::Error(error) = response {
+                                        println!("Swarm2: Received error response: {}", error);
+                                        assert_eq!(error, "Bead retrieval not implemented yet");
+                                        break;
+                                    } else if let BeadResponse::Beads(beads) = response {
```
Sansh2356: Since both the `swarms` are responding to a `beadRequest` in form of an error as of now this `else if` branch will never be accessed so it can be taken back .

Sansh2356: Other than that looks good to me for merging further will move onto the `handler` implementations that will require deeper discussion .

Sansh2356 (approved):
LGTM .

mcelrath (approved):
This looks very good, merging.

We need to prioritize performing a functional test with actual bead data as soon as we are able.

PR #216, opened 2025-06-26
Adding floodsub to p2p network stack and changing the `kad` configuration for querying the boot node.

Sansh2356:
- Adding the `floodsub` support for broadcast . 
- Changing `kad` config for boot node querying for nearest peers .


Sansh2356:
The retrieval of keys from the `keystore` will work as it should in case of production env but while testing the network stack locally we would be requiring to uncomment the loading of keystore and adding a `keypair` alternatively . It can be solved by using the current process running on the OS and if multiple instances are there than spawning a new key instead of loading the same key again for multiple nodes to be spawned on local node but for that an additional crate for fetching the process will be required but that can be added in `dev-dependency` . @mcelrath your view on this .

Sansh2356:
On the `graceful shutdown` of the braid-node we currently have implementation without `CancellationToken`, can be seen here [https://tokio.rs/tokio/topics/shutdown](url) . Should we add it as of now or leave it for afterwards ?, @mcelrath .

ABD-AZE:
LGTM

node/src/behaviour/mod.rs
```diff
@@ -16,6 +17,7 @@ use std::{error::Error, time::Duration};
 pub const KADPROTOCOLNAME: StreamProtocol = StreamProtocol::new("/braidpool/kad/1.0.0");
 pub const IDENTIFYPROTOCOLNAME: StreamProtocol = StreamProtocol::new("/braidpool/identify/1.0.0");
 pub const BEAD_SYNC_PROTOCOL: StreamProtocol = StreamProtocol::new("/braidpool/bead-sync/1.0.0");
+pub const FLOODSUBPROTOCOLNAME: StreamProtocol = StreamProtocol::new("/floodsub/1.0.0");
```
mcelrath: This should probably be something like:
```
pub const BEAD_ANNOUNCE_PROTOCOL: StreamProtocol = StreamProtocol::new("/braidpool/bead_announce/1.0.0");
```
and similarly for the other constants that end in "NAME" above.

node/src/behaviour/mod.rs
```diff
@@ -40,6 +42,7 @@ pub struct BraidPoolBehaviour {
     pub identify: identify::Behaviour,
     pub ping: ping::Behaviour,
     pub bead_sync: request_response::Behaviour<BeadCodec>,
+    pub flood_sub: floodsub::Floodsub,
```
mcelrath: again please name this `bead_announce` or something similar.

node/src/main.rs
```diff
@@ -83,7 +86,8 @@ async fn main() -> Result<(), Box<dyn Error>> {
             }
         }
     }
-
+    //for local testing comment this loading of keypair from keystore
```
mcelrath, line 89: Can we please create unit test file for this code that does exactly this?

node/src/main.rs
```diff
@@ -130,7 +140,11 @@ async fn main() -> Result<(), Box<dyn Error>> {
     )
     .parse()
     .expect("Failed to create multiaddress");
-
+    //subscribing to the braidpool topic for broadcasting bead_found and other peer_communications belonging to a particular mesh
```
mcelrath: floodsub doesn't use meshes, right?

mcelrath (changes requested):

node/src/behaviour/mod.rs
Sansh2356: ACK .

mcelrath:
> The retrieval of keys from the `keystore` will work as it should in case of production env but while testing the network stack locally we would be requiring to uncomment the loading of keystore and adding a `keypair` alternatively . It can be solved by using the current process running on the OS and if multiple instances are there than spawning a new key instead of loading the same key again for multiple nodes to be spawned on local node but for that an additional crate for fetching the process will be required but that can be added in `dev-dependency` . @mcelrath your view on this .

I think this will be a systematic issue moving forward and requires a systematic solution. We'll need mock keys for unit tests. Can we create a mock keystore for unit tests, so the code path in the main code is not different between testing and production?

node/src/main.rs
Sansh2356: It uses a single mesh or no mesh can also be said, 
simply comment must be cropped to topic instead of mesh ACK .

Sansh2356:
> > The retrieval of keys from the `keystore` will work as it should in case of production env but while testing the network stack locally we would be requiring to uncomment the loading of keystore and adding a `keypair` alternatively . It can be solved by using the current process running on the OS and if multiple instances are there than spawning a new key instead of loading the same key again for multiple nodes to be spawned on local node but for that an additional crate for fetching the process will be required but that can be added in `dev-dependency` . @mcelrath your view on this .
> 
> I think this will be a systematic issue moving forward and requires a systematic solution. We'll need mock keys for unit tests. Can we create a mock keystore for unit tests, so the code path in the main code is not different between testing and production?

Yes this way can also be used to distinguish between both .

mcelrath:
> On the `graceful shutdown` of the braid-node we currently have implementation without `CancellationToken`, can be seen here [https://tokio.rs/tokio/topics/shutdown](url) . Should we add it as of now or leave it for afterwards ?, @mcelrath .

This looks like a good idea but `CancellationToken` is specific to tokio, no? We will need to cancel network tasks but also shut down cmempoold, write the state to several db's (beads, peers, etc), so the scope of "graceful shutdown" is larger than just the network stack. Can this `CancellationToken` be used for e.g. db shutdown? Is it the right tool for that or is there a better approach?

mcelrath (approved):
Below is a Copilot review, I asked it specifically to suggest new tests. I'm approving this but @Sansh2356 I'd like you to add creating more tests to your TODO list, specifically for the non-happy path. Not all of the suggestions below are relevant, but many are. "Duplicate messages" is probably the most important since we can basically expect to receive the same bead through multiple channels almost 100% of the time.

Read through it, tested it, looks good to me!

----

The diff you provided shows significant changes to the test module in `node/src/behaviour/tests.rs`, primarily introducing a new test: `test_floodsub_message_propagation`, and refactoring existing code to use a new `bead_announce` field rather than `flood_sub`.

Let's break down your questions:

---

## 1. **Are the tests here effective coverage of this new code?**

### **What Changed in the Code**

- The `BraidPoolBehaviour` struct replaces the `flood_sub` field with `bead_announce` and the associated protocol constant.
- The logic for constructing and subscribing to the floodsub topic now uses `bead_announce`.
- A new test, `test_floodsub_message_propagation`, has been added.

### **What the New Test Covers**

`test_floodsub_message_propagation`:

- **Swarm Setup**: Creates two swarms (nodes) and ensures both listen on different ports.
- **Topic Subscription**: Subscribes both swarms using `bead_announce` to a common topic.
- **Connection**: Dials one swarm from the other, establishing a connection.
- **Partial View**: Adds each other's peer ID to the partial view for announcement.
- **Publishing**: One swarm publishes a serialized `Bead` to the topic.
- **Receiving**: The other swarm listens for the message, deserializes it, and asserts that the received data matches the original.

**This test verifies:**

- That the `bead_announce` field is correctly hooked up and functional.
- That message propagation over floodsub works as intended.
- That serialization and deserialization of the announced struct (`Bead`) work as expected.
- That network events and peer connections are handled robustly.

### **Does it Cover the Changes?**

- **Field usage**: Yes, it exercises the new `bead_announce` field directly.
- **Protocol change**: Yes, the new protocol constant is used for topic/subscription.
- **Round-trip bead propagation**: Yes, it covers the core functionality of floodsub-based bead announcement.

**Limitations:**

- The test is integration-style and covers the happy path of bead announcement and reception.
- It does not cover error cases (failed deserialization, duplicate subscriptions, network failures).
- It does not test edge cases such as rapid re-subscription, unsubscribing, or multiple topics.
- It does not test what happens if a message is malformed or if an unexpected peer connects.

---

## 2. **Other Tests You Might Suggest**

Here are some additional tests that would further strengthen coverage:

### **a) Negative/Edge Cases**

- **Malformed message**: Publish a message that cannot be deserialized as a `Bead`, and assert that the receiver handles this gracefully (no panic/crash).
- **Unsubscribed topic**: Publish on a topic to which the receiver is not subscribed, verify that it is not delivered.
- **Duplicate message**: Publish the same bead twice and ensure the receiver handles duplicates appropriately (if deduplication is required).
- **Network failure**: Simulate a connection drop during propagation, ensure the system behaves as expected.

### **b) Protocol/Peer Management**

- **Unsubscribe and resubscribe**: Have a peer unsubscribe and then resubscribe, ensure they receive messages only when subscribed.
- **Multiple topics**: Subscribe to multiple topics, send messages on both, ensure only the correct topic's messages are received.
- **Multiple receivers**: Subscribe multiple peers to the topic, verify that all receive the message.

### **c) Behavioural**

- **Floodsub partial view churn**: Remove and re-add peers to the partial view, ensure correct message delivery.
- **Concurrent publishing**: Publish from multiple swarms at once, check for message loss/race conditions.

### **d) Performance/Stress**

- **High volume**: Send a high number of bead messages in rapid succession, ensure the system remains stable.

---

## **Summary**

- **Current tests**: The new test gives good coverage for the most important "happy path" of the new code and the protocol/field change.
- **Suggested additions**: Add tests for error handling, edge cases, multiple peers/topics, and stress scenarios for robust coverage.

If you want specific code examples for any of the suggested tests, let me know!

PR #219, opened 2025-06-28
peer manager and bead channel

ABD-AZE:


Sansh2356 (dismissed):
LGTM @ABD-AZE , regarding the `spawn::blocking` we can follow its removal if you want that to go ahead due to the reasons we have discussed on the call, we can persist with this also . Other than that the `channel` handling looks fine as per as the size of channel is concerned @mcelrath what do you have in mind regarding this ? .

node/Cargo.toml
```diff
@@ -23,6 +23,7 @@ bitcoincore-rpc = { version = "0.19.0"}
 bitcoincore-rpc-json = { version = "0.19.0"}
 bitcoincore-zmq = { version = "1.5.2", git = "https://github.com/antonilol/rust-bitcoincore-zmq.git", features = ["async"] }
 env_logger = "0.11"
+fastrand = "2.0.1"
```
mcelrath: This doesn't appear to be used?

Also generally prefer standard libraries over new dependencies.

node/src/braid/mod.rs
```diff
@@ -33,6 +35,17 @@ impl Braid {
             cohorts: vec![Cohort(genesis_beads.clone())],
             orphan_beads: Vec::new(),
             genesis_beads: genesis_beads,
+            request_tx: tokio::sync::mpsc::channel(1).0, // Default sender, will be set later
```
mcelrath: I don't think this belongs in the Braid class.

The Braid is a singleton object, and a data object. I don't want to populate it with network dependencies. The node should be responsible for updating the Braid and adding data to it.

Node is where all the network dependencies should live.

node/src/main.rs
```diff
@@ -343,17 +360,21 @@ async fn main() -> Result<(), Box<dyn Error>> {
                                     request_id,
                                     channel,
                                 } => {
+                                    // Handle the bead sync request here
                                     match request {
                                         bead::BeadRequest::GetBeads(hashes) => {
-                                            let beads = Vec::new();
+                                            // Get the beads from the local store
```
mcelrath: The "local store" is the braid object. Use its data to respond to requests.

node/src/main.rs
```diff
@@ -343,17 +360,21 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                                            let beads = Vec::new(); // Replace with actual logic to fetch beads
                                             swarm.behaviour_mut().respond_with_beads(channel, beads);
                                         }
                                         bead::BeadRequest::GetTips => {
-                                            let tips = HashSet::new();
+                                            // Get the tips from the local store
```
mcelrath: Use the Braid, Luke.

node/src/main.rs
```diff
@@ -363,28 +384,89 @@ async fn main() -> Result<(), Box<dyn Error>> {
                                     response,
                                 } => {
                                     match response {
+                                        // might consider spawning this task on the blocking threadpool if the ex. time is more
                                         bead::BeadResponse::Beads(beads) => {
-                                            // Handle beads
+                                            tokio::task::spawn_blocking({
+                                                || {
+                                                    // let mut lock = braid.lock().await;
```
mcelrath: What's this supposed to do?

mcelrath (changes requested):
I think we need to rebase this off the `replace-beadhash-with-bead` branch, which implements all the braid functions. That way we can fill in the network messages to send/receive beads, tips, genesis, etc, which are currently stubs in this PR.

node/src/braid/mod.rs
ABD-AZE: How should we request beads from the swarm then? I mean the logic for which beads are missing would be written within the braid, to request those missing beads there has to be some way to communicate between the braid and the swarm.

node/src/main.rs
ABD-AZE: There were changes to be done in the braid when I put this PR so I was not sure whether the same methods will be used to search the beads using the beadhash. Anyways I'll make the changes now.

node/src/main.rs
ABD-AZE: The braid would be shared between multiple threads, so I have used arc<mutex<>>, this line acquires the lock.

ABD-AZE: This part of the code also depends upon the braid module. Handling the responses from peers includes performing checks on the response received. For eg: for requesting genesis, response from multiple peers will be compared, in case of a false response from a peer, the score of that peer is updated in the peer manager. For that we need to have some methods in braid which returns the status of response received. Based on the response the peer manager is updated.

node/src/braid/mod.rs
mcelrath: Well there are two ways:
1. During IBD, peers should be streaming a bunch of beads to us, we shouldn't be requesting them individually. If we need to re-request one due to e.g. network data corruption, we can do so by bead hash.
2. We received a New Bead broadcast on floodsub, and are not yet caught up, or the bead has missing parents.

Now during IBD we can use #1 only and just wait for our peers to stream all beads to us. We *could* request parents of newly-seen beads and fill in the history *backwards* from the tip as well. I don't think we should do this though since it's a lot of round trips. We're already asking peers for beads. We can do that forwards or backwards. If we want to do it backwards we should add a dedicated way to do that instead of requesting beads one at a time from peers.

So basically for #1 we have to find a peer willing to do IBD and stream *everything* to us. This isn't a single bead request and needs to be a separate RequestResponse handler and flow.

node/src/bead/mod.rs
```diff
@@ -84,6 +87,9 @@ impl Encodable for BeadRequest {
             BeadRequest::GetGenesis => {
                 2u8.consensus_encode(writer) // 2 for GetGenesis
             }
+            BeadRequest::GetAllBeads => {
+                3u8.consensus_encode(writer) // 3 for GetAllBeads
```
mcelrath: Sicnce 1u8, 2u8, 3u8 are really "messages" we should be using an enum or pub const so we can use an alphanumeric name here instead of an opaque number.

node/src/bead/mod.rs
```diff
@@ -103,6 +109,7 @@ impl Decodable for BeadRequest {
             }
             1 => Ok(BeadRequest::GetTips),
             2 => Ok(BeadRequest::GetGenesis),
+            3 => Ok(BeadRequest::GetAllBeads),
```
mcelrath: Need to use a symbolic name here instead of 1,2,3.

node/src/bead/mod.rs
```diff
@@ -126,7 +133,7 @@ impl Encodable for BeadResponse {
             BeadResponse::Tips(tips) => {
                 let mut written = 0;
                 written += 1u8.consensus_encode(writer)?; // 1 for Tips
-                let tips_vec = hashset_to_vec_deterministic(tips);
+                let tips_vec = tips;
```
mcelrath: Is this local variable really necessary?

node/src/bead/mod.rs
```diff
@@ -136,16 +143,25 @@ impl Encodable for BeadResponse {
             BeadResponse::Genesis(genesis) => {
                 let mut written = 0;
                 written += 2u8.consensus_encode(writer)?; // 2 for Genesis
-                let genesis_vec = hashset_to_vec_deterministic(genesis);
+                let genesis_vec = genesis;
```
mcelrath: Is this local variable really necessary?

node/src/bead/mod.rs
```diff
@@ -136,16 +143,25 @@ impl Encodable for BeadResponse {
...
                 written += (genesis_vec.len() as u32).consensus_encode(writer)?;
                 for hash in genesis_vec {
                     written += hash.consensus_encode(writer)?;
                 }
                 Ok(written)
             }
+            BeadResponse::GetAllBeads(beads) => {
+                let mut written = 0;
+                written += 3u8.consensus_encode(writer)?; // 3 for GetAllBeads
```
mcelrath: Use alphanumeric name instead of 3u8.

node/src/main.rs
```diff
@@ -175,216 +189,322 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-                                .iter()
-                                .any(|p| *p == BEAD_ANNOUNCE_PROTOCOL)
-                            {
-                                log::info!("PEER ADDED TO FLOODSUB MESH {:?}", peer_id);
-                                for addr in info_reference.clone().listen_addrs {
-                                    swarm
-                                        .behaviour_mut()
-                                        .bead_announce
-                                        .add_node_to_partial_view(peer_id);
+                                bead::BeadRequest::GetTips => {
```
mcelrath, line 416: This shouldn't require a lock -- we're not modifying the Braid.

node/src/main.rs
```diff
@@ -175,216 +189,322 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                                        .tips
+                                        .iter()
+                                        .filter_map(|index| braid_lock.beads.get(*index))
+                                        .cloned()
+                                        .map(|bead| bead.block_header.block_hash())
+                                        .collect();
+                                    swarm.behaviour_mut().respond_with_tips(channel, tips);
+                                }
+                                bead::BeadRequest::GetGenesis => {
+                                    let braid_lock = braid.lock().await;
```
mcelrath: This shouldn't require a lock.

node/src/main.rs
```diff
@@ -175,216 +189,322 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-                                        }
-                                        bead::BeadRequest::GetTips => {
-                                            let tips = HashSet::new();
-                                            swarm.behaviour_mut().respond_with_tips(channel, tips);
-                                        }
-                                        bead::BeadRequest::GetGenesis => {
-                                            let genesis = HashSet::new();
-                                            swarm.behaviour_mut().respond_with_genesis(channel, genesis);
+                            match response {
+                                bead::BeadResponse::Beads(beads) => {
```
mcelrath: This code is duplicate with BeadResponse::GetAllBeads.

node/src/main.rs
```diff
@@ -175,216 +189,322 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-                                        bead::BeadResponse::Genesis(genesis) => {
-                                            // Handle genesis
+                                        braid::GenesisCheckStatus::GenesisBeadsCountMismatch => {
+                                            log::warn!("Genesis beads count mismatch");
                                         }
-                                        bead::BeadResponse::Error(error) => {
-                                            // Handle error
+                                    }
+                                }
+                                bead::BeadResponse::GetAllBeads(beads) => {
```
mcelrath: Use:
```
bead::BeadResponse::Beads(beads) | bead::BeadResponse::GetAllBeads(beads) => {
```
instead of duplicated code.

mcelrath (changes requested):
Minor suggestions below.

I don't like all the explicit lock handling here. It's a recipe for bugs. I'd like to find a more idiomatic and automatic way to do this.

What we actually want is a single-writer, multiple-reader, type of lock. So functions like GetTips and GetGenesis do not have to worry about locks at all, but block if there is a writer.

The only writer is extend(), and *ideally* any function which needs to modify the braid acquires a mutable reference, which automatically acquires a write lock, and drops the write lock when the reference goes out of scope.

OR can we put the write lock inside extend() itself? But I guess that means everyone needs a mutable reference to the Braid.

Looking for some good ideas here to reduce the amount of "boilerplate" involved in using the Braid object.

ABD-AZE:
> Minor suggestions below.
> 
> I don't like all the explicit lock handling here. It's a recipe for bugs. I'd like to find a more idiomatic and automatic way to do this.
> 
> What we actually want is a single-writer, multiple-reader, type of lock. So functions like GetTips and GetGenesis do not have to worry about locks at all, but block if there is a writer.
> 
> The only writer is extend(), and *ideally* any function which needs to modify the braid acquires a mutable reference, which automatically acquires a write lock, and drops the write lock when the reference goes out of scope.
> 
> OR can we put the write lock inside extend() itself? But I guess that means everyone needs a mutable reference to the Braid.
> 
> Looking for some good ideas here to reduce the amount of "boilerplate" involved in using the Braid object.

We can handle this situation by moving the lock inside the braid object. During the extend function we can acquire the write lock. And during the rest of the functions which read the braid, we'll acquire the read lock.
But again does a lock being part of the braid structure look weird? If not then we can go ahead with this approach otherwise explicit handling of lock is the only solution.

Sansh2356 (dismissed):
Looks good to me after the `RwLock` discussion we held .

mcelrath (approved):

PR #220, opened 2025-06-28
Consensus functions implementation native to braidpool along with unit-tests  

Sansh2356:
This PR will do the `rust side` implementation of consensus functionalities inherit to `braidpool` the braid functions in the rust.


Aritra8438 (dismissed):
Ideally we should ponder how to make these functions more optimized, especially the `extend()` function.
But, as this PR is veryyy important, we should merge this for subsequent development, and we can address those changes in further PRs, that's my opinion.

I would also like to list a few of them.
- [ ] Add `bead <-> cohort` mapping.
- [ ] Add `parents` and `children` props to the braid class.

Sansh2356:
> Ideally we should ponder how to make these functions more optimized, especially the `extend()` function. But, as this PR is veryyy important, we should merge this for subsequent development, and we can address those changes in further PRs, that's my opinion.
> 
> I would also like to list a few of them.
> 
>     * [ ]  Add `bead <-> cohort` mapping.
> 
>     * [ ]  Add `parents` and `children` props to the braid class.

Agreed completely @Aritra8438 , we will subsequently address these optimizations issues but before that we can make the other higher priority tasks done and build optimizations further on those tasks , as far as the `parents` and `children` attribute addition is concerned i agree with that also, but for the `bead` to `cohort` mapping i am not to sure what are you pointing out is it `Cohorts` mapped with `start_bead` ? .

Aritra8438:
My plan was to store a mapping between the beads and the cohort it is in.
Currently, we are looking up the cohort for a bead in `o(n)`, if we keep this, the look up will take constant time, though we need to think of an optimized updation.

node/src/braid/mod.rs
```diff
@@ -118,32 +165,866 @@ impl Braid {
...
+            return Ordering::Less;
+        }
+        if dwork[&a] > dwork[&b] {
+            return Ordering::Greater;
+        }
+
+        let awork = match awork_or_not {
+            Some(val) => val,
+            None => {
+                panic!("No ancestor work supplied");
```
mcelrath: Panics in this file will terminate the node. 

Can we come up with some reasonable error handling instead?

node/src/braid/mod.rs
```diff
@@ -1,11 +1,13 @@
...
+use std::collections::{HashMap, HashSet};
+pub const FIXED_BEAD_WORK: u32 = 1;
+#[derive(Clone, Debug, Serialize, PartialEq)]
 
-#[derive(Clone, Debug, Serialize)]
-
-pub(crate) struct Cohort(HashSet<BeadHash>);
+pub(crate) struct Cohort(HashSet<usize>);
 
 pub enum AddBeadStatus {
```
mcelrath, line 14: This enum isn't actually used by Braid::extend.

node/src/braid/mod.rs
```diff
@@ -17,99 +19,144 @@ pub enum AddBeadStatus {
...
         for (i, cohort) in self.cohorts.iter().enumerate().rev() {
-            // Find which parents are in this cohort
-            for parent in &bead.committed_metadata.parents {
-                if cohort.0.contains(parent) {
-                    found_parents.insert(*parent);
+            // Find which parent indices are in this cohort
+            for parent_hash in &bead.committed_metadata.parents {
+                if let Some(parent_index) = self
+                    .beads
+                    .iter()
```
mcelrath: This iterates over all beads so is O(n) and there may be millions of beads.

Use 
```
if let Some(&parent_index) = self.bead_index_mapping.get(parent_hash)
```

node/src/braid/mod.rs
```diff
@@ -17,99 +19,144 @@ pub enum AddBeadStatus {
...
         } else {
             self.cohorts.clear();
         }
 
+        // Remove parents from tips if present
+        for parent_hash in &bead.committed_metadata.parents {
+            // Find the index of the parent bead
+            if let Some(parent_index) = self
+                .beads
+                .iter()
```
mcelrath: This iterates over all beads so is O(n) and there may be millions of beads.

Use 
```
if let Some(&parent_index) = self.bead_index_mapping.get(parent_hash)
```

node/src/braid/mod.rs
```diff
@@ -118,32 +165,866 @@ impl Braid {
...
+                    for bead in braid_tips.difference(&cohort) {
+                        tail.insert(*bead);
+                    }
+                } else {
+                    for t in &cohort {
+                        tail.remove(t);
+                    }
+                }
+                oldcohort = cohort.clone();
+                let mut key_set: HashSet<usize> = HashSet::new();
```
mcelrath: Use:
```
let key_set: HashSet<usize> = ancestor.keys().cloned().collect();
```
instead

node/src/braid/mod.rs
```diff
@@ -17,99 +19,144 @@ pub enum AddBeadStatus {
...
-            .parents
-            .iter()
-            .all(|p| self.beads.contains(p))
-        {
-            return false;
+        for parent_hash in &bead.committed_metadata.parents {
+            // Check if we already have this parent in our beads
+            let parent_exists = self
+                .beads
+                .iter()
```
mcelrath: use `self.bead_index_mapping.contains_key(parent_hash)` instead

node/src/braid/mod.rs
```diff
@@ -17,99 +19,144 @@ pub enum AddBeadStatus {
...
+                    return false;
+                }
+            }
         }
         // Already seen this bead
         let bead_hash = bead.block_header.block_hash();
-        if self.beads.contains(&bead_hash) {
+        if self
+            .beads
+            .iter()
```
mcelrath, line 83: use `self.bead_index_mapping.contains_key(&bead_hash)` instead

node/src/braid/mod.rs
```diff
@@ -118,32 +165,866 @@ impl Braid {
...
+    ) -> &'a mut HashMap<usize, HashSet<usize>> {
+        let mut dequeue: VecDeque<(usize, bool)> = VecDeque::new();
+        let current_bead_index = braid_obj.bead_index_mapping[&current_block_hash];
+        dequeue.push_back((current_bead_index, false));
+        while dequeue.len() > 0 {
+            let (current, is_processed) = dequeue[dequeue.len() - 1];
+
+            if is_processed == true {
+                dequeue.pop_back();
+                if let Some(current_ancestor) = ancestors.get_mut(&current) {
```
mcelrath, line 345: How about:

```

  let current_ancestor = ancestors.entry(current).or_default();
  current_ancestor.clear();
  if let Some(parents_beads) = parents.get(&current) {
      current_ancestor.extend(parents_beads.iter());
  }
```
which only does one HashMap lookup using `entry()` instead of two (`get_mut()` and `insert()`).

node/src/braid/mod.rs
```diff
@@ -1,11 +1,13 @@
 use crate::bead::Bead;
-use crate::utils::BeadHash;
-use ::serde::Serialize;
-use std::collections::HashSet;
+use crate::utils::{retrieve_bead, BeadHash};
+use serde::Serialize;
+use std::cmp::Ordering;
+use std::collections::VecDeque;
+use std::collections::{HashMap, HashSet};
+pub const FIXED_BEAD_WORK: u32 = 1;
```
mcelrath: "work" is a 256-bit quantity. It's an estimate of the number of sha256d hashes performed and is 1/target.

This misunderstanding appears several times in this file...

node/src/braid/mod.rs
```diff
@@ -118,32 +165,866 @@ impl Braid {
...
+                    }
+                } else {
+                    let mut val_set: HashSet<usize> = HashSet::new();
+                    if let Some(parents_beads) = parents.get(&current) {
+                        val_set.extend(parents_beads);
+                    }
+                    ancestors.insert(current, val_set);
+                }
+                if let Some(parent_indices) = parents.get(&current) {
+                    let ancestor_ref = ancestors.to_owned();
```
mcelrath, line 358: This clones the entire ancestors HashMap. Can that be avoided?

node/src/braid/mod.rs
```diff
@@ -17,99 +19,144 @@ pub enum AddBeadStatus {
...
+    pub fn new(genesis_beads: HashSet<Bead>) -> Self {
+        let mut beads = Vec::new();
+        let mut bead_indices = HashSet::new();
+        let mut bead_index_mapping = HashMap::new();
 
-    pub fn add_bead(&mut self, bead: Bead) -> AddBeadStatus {
-        if self.beads.contains(&bead.block_header.block_hash()) {
-            return AddBeadStatus::DagAlreadyContainsBead;
+        for (index, bead) in genesis_beads.into_iter().enumerate() {
+            beads.push(bead.clone());
```
mcelrath, line 40: This clone() is unnecessary.

node/src/braid/tests.rs
```diff
@@ -1 +1,1438 @@
...
+            for parent_beads in beads.1 {
+                current_bead_parents.insert(parent_beads);
+            }
+            current_braid_parents.insert(beads.0, current_bead_parents);
+        }
+        let reversed_beads = reverse(&current_file_braid, &current_braid_parents);
+        let current_braid_cohorts = file_braid.cohorts;
+
+        let computed_val = cohort(&current_file_braid, &reversed_beads, None, None);
+        //TODO:assetion to be done
```
mcelrath, line 1074: What is to be done here?

node/src/braid/mod.rs
```diff
@@ -17,99 +19,144 @@ pub enum AddBeadStatus {
...
+                .beads
+                .iter()
+                .any(|b| b.block_header.block_hash() == *parent_hash);
+
+            if !parent_exists {
+                // Try to retrieve the parent
+                if let Some(retrieved_bead) = retrieve_bead(*parent_hash) {
+                    self.extend(&retrieved_bead);
+                } else {
+                    // Parent not found and can't be retrieved
```
mcelrath, line 73: This is a silent failure. Here we need to be adding this bead to `orphan_beads`.

At the end of `extend()` we need to be processing anything in `orphan_beads` in case a missing parent was just added.

node/src/braid/mod.rs
```diff
@@ -118,32 +165,866 @@ impl Braid {
 
         true
     }
```
mcelrath, line 158: After successfully extending the braid, we need to process any `orphan_beads` if nonempty.

node/src/braid/mod.rs
```diff
@@ -118,32 +165,866 @@ impl Braid {
...
+        let work: HashMap<usize, u32> = parents.keys().map(|&k| (k, FIXED_BEAD_WORK)).collect();
+        let bead_work = match bead_work_or_not {
+            Some(val) => val,
+            None => &work,
+        };
+        let mut previous_work: u32 = 0;
+        let rev_cohorts = match in_cohorts_or_not {
+            Some(val) => {
+                let mut val_ref = val;
+                val_ref.reverse();
```
mcelrath, line 748: This modifies the input parameter.

node/src/braid/mod.rs
```diff
@@ -118,32 +165,866 @@ impl Braid {
...
+    pub fn updating_ancestors<'a>(
+        braid_obj: &Braid,
+        current_block_hash: BeadHash,
+        ancestors: &'a mut HashMap<usize, HashSet<usize>>,
+        parents: &HashMap<usize, HashSet<usize>>,
+    ) -> &'a mut HashMap<usize, HashSet<usize>> {
+        let mut dequeue: VecDeque<(usize, bool)> = VecDeque::new();
+        let current_bead_index = braid_obj.bead_index_mapping[&current_block_hash];
+        dequeue.push_back((current_bead_index, false));
+        while dequeue.len() > 0 {
```
mcelrath: A more rusty idiomatic way would be:
```
while let Some((current, is_processed)) = dequeue.pop_back()
```

mcelrath (changes requested):

node/src/braid/tests.rs
Sansh2356: Assertion is left it has been mentioned .

node/src/braid/mod.rs
Sansh2356: ACK.

node/src/braid/mod.rs
Sansh2356: This is for the type definition of `cohort` type it is being modified by `extend` but by the `self` object since `extend` is a class based defined function .

node/src/braid/mod.rs
Sansh2356: Here `cloning` is being used which might be expensive in terms of memory that's why i had opted for reference instead .

Sansh2356: ACK .

node/src/braid/mod.rs
Sansh2356: This had to be done previously was in mind truly just missed it .

node/src/braid/mod.rs
Sansh2356: Can be done but IMO it provides more readability over the above no change in other terms though since the above and latter both proves to be same TC and SC, since lookup will be O(1) but it will come down to O(N) of `clear` function .

node/src/braid/mod.rs
Sansh2356: This is the only portion in which `cloning` is required due to the `lifetime` issue as the `ancestor` value is being accessed both outside as well as inside the loop thus when it is passed to the other function it's ownership gets borrowed leading to `ancestor does not long enough` if it is ownership is not cloned, i think this is manageable .

node/src/braid/mod.rs
Sansh2356: Yes it needs to reverse the order of the input .

node/src/braid/mod.rs
```diff
@@ -17,99 +21,134 @@ pub enum AddBeadStatus {
...
         // Don't have all parents
-        if !bead
-            .committed_metadata
-            .parents
-            .iter()
-            .all(|p| self.beads.contains(p))
-        {
-            return false;
+        for parent_hash in &bead.committed_metadata.parents {
+            self.bead_index_mapping.contains_key(parent_hash);
```
mcelrath: line is duplicated...

node/src/braid/mod.rs
```diff
@@ -118,32 +157,860 @@ impl Braid {
...
+        awork_or_not: Option<&HashMap<usize, BigUint>>,
+    ) -> Result<Ordering, BraidError> {
+        if dwork[&a] < dwork[&b] {
+            return Ok(Ordering::Less);
+        }
+        if dwork[&a] > dwork[&b] {
+            return Ok(Ordering::Greater);
+        }
+
+        let awork = awork_or_not.ok_or(MissingAncestorWork)?;
```
mcelrath: I don't understand, why would this ever be "missing"? It might be zero (genesis) but why missing? Missing parents? (in which case we should return an error if either bead is not in the braid).

mcelrath (dismissed):

node/src/braid/mod.rs
Sansh2356: Yes it should ideally be the case making it non-optional since without ancestor_work existence of the valid bead will always be false hence no suitable logic behind keeping it optional .

mcelrath (approved):
Great!

PR #222, opened 2025-07-02
Implement rpc server

Aritra8438:
## Please don't merge this.

This is a very primitive design for the rpc server.
- It introduces a `braidmanager` class which consists of the `braid`, `beads`, `commands`.
- Please note that. we can remove the beads and that's our next goal. The `beads` property was needed when we had only the `beadhash`.
- Below we will track a list of commands that will be exposed through the `rpc` server.

- [x] gettips
- [x] getbead
- [x] ~getcohortcount~ getcohortscount
- [x] ~getbeadcount~ getbeadscount

If you have any more suggestions, feel free to add!


node/src/main.rs
```diff
@@ -39,9 +46,67 @@ const ADDR_REFRENCE: &str =
...
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error>> {
+    let (sender, mut receiver): (Sender<String>, Receiver<String>) = mpsc::channel(32);
+
     let args = cli::Cli::parse();
+    if args.rpc != None {}
     setup_logging();
     setup_tracing()?;
+
+    // Initialize braid with empty genesis beads initially
```
mcelrath, line 56: What we should do here is either:
1. Receive a genesis bead from peers, do IBD, and decide we're "up to date".
2. If we're the first node or isolated, get the tip from Bitcoind/cmempoold and use that as our genesis.

node/src/braid_manager.rs
```diff
@@ -0,0 +1,91 @@
...
+    },
+    GetBeadCount {
+        respond_to: oneshot::Sender<usize>,
+    },
+    GetCohortCount {
+        respond_to: oneshot::Sender<usize>,
+    },
+}
+
+pub struct BraidManager {
```
mcelrath, line 28: I don't think we need a "manager" here. There will only be one braid in general, and having two will be rare (at difficulty retargets, or if we do sub-braids).

The below duplicates storage -- both the Braid class and bead_storage duplicate the storage of beads.

Instead we should put RPC handler methods in the Node class, with access to the Braid object.

node/src/main.rs
```diff
@@ -39,9 +46,67 @@ const ADDR_REFRENCE: &str =
...
+    // Initializing the braid_manager also initializes the braid
+    let genesis_beads = HashSet::new();
+    let (braid_manager, braid_tx) = BraidManager::new(genesis_beads);
+
+    tokio::spawn(async move {
+        braid_manager.run().await;
+    });
+
+    let test_bead1 = create_test_bead(1, None); // Genesis bead
```
mcelrath, line 65: This doesn't belong here, it should be in a test.

node/src/main.rs
```diff
@@ -363,16 +428,16 @@ async fn main() -> Result<(), Box<dyn Error>> {
                                     response,
                                 } => {
                                     match response {
-                                        bead::BeadResponse::Beads(beads) => {
+                                        BeadResponse::Beads(beads) => {
                                             // Handle beads
                                         }
-                                        bead::BeadResponse::Tips(tips) => {
+                                        BeadResponse::Tips(tips) => {
                                             // Handle tips
```
mcelrath, line 435: For all these, call the corresponding Braid object method, and create a proper response.

node/src/rpc_server.rs
```diff
@@ -0,0 +1,350 @@
...
+                let handler_tx_ref = rpc_handler_sender.clone();
+                tokio::spawn(handle_request(
+                    rpc_method.clone(),
+                    method_params,
+                    client_res,
+                    handler_tx_ref,
+                ));
+            }
+        }
+        "getbeadcount" => {
```
mcelrath, line 86: I don't understand how this is actually getting the bead count. It needs to have access to the Braid (which is owned by the Node), and call the Braid method to return the response.

node/src/rpc_server.rs
```diff
@@ -0,0 +1,350 @@
...
+//that is the JSON-RPC handle to initiate the RPC context
+//supporting both http and websockets
+#[rpc(server)]
+pub trait Rpc {
+    //RPC methods supported by braid-API
+    #[method(name = "getbead")]
+    async fn get_bead(&self, bead_hash: String) -> Result<String, ErrorObjectOwned>;
+
+    #[method(name = "addbead")]
+    async fn add_bead(&self, bead_data: String) -> Result<String, ErrorObjectOwned>;
```
mcelrath, line 146: See Braid::extend()

mcelrath (changes requested):
The biggest problem with this PR is that the RPC cannot exist as an independent entity from the Node. It is part of the node and can't do it's job without access to the data held by the node (in particular, the Braid object).

There are a couple of ways to do this:
1. Make the RPC methods be handlers on the Node object, so that they have access to all the node data (Braid, Peers, etc).
2. Pass the relevant immutable object (by reference!) to the RPC object (e.g. Braid&)

Option 1 will make the Node code fairly long and un-modular, but maybe we can separate the RPC handlers in to separate files? Option 2 requires passing a lot of large objects and questions of ownership and lifetime become important.

Sansh2356:
Updated further with a new PR #224  marking this as stale hence closing it .

PR #224, opened 2025-07-09
This is updated `RPC` implementation according to the new structure we had decided 

Sansh2356:
- It includes `rpc` methods accessible both via `cli` as well as from frontend via `curl` or `axios`.
- Unit tests regarding each method is added respectively .
- This PR also involves the replacing of snippets which are common and also the module accessibility structure as well.

### DIFF
--- a/node/src/bead/tests.rs
+++ b/node/src/bead/tests.rs
@@ -14,8 +14,6 @@ use bitcoin::consensus::encode::Encodable;
 use bitcoin::consensus::serialize;
 use bitcoin::consensus::DeserializeError;
 use bitcoin::ecdsa::Signature;
-use bitcoin::p2p::Address as P2P_Address;
-use bitcoin::p2p::ServiceFlags;
 use bitcoin::BlockHash;
 use bitcoin::BlockHeader;
 use bitcoin::BlockTime;
@@ -23,22 +21,22 @@ use bitcoin::BlockVersion;
 use bitcoin::CompactTarget;
 use bitcoin::EcdsaSighashType;
 use bitcoin::TxMerkleNode;
-use core::net::SocketAddr;
 use futures::executor::block_on;
 use libp2p::request_response::Codec;
 use std::collections::HashSet;
 use std::io::Cursor;
-use std::net::{IpAddr, Ipv4Addr};
 use std::str::FromStr;
 #[test]
 
 fn test_serialized_committed_metadata() {
-    let test_sock_add = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8888);
-    let _address = P2P_Address::new(&test_sock_add.clone(), ServiceFlags::NONE);
+    // let test_sock_add = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8888);
+    // let _address = P2P_Address::new(&test_sock_add.clone(), ServiceFlags::NONE);
+    let _address = String::from("127.0.0.1:8000");
     let public_key = "020202020202020202020202020202020202020202020202020202020202020202"
         .parse::<bitcoin::PublicKey>()
         .unwrap();
-    let socket = bitcoin::p2p::address::AddrV2::Ipv4(Ipv4Addr::new(127, 0, 0, 1));
+    // let socket = bitcoin::p2p::address::AddrV2::Ipv4(Ipv4Addr::new(127, 0, 0, 1));
+    let socket = String::from("127.0.0.1");
     let time_val = Time::from_consensus(1653195600).unwrap();
     let parent_hash_set: HashSet<BlockHash> = HashSet::new();
     let time_hash_set = TimeVec(Vec::new());
@@ -102,12 +100,11 @@ fn test_serialized_uncommitted_metadata() {
 #[test]
 
 fn test_serialized_bead() {
-    let test_sock_add = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8888);
-    let _address = P2P_Address::new(&test_sock_add.clone(), ServiceFlags::NONE);
+    let _address = String::from("127.0.0.1:8000");
     let public_key = "020202020202020202020202020202020202020202020202020202020202020202"
         .parse::<bitcoin::PublicKey>()
         .unwrap();
-    let socket = bitcoin::p2p::address::AddrV2::Ipv4(Ipv4Addr::new(127, 0, 0, 1));
+    let socket = String::from("127.0.0.1");
     let time_hash_set = TimeVec(Vec::new());
     let parent_hash_set: HashSet<BlockHash> = HashSet::new();
     let weak_target = CompactTarget::from_consensus(32);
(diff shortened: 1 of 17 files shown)


node/src/braid/tests.rs
```diff
@@ -71,10 +66,12 @@ fn emit_bead() -> Bead {
         .as_secs() as u32;
     let current_time = Time::from_consensus(now).unwrap();
 
-    let test_sock_add = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8888);
-    let _address = P2P_Address::new(&test_sock_add.clone(), ServiceFlags::NONE);
+    // let test_sock_add = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8888);
```
mcelrath: Is this commented out on purpose?

node/src/rpc_server.rs
```diff
@@ -0,0 +1,419 @@
...
+//server building
+//running a server in seperate spawn event
+pub async fn run_rpc_server(braid_shared_pointer: Arc<RwLock<Braid>>) -> Result<SocketAddr, ()> {
+    //Initializing the middleware
+    let rpc_middleware =
+        jsonrpsee::server::middleware::rpc::RpcServiceBuilder::new().layer_fn(LoggingMiddleware);
+    //building the context/server supporting the http transport and ws
+    let server = jsonrpsee::server::Server::builder()
+        .set_rpc_middleware(rpc_middleware)
+        .build("127.0.0.1:8888")
```
mcelrath: Add configuration for this in config.rs

node/src/rpc_server.rs
```diff
@@ -0,0 +1,419 @@
...
+#[tokio::test]
+pub async fn test_extend_rpc() {
+    let test_bead1 = create_test_bead(1, None);
+    let genesis_beads = vec![test_bead1.clone()];
+
+    let braid: Arc<RwLock<braid::Braid>> = Arc::new(RwLock::new(braid::Braid::new(genesis_beads)));
+
+    let _ = run_rpc_server(Arc::clone(&braid)).await.unwrap();
+
+    let server_addr = "127.0.0.1:8888";
```
mcelrath: -> config.rs

mcelrath (commented):
This looks good!

(Still need to test it)

node/src/braid/tests.rs
Sansh2356: Yes this should be removed .

node/src/rpc_server.rs
Sansh2356: ACK

node/src/rpc_server.rs
Sansh2356: ACK

node/src/bead/tests.rs
```diff
@@ -14,31 +14,26 @@ use bitcoin::consensus::encode::Encodable;
...
 use std::collections::HashSet;
 use std::io::Cursor;
-use std::net::{IpAddr, Ipv4Addr};
 use std::str::FromStr;
 #[test]
 
 fn test_serialized_committed_metadata() {
-    let test_sock_add = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8888);
-    let _address = P2P_Address::new(&test_sock_add.clone(), ServiceFlags::NONE);
+    let _address = String::from("127.0.0.1:8000");
```
mcelrath, line 32: Did you do this to remove the networking dependencies in this file? That's probably a good idea.

node/src/braid/mod.rs
```diff
@@ -65,6 +64,27 @@ impl Braid {
     pub fn extend(&mut self, bead: &Bead) -> AddBeadStatus {
         // No parents: bad block
         if bead.committed_metadata.parents.is_empty() {
+            // Check if we already have this bead
+            let bead_hash = bead.block_header.block_hash();
+            if self
+                .beads
+                .iter()
+                .any(|b| b.block_header.block_hash() == bead_hash)
```
mcelrath: This is an O(n) operation and self.beads can be large. Use self.bead_index_mapping instead.

node/src/braid/mod.rs
```diff
@@ -65,6 +64,27 @@ impl Braid {
     pub fn extend(&mut self, bead: &Bead) -> AddBeadStatus {
         // No parents: bad block
         if bead.committed_metadata.parents.is_empty() {
+            // Check if we already have this bead
```
mcelrath: I don't understand this block and I don't think it's correct. We should never be calling extend() with the genesis beads. We should be calling new() with the genesis beads. 

So I think the original code which just returns InvalidBead is correct.

node/src/rpc_server.rs
```diff
@@ -0,0 +1,436 @@
...
+
+            (rpc_method, method_params)
+        }
+        RpcCommand::GetBeadCount => {
+            let rpc_method = String::from("getbeadcount");
+            let method_params = ArrayParams::new();
+
+            (rpc_method, method_params)
+        }
+        RpcCommand::GetCohortCount => {
```
mcelrath, line 80:
```
curl --json '{"jsonrpc": "2.0", "method": "getbeadcount", "id": 2}' http://localhost:6682/
{"jsonrpc":"2.0","id":2,"result":"0"}
curl --json '{"jsonrpc": "2.0", "method": "getcohortcount", "id": 2}' http://localhost:6682/
{"jsonrpc":"2.0","id":2,"result":"1"}
```

zero beads and one cohort? How did that happen?

node/src/main.rs
```diff
@@ -115,8 +98,22 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+    //spawning the rpc server
+    if let Some(rpc_command) = args.command {
+        let server_address = tokio::spawn(run_rpc_server(Arc::clone(&braid)));
+        let socket_address = server_address.await.unwrap().unwrap();
+        let parsing_handle =
+            tokio::spawn(parse_arguments(rpc_command, socket_address.clone())).await;
+    } else {
+        //running the rpc server and updating the reference counter
+        //for shared ownership
+        let server_handler = tokio::spawn(run_rpc_server(Arc::clone(&braid))).await;
```
mcelrath, line 115: Gemini says this:
```

   1. RPC Server in `main.rs`: The RPC server is started and immediately awaited in
      main.rs. This will block the rest of the node's initialization (like the libp2p
      swarm) until an RPC command is handled and the server is shut down. The RPC
      server should likely run in a separate, non-blocking task.
```
But I don't think this is correct. I can see startup proceeding normally and I don't think this line is blocking.

What do you think?

mcelrath (changes requested):
Tested this with curl, seems fine. We need data!

Just fix the logic in Braid::extend() otherwise this looks good.

node/src/braid/mod.rs
Sansh2356: ACK

node/src/braid/mod.rs
Sansh2356: ACK .

node/src/bead/tests.rs
Sansh2356: Yes along with that for `jsonifying` the requests to forward it to the server .

node/src/main.rs
Sansh2356: The issue for 0 bead count and 1 cohort count, @mcelrath  is this  
`cohorts: vec![Cohort(bead_indices.clone())],`
Since if while initializing a braid via Braid::new if the input genesis beads is an empty vector of Beads then the Braid that is returned it has a cohort appended to its cohort array which is empty ,as :new does not handle the case when the Braid is initialized with no genesis .

node/src/rpc_server.rs
Sansh2356: ACK below conversation .

mcelrath (approved):

PR #225, opened 2025-07-09
Add the logic for orphan beads

Aritra8438:
This PR adds the logic to process `orphan beads`.


Sansh2356 (approved):
LGTM the `orphan` logic processing the beads recursively is appropriate it will solve the issues during the extend test according to the hard cases as we have that @mcelrath had derived from the simulation .

mcelrath (approved):
Looks correct to me.

PR #227, opened 2025-07-14
Add `json` tests for `braid::extend()`

Aritra8438:
This PR adds `json` files as tests for extend function.


Sansh2356 (approved):
LGTM. Updating the existing `extend` function and adding `cohort_tips` attribute along with the merging of `extend` functions for the simulation generated test files from our test directory is good and was required as conveyed by @mcelrath .

PR #228, opened 2025-07-15
fix: Introduces error types and fallback logic

zaidmstrr:
Address these [comments](https://github.com/braidpool/braidpool/pull/213#discussion_r2202783776), [2](https://github.com/braidpool/braidpool/pull/213#discussion_r2202784827), [3](https://github.com/braidpool/braidpool/pull/213#discussion_r2202781636)

Introduces fallback logic for different connection errors. And introduces an enum for error classification.

### DIFF
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -45,10 +45,10 @@ pub struct Cli {
     pub zmqhashblockport: u16,
 
     /// Use IPC Bitcoin Core communication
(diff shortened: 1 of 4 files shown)


Sansh2356:
Looks fine to me all comments given by @mcelrath has been addressed, and the structure and flow is also apt it has already been discussed over lengths in previous discussions as well between myself and @zaidmstrr .

PR #230, opened 2025-07-22
MUI Removal, Code Refactor and Theme updates

its-me-abhishek:
- Removed MUI as dependency all over the dashboard
- Used already installed icons in place of MUI
- Replaced theme-ing with Tailwind based approach instead
- Updated tests wherever required
- Refactored BraidpoolDAG, and made it as modular as possible
- TODO in follow up PRs - Move constants to separate files, remove inline CSS wherever it is being used as of now (very few files), Fix UI / Alignments / Colors overall


priyashuu:
Do we have KPI cards? I don't think we do. If not, we should probably remove the related props from the types.
image

priyashuu:
Hey! I noticed we're currently using both react-router-dom and a currentPage state in the Header component to handle navigation. Since we already have URL-based routing set up in App.tsx, I think we can simplify things by fully switching to react-router-dom for navigation.

That would mean:

Removing the currentPage prop and related logic from the Header,

Updating the nav buttons to use <Link> or useNavigate(),

priyashuu:
As we discussed earlier, I think it would be better if we remove the Installations tab. Just wanted to check in do you think we should still keep it, or go ahead and remove it? Lmk what you think!

priyashuu:
Remove unused layout constants COLUMN_WIDTH and VERTICAL_SPACING from the `BraidPoolDag.tsk` file
Both COLUMN_WIDTH and VERTICAL_SPACING are declared but not used anywhere in the component:
Screenshot 2025-07-23 003642

dashboard/src/components/BraidPoolDAG/BraidPoolDAGUtils.ts
```diff
@@ -0,0 +1,233 @@
...
+  const lines: Array<[Position, Position]> = [];
+
+  remainingNodes.forEach((node) => {
+    const bead = node.id;
+    const baseX = margin.left + proposedX[bead] * COLUMN_WIDTH;
+    let currentY = centerY;
+    let distance = 0;
+    let direction = 1;
+
+    while (true) {
```
priyashuu, line 146: you can extract the collision and intersection checks into reusable helper functions to make this cleaner should be straightforward,

priyashuu (changes requested):

priyashuu (approved):

PR #91, opened 2025-03-28
WIP reproducible consensus graphs

mcelrath:
New iPython notebook for consensus doc graphs (WIP).

Modify simulator to not compute cohorts when in operation, if difficulty adjustment is not being used. (Makes it a lot faster for thick braids used in the consensus doc for low difficulty in the graph).

PR #233, opened 2025-07-30
Implementation of `downstream` connection with `braidpool` via stratum

Sansh2356:
- `Stratum` implementation for downstream to upstream communication.
- `IPC` template consumption and `mining.notify` for the job transfer to the downstream node.
- `Tracker` for job mapping wrt to the `downstream` nodes .
- `Client2Server` for messages such as `mining.submit` and difficulty adjustment according to the valid mined bead.
- `Validation` of job received from the downstream and constructing valid `bead` .

### DIFF
--- a/.github/workflows/dashboard-test.yml
+++ b/.github/workflows/dashboard-test.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Build project
         run: npm run build --if-present
         working-directory: dashboard
-      #TO BE REINSTATED 
-      # - name: Run tests
-      #   run: npm test
-      #   working-directory: dashboard
+     
+      - name: Run tests
+        run: npm test
+        working-directory: dashboard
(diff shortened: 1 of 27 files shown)


mcelrath:
Regarding difficulty adjustment:

There are TWO difficulties involved here:
1) a per-miner difficulty.
2) the Braidpool network difficulty.

(1) We will have a measurement of the network latency parameter `a` from the braid. The per-miner difficulty should be chosen so that each mining device produces a response regularly, (roughly every `a` (seconds) or some multiple of `a`). The way ckpool targets difficulty is relevant here. We will want to keep statistics about mining devices using the miner's responses (so we can track individual-device hashrate and problems with devices). Note that this rate is not a constant (because `a` is not constant, and devices can be over/under clocked).

(2) The network difficulty is the subject of my work in PR #91 . You can think of that work as outputting the parameters `a` and `lambda` from network/braid measurements which you can use in (1). Secondly it decides when a share becomes a bead. As discussed on the call today, to start we can set this to a constant for simplicity to start, let's say 1/500 of bitcoin's difficulty. (500*bitcoin_target).

PR #234, opened 2025-07-31
feat: Add valid template creator

zaidmstrr:
This PR adds functionality to compute valid block template data by replacing and recomputing the coinbase and merkle root from the block template received  through the `IPC`  function call. The `merkle root` is created using merkle paths only, not by using each transaction IDs.  The current implementation doesn't return the block header only; it will return whole valid block hex for now, but functions from `template_creator` are exposed publicly and can be used to create the valid block header only.

It also introduces some new functions in `client.rs`:
- `check_block`: Can be used to check the correctness of the block data.
- `get_block_template_components`: returns full `block_data`, `merkle path`, `coinbase`, `fees`, `header`,  `coinbase commitment`.
- `wait_for_tip_change`: Used to track the new blockchain tip (works as block notification in our case)

This PR also removes the dependency of the chain [interface](https://github.com/bitcoin/bitcoin/pull/29409) which is not currently part of the actual bitcoin implementation.


node/src/ipc/client.rs
```diff
@@ -31,12 +22,27 @@ pub fn bytes_to_hex(bytes: &[u8]) -> String {
...
+    pub reason: String,
+    pub debug: String,
+    pub result: bool,
+}
+
+#[derive(Debug, Clone)]
+pub struct BlockTemplateComponents {
+    pub header: Vec<u8>,
+    pub coinbase_transaction: Vec<u8>,
+    pub fees: Vec<u64>,
```
mcelrath, line 36: Is this necessary? Fees aren't explicit anywhere in bitcoin. Instead the coinbase contains an output for the mining reward that includes fees and must be no larger than the fees+subsidy.

So having a fees field here means that the fees are implicitly in two places: the coinbase outputs and this field.

mcelrath (approved):
Future work:

1. Add bitcoin as a submodule dependency so we can get a specific release of bitcoin
2. Automatically compile capnp files within the bitcoin directory
3. Reformat capnp output using rustfmt.

PR #235, opened 2025-08-02
Add more rpc endpoints

Aritra8438:
This PR dds more RPC endpoint and an AI-generated documentation for those endpoints.


node/RPC_API_DOCUMENTATION.md
```diff
@@ -0,0 +1,344 @@
...
+```
+
+**Error Codes:**
+- **2**: Internal error (JSON serialization failed)
+- **5**: No cohorts available
+- **6**: Cohort index out of bounds
+
+---
+
+## Command Line Interface
```
mcelrath, line 244: This is incorrect, and also doesn't work.

I have to run (no "rpc" parameter):
```
target/debug/node --ipc-socket=/tmp/bitcoin-ipc.sock get-tips
```
but this fails with "Address already in use" because it tries to bind to the same port as the (running) Braidpool node.

mcelrath (changes requested):
These RPCs are currently returning 0 beads and empty tips.

When the braid is empty we should be returning the Bitcoin tip as the genesis bead. This isn't necessarily a problem with this PR or the RPCs though.

Please move these Markdown docs into a new directory `docs/rpc` and fix the CLI.

parthdude07:
Hi @mcelrath @priyashuu , I’d like to work on this.

I'll fix the CLI "address in use" error by separating client logic from server startup, ensure get_tips falls back to the Bitcoin tip when the graph is empty, and move the docs to docs/rpc/.

Please assign this to me.

parthdude07:
@mcelrath 
I have implemented the fixes you requested.
I am currently verifying these changes locally with manual  tests. Before I submit the superseding PR, are there any specific integration test scripts or edge cases you would like me to run to ensure no regression?

priyashuu:
> @mcelrath I have implemented the fixes you requested. I am currently verifying these changes locally with manual tests. Before I submit the superseding PR, are there any specific integration test scripts or edge cases you would like me to run to ensure no regression?

I think you should push your changes; we will review them. We will let you know what's left.

PR #236, opened 2025-08-03
refactor: move all constants and URLs to files

its-me-abhishek:
- Created Constants.ts files in subfolders, wherever there are constants that can/are be/being re-used.
- Added a centralised URLs.ts file as well, for better URL fetching, and updating.


priyashuu:
LGTM! The Next TODO is to refactor all hardcoded API URLs and WebSocket endpoints to use environment variables via .env files.

priyashuu (approved):

PR #240, opened 2025-08-11
Added support for IBD 

ABD-AZE:
## Additional changes made
- Added helper functions (reset , get_beads_after) in braid/mod.rs .
- Added tests for verifying correct serialisation and deserialisation of enum variants in bead/mod.rs .
- Added extensive tests for get_beads_after() function .
- Added better error handling by introducing BeadSyncError enum .
- Complete `IBD` functionality following the flow as `GetTips` ---> `GetBeads` ----> `GetData` along with acknowledgement.
- `IBDHandler` for managing the cache requirements corresponding to `IBD` .
- Tested over the cases such as 
  * A new node is spawned containing zero beads and undergoes IBD from a sync node .
  * A new node was spawned and after syncing but it has not mined and the sync node also didn't mined .
  * A new node after syncing mined beads and the peer node was also connected .
  * A new node after syncing mined beads and the peer got disconnected and reconnected to undergo IBD wrt new node which mined some beads during the time the sync node got disconnected .
  * Vice-versa of 4th just mining being done at the end of peer node instead of new node .

### NOTE: Algorithm works only if Braid::cohorts vector is topologically sorted
I have read the extend function and it seems that the above condition is always true, please review this once again for any edge cases.

### DIFF
--- a/node/src/bead/mod.rs
+++ b/node/src/bead/mod.rs
@@ -70,7 +70,7 @@ impl Decodable for Bead {
 #[derive(Debug, Clone, PartialEq)]
 pub enum BeadRequest {
     // Request beads from a specific set of hashes
-    GetBeads(HashSet<BeadHash>),
+    GetBeads(Vec<BeadHash>),
     // Request the latest tips from a peer
     GetTips,
     GetGenesis,
@@ -138,9 +138,8 @@ impl Encodable for BeadRequest {
             BeadRequest::GetBeads(hashes) => {
                 let mut written = 0;
                 written += GET_BEADS.consensus_encode(writer)?; // 0 for GetBeads
-                let hashes_vec = hashset_to_vec_deterministic(hashes);
-                written += (hashes_vec.len() as u32).consensus_encode(writer)?;
-                for hash in hashes_vec {
+                written += (hashes.len() as u32).consensus_encode(writer)?;
+                for hash in hashes {
                     written += hash.consensus_encode(writer)?;
                 }
                 Ok(written)
@@ -173,10 +172,10 @@ impl Decodable for BeadRequest {
         match request_type {
             GET_BEADS => {
                 let count = u32::consensus_decode(d)?;
-                let mut hashes = HashSet::new();
+                let mut hashes = Vec::new();
                 for _ in 0..count {
                     let hash = BeadHash::consensus_decode(d)?;
-                    hashes.insert(hash);
+                    hashes.push(hash);
                 }
                 Ok(BeadRequest::GetBeads(hashes))
             }
--- a/node/src/bead/mod.rs
+++ b/node/src/bead/mod.rs
@@ -17,7 +17,8 @@ const GET_BEADS: u8 = 0;
 const GET_TIPS: u8 = 1;
 const GET_GENESIS: u8 = 2;
 const GET_ALL_BEADS: u8 = 3;
-const BEAD_RESPONSE_ERROR: u8 = 4;
+const GET_BEADS_AFTER: u8 = 4;
+const BEAD_RESPONSE_ERROR: u8 = 5;
 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct Bead {
@@ -74,6 +75,7 @@ pub enum BeadRequest {
     GetTips,
     GetGenesis,
     GetAllBeads,
+    GetBeadsAfter(Vec<BeadHash>),
 }
 
 // Response types for bead download
@@ -87,8 +89,47 @@ pub enum BeadResponse {
     Genesis(Vec<BeadHash>),
     // Get all beads for IBD
     GetAllBeads(Vec<Bead>),
+    // Get beads after a specific set of hashes
+    GetBeadsAfter(Vec<Bead>),
     // Error response
-    Error(String),
+    Error(BeadSyncError),
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum BeadSyncError {
+    GenesisMismatch,
+    Other(String),
+}
+
+impl Encodable for BeadSyncError {
+    fn consensus_encode<W: Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
+        match self {
+            BeadSyncError::GenesisMismatch => 0u8.consensus_encode(w),
+            BeadSyncError::Other(message) => {
+                let mut written = 0;
+                written += 1u8.consensus_encode(w)?;
+                written += message.consensus_encode(w)?;
+                Ok(written)
+            }
+        }
+    }
+}
+
+impl Decodable for BeadSyncError {
+    fn consensus_decode<D: BufRead + ?Sized>(d: &mut D) -> Result<Self, Error> {
+        let error_type = u8::consensus_decode(d)?;
+        match error_type {
+            0 => Ok(BeadSyncError::GenesisMismatch),
+            1 => {
+                let message = String::consensus_decode(d)?;
+                Ok(BeadSyncError::Other(message))
+            }
+            _ => Err(Error::from(io::Error::new(
+                io::ErrorKind::InvalidData,
+                "Invalid BeadSyncError type",
+            ))),
+        }
+    }
 }
 
 impl Encodable for BeadRequest {
@@ -113,6 +154,15 @@ impl Encodable for BeadRequest {
             BeadRequest::GetAllBeads => {
                 GET_ALL_BEADS.consensus_encode(writer) // 3 for GetAllBeads
             }
+            BeadRequest::GetBeadsAfter(hashes) => {
+                let mut written = 0;
+                written += GET_BEADS_AFTER.consensus_encode(writer)?;
+                written += (hashes.len() as u32).consensus_encode(writer)?;
+                for hash in hashes {
+                    written += hash.consensus_encode(writer)?;
+                }
+                Ok(written)
+            }
         }
     }
 }
@@ -133,6 +183,15 @@ impl Decodable for BeadRequest {
             GET_TIPS => Ok(BeadRequest::GetTips),
             GET_GENESIS => Ok(BeadRequest::GetGenesis),
             GET_ALL_BEADS => Ok(BeadRequest::GetAllBeads),
+            GET_BEADS_AFTER => {
+                let count = u32::consensus_decode(d)?;
+                let mut hashes = Vec::new();
+                for _ in 0..count {
+                    let hash = BeadHash::consensus_decode(d)?;
+                    hashes.push(hash);
+                }
+                Ok(BeadRequest::GetBeadsAfter(hashes))
+            }
             _ => Err(Error::from(io::Error::new(
                 io::ErrorKind::InvalidData,
                 "Invalid BeadRequest type",
@@ -182,8 +241,17 @@ impl Encodable for BeadResponse {
             }
             BeadResponse::Error(error) => {
                 let mut written = 0;
-                written += BEAD_RESPONSE_ERROR.consensus_encode(writer)?; // 4 for Error
+                written += BEAD_RESPONSE_ERROR.consensus_encode(writer)?;
                 written += error.consensus_encode(writer)?;
+                Ok(written)
+            }
+            BeadResponse::GetBeadsAfter(beads) => {
+                let mut written = 0;
+                written += GET_BEADS_AFTER.consensus_encode(writer)?;
+                written += (beads.len() as u32).consensus_encode(writer)?;
+                for bead in beads {
+                    written += bead.consensus_encode(writer)?;
+                }
                 Ok(written)
             }
         }
@@ -231,8 +299,17 @@ impl Decodable for BeadResponse {
                 Ok(BeadResponse::GetAllBeads(beads))
             }
             BEAD_RESPONSE_ERROR => {
-                let error = String::consensus_decode(d)?;
+                let error = BeadSyncError::consensus_decode(d)?;
                 Ok(BeadResponse::Error(error))
+            }
+            GET_BEADS_AFTER => {
+                let count = u32::consensus_decode(d)?;
+                let mut beads = Vec::new();
+                for _ in 0..count {
+                    let bead = Bead::consensus_decode(d)?;
+                    beads.push(bead);
+                }
+                Ok(BeadResponse::GetBeadsAfter(beads))
             }
             _ => Err(Error::from(io::Error::new(
                 io::ErrorKind::InvalidData,
--- a/node/src/bead/mod.rs
+++ b/node/src/bead/mod.rs
@@ -1,6 +1,6 @@
 use crate::committed_metadata::CommittedMetadata;
 use crate::uncommitted_metadata::UnCommittedMetadata;
-use crate::utils::{hashset_to_vec_deterministic, BeadHash};
+use crate::utils::BeadHash;
 use async_trait::async_trait;
 use bitcoin::consensus::encode::Decodable;
 use bitcoin::consensus::encode::Encodable;
@@ -11,7 +11,6 @@ use libp2p::futures::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 use libp2p::request_response::Codec;
 use libp2p::StreamProtocol;
 use serde::{Deserialize, Serialize};
-use std::collections::HashSet;
 
 const GET_BEADS: u8 = 0;
 const GET_TIPS: u8 = 1;
--- a/node/src/bead/mod.rs
+++ b/node/src/bead/mod.rs
@@ -89,7 +89,7 @@ pub enum BeadResponse {
     // Get all beads for IBD
     GetAllBeads(Vec<Bead>),
     // Get beads after a specific set of hashes
-    GetBeadsAfter(Vec<Bead>),
+    GetBeadsAfter(Vec<BeadHash>),
     // Error response
     Error(BeadSyncError),
 }
@@ -247,8 +247,8 @@ impl Encodable for BeadResponse {
                 let mut written = 0;
                 written += GET_BEADS_AFTER.consensus_encode(writer)?;
                 written += (beads.len() as u32).consensus_encode(writer)?;
-                for bead in beads {
-                    written += bead.consensus_encode(writer)?;
+                for bead_hash in beads {
+                    written += bead_hash.consensus_encode(writer)?;
                 }
                 Ok(written)
             }
@@ -302,12 +302,12 @@ impl Decodable for BeadResponse {
             }
             GET_BEADS_AFTER => {
                 let count = u32::consensus_decode(d)?;
-                let mut beads = Vec::new();
+                let mut bead_hashes = Vec::new();
                 for _ in 0..count {
-                    let bead = Bead::consensus_decode(d)?;
-                    beads.push(bead);
+                    let bead_hash = BeadHash::consensus_decode(d)?;
+                    bead_hashes.push(bead_hash);
                 }
-                Ok(BeadResponse::GetBeadsAfter(beads))
+                Ok(BeadResponse::GetBeadsAfter(bead_hashes))
             }
             _ => Err(Error::from(io::Error::new(
                 io::ErrorKind::InvalidData,
--- a/node/src/bead/mod.rs
+++ b/node/src/bead/mod.rs
@@ -11,14 +11,6 @@ use libp2p::futures::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 use libp2p::request_response::Codec;
 use libp2p::StreamProtocol;
 use serde::{Deserialize, Serialize};
-
-const GET_BEADS: u8 = 0;
-const GET_TIPS: u8 = 1;
-const GET_GENESIS: u8 = 2;
-const GET_ALL_BEADS: u8 = 3;
-const GET_BEADS_AFTER: u8 = 4;
-const BEAD_RESPONSE_ERROR: u8 = 5;
-
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct Bead {
     pub block_header: BlockHeader,
@@ -62,6 +54,31 @@ impl Decodable for Bead {
             committed_metadata,
             uncommitted_metadata,
         })
+    }
+}
+
+#[repr(u8)]
+#[derive(Debug, Clone, Copy)]
+pub enum BeadRequestType {
+    GetBeads = 0,
+    GetTips = 1,
+    GetGenesis = 2,
+    GetAllBeads = 3,
+    GetBeadsAfter = 4,
+    BeadResponseError = 5,
+}
+//Converting from u8 to `RequestType`/ `ResponseType`
+impl BeadRequestType {
+    pub fn from_u8(v: u8) -> Option<Self> {
+        match v {
+            0 => Some(Self::GetBeads),
+            1 => Some(Self::GetTips),
+            2 => Some(Self::GetGenesis),
+            3 => Some(Self::GetAllBeads),
+            4 => Some(Self::GetBeadsAfter),
+            5 => Some(Self::BeadResponseError),
+            _ => None,
+        }
     }
 }
 
@@ -136,25 +153,21 @@ impl Encodable for BeadRequest {
         match self {
             BeadRequest::GetBeads(hashes) => {
                 let mut written = 0;
-                written += GET_BEADS.consensus_encode(writer)?; // 0 for GetBeads
+                written += (BeadRequestType::GetBeads as u8).consensus_encode(writer)?;
                 written += (hashes.len() as u32).consensus_encode(writer)?;
                 for hash in hashes {
                     written += hash.consensus_encode(writer)?;
                 }
                 Ok(written)
             }
-            BeadRequest::GetTips => {
-                GET_TIPS.consensus_encode(writer) // 1 for GetTips
-            }
-            BeadRequest::GetGenesis => {
-                GET_GENESIS.consensus_encode(writer) // 2 for GetGenesis
-            }
+            BeadRequest::GetTips => (BeadRequestType::GetTips as u8).consensus_encode(writer),
+            BeadRequest::GetGenesis => (BeadRequestType::GetGenesis as u8).consensus_encode(writer),
             BeadRequest::GetAllBeads => {
-                GET_ALL_BEADS.consensus_encode(writer) // 3 for GetAllBeads
+                (BeadRequestType::GetAllBeads as u8).consensus_encode(writer)
             }
             BeadRequest::GetBeadsAfter(hashes) => {
                 let mut written = 0;
-                written += GET_BEADS_AFTER.consensus_encode(writer)?;
+                written += (BeadRequestType::GetBeadsAfter as u8).consensus_encode(writer)?;
                 written += (hashes.len() as u32).consensus_encode(writer)?;
                 for hash in hashes {
                     written += hash.consensus_encode(writer)?;
@@ -167,9 +180,12 @@ impl Encodable for BeadRequest {
 
 impl Decodable for BeadRequest {
     fn consensus_decode<D: BufRead + ?Sized>(d: &mut D) -> Result<Self, Error> {
-        let request_type = u8::consensus_decode(d)?;
+        let request_type_u8 = u8::consensus_decode(d)?;
+        let request_type = BeadRequestType::from_u8(request_type_u8).ok_or_else(|| {
+            io::Error::new(io::ErrorKind::InvalidData, "Invalid bead request type")
+        })?;
         match request_type {
-            GET_BEADS => {
+            BeadRequestType::GetBeads => {
                 let count = u32::consensus_decode(d)?;
                 let mut hashes = Vec::new();
                 for _ in 0..count {
@@ -178,10 +194,10 @@ impl Decodable for BeadRequest {
                 }
                 Ok(BeadRequest::GetBeads(hashes))
             }
-            GET_TIPS => Ok(BeadRequest::GetTips),
-            GET_GENESIS => Ok(BeadRequest::GetGenesis),
-            GET_ALL_BEADS => Ok(BeadRequest::GetAllBeads),
-            GET_BEADS_AFTER => {
+            BeadRequestType::GetTips => Ok(BeadRequest::GetTips),
+            BeadRequestType::GetGenesis => Ok(BeadRequest::GetGenesis),
+            BeadRequestType::GetAllBeads => Ok(BeadRequest::GetAllBeads),
+            BeadRequestType::GetBeadsAfter => {
                 let count = u32::consensus_decode(d)?;
                 let mut hashes = Vec::new();
                 for _ in 0..count {
@@ -203,7 +219,7 @@ impl Encodable for BeadResponse {
         match self {
             BeadResponse::Beads(beads) => {
                 let mut written = 0;
-                written += GET_BEADS.consensus_encode(writer)?; // 0 for Beads
+                written += (BeadRequestType::GetBeads as u8).consensus_encode(writer)?;
                 written += (beads.len() as u32).consensus_encode(writer)?;
                 for bead in beads {
                     written += bead.consensus_encode(writer)?;
@@ -212,7 +228,7 @@ impl Encodable for BeadResponse {
             }
             BeadResponse::Tips(tips) => {
                 let mut written = 0;
-                written += GET_TIPS.consensus_encode(writer)?; // 1 for Tips
+                written += (BeadRequestType::GetTips as u8).consensus_encode(writer)?;
                 written += (tips.len() as u32).consensus_encode(writer)?;
                 for tip in tips {
                     written += tip.consensus_encode(writer)?;
@@ -221,7 +237,7 @@ impl Encodable for BeadResponse {
             }
             BeadResponse::Genesis(genesis) => {
                 let mut written = 0;
-                written += GET_GENESIS.consensus_encode(writer)?; // 2 for Genesis
+                written += (BeadRequestType::GetGenesis as u8).consensus_encode(writer)?;
                 written += (genesis.len() as u32).consensus_encode(writer)?;
                 for hash in genesis {
                     written += hash.consensus_encode(writer)?;
@@ -230,7 +246,7 @@ impl Encodable for BeadResponse {
             }
             BeadResponse::GetAllBeads(beads) => {
                 let mut written = 0;
-                written += GET_ALL_BEADS.consensus_encode(writer)?; // 3 for GetAllBeads
+                written += (BeadRequestType::GetAllBeads as u8).consensus_encode(writer)?;
                 written += (beads.len() as u32).consensus_encode(writer)?;
                 for bead in beads {
                     written += bead.consensus_encode(writer)?;
@@ -239,13 +255,13 @@ impl Encodable for BeadResponse {
             }
             BeadResponse::Error(error) => {
                 let mut written = 0;
-                written += BEAD_RESPONSE_ERROR.consensus_encode(writer)?;
+                written += (BeadRequestType::BeadResponseError as u8).consensus_encode(writer)?;
                 written += error.consensus_encode(writer)?;
                 Ok(written)
             }
             BeadResponse::GetBeadsAfter(beads) => {
                 let mut written = 0;
-                written += GET_BEADS_AFTER.consensus_encode(writer)?;
+                written += (BeadRequestType::GetBeadsAfter as u8).consensus_encode(writer)?;
                 written += (beads.len() as u32).consensus_encode(writer)?;
                 for bead_hash in beads {
                     written += bead_hash.consensus_encode(writer)?;
@@ -258,9 +274,12 @@ impl Encodable for BeadResponse {
 
 impl Decodable for BeadResponse {
     fn consensus_decode<D: BufRead + ?Sized>(d: &mut D) -> Result<Self, Error> {
-        let response_type = u8::consensus_decode(d)?;
+        let response_type_u8 = u8::consensus_decode(d)?;
+        let response_type = BeadRequestType::from_u8(response_type_u8).ok_or_else(|| {
+            io::Error::new(io::ErrorKind::InvalidData, "Invalid bead response type")
+        })?;
         match response_type {
-            GET_BEADS => {
+            BeadRequestType::GetBeads => {
                 let count = u32::consensus_decode(d)?;
                 let mut beads = Vec::new();
                 for _ in 0..count {
@@ -269,7 +288,7 @@ impl Decodable for BeadResponse {
                 }
                 Ok(BeadResponse::Beads(beads))
             }
-            GET_TIPS => {
+            BeadRequestType::GetTips => {
                 let count = u32::consensus_decode(d)?;
                 let mut tips = Vec::new();
                 for _ in 0..count {
@@ -278,7 +297,7 @@ impl Decodable for BeadResponse {
                 }
                 Ok(BeadResponse::Tips(tips))
             }
-            GET_GENESIS => {
+            BeadRequestType::GetGenesis => {
                 let count = u32::consensus_decode(d)?;
                 let mut genesis = Vec::new();
                 for _ in 0..count {
@@ -287,7 +306,7 @@ impl Decodable for BeadResponse {
                 }
                 Ok(BeadResponse::Genesis(genesis))
             }
-            GET_ALL_BEADS => {
+            BeadRequestType::GetAllBeads => {
                 let count = u32::consensus_decode(d)?;
                 let mut beads = Vec::new();
                 for _ in 0..count {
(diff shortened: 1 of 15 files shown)


mcelrath (commented):
Overall this is good, it just lacks tests. Here is a list from Gemini of missing tests:


   * No `get_beads_after` Logic Tests: The new Braid::get_beads_after() function in
     node/src/braid/mod.rs, which contains the core logic for this feature, has no unit
     tests. As I mentioned in the review, the logic of this function might be fragile,
     making the lack of tests a critical issue.
   * No Serialization Tests: The new BeadRequest::GetBeadsAfter and
     BeadResponse::GetBeadsAfter variants are not tested for serialization and
     deserialization in node/src/bead/tests.rs. This means there's no guarantee they can be
     correctly encoded and decoded for transmission over the network.
   * No `BeadSyncError` Serialization Tests: The new BeadSyncError enum itself lacks unit
     tests for its Encodable and Decodable implementations.
   * No End-to-End Test: There are no integration tests that simulate a full
     request-response cycle for the GetBeadsAfter flow between two peers. The test in
     node/src/behaviour/tests.rs only covers the older GetBeads flow.
   * No `GenesisMismatch` Test: The new logic for handling a GenesisMismatch error (where a
     peer receives this error and then requests the genesis beads) is not covered by any
     test.

node/src/braid/mod.rs
```diff
@@ -249,7 +258,34 @@ impl Braid {
...
+            let bead_hash = bead.block_header.block_hash();
+            if flag {
+                response_beads.push(bead.clone());
+            } else {
+                if old_tips_and_genesis.contains(&bead_hash) {
+                    flag = true;
+                    continue;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
mcelrath: I think there's significant potential for missing beads at the start point here. We really should use the *minimum* index, but even that is not a guarantee. I'd prefer something like this:
1. Find the earliest cohort that has any bead in old_tips_and_genesis as a member.
2. Re-send that entire cohort and go forward from there.

node/src/braid/mod.rs
```diff
@@ -249,7 +258,34 @@ impl Braid {
...
+        for bead in &self.beads {
+            let bead_hash = bead.block_header.block_hash();
+            if flag {
+                response_beads.push(bead.clone());
+            } else {
+                if old_tips_and_genesis.contains(&bead_hash) {
+                    flag = true;
+                    continue;
+                }
```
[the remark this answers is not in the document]
mcelrath: This logic is a little convoluted. At the very least we should rename `flag` to something like `found_start` or something. Also you should have used `bead_index_mapping` to get the indices instead of iterating over all beads.

But the above cohort-based start point would be even better.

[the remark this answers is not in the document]
ABD-AZE: bead_index_mapping doesn't guarantee that the order of the beads would be in topologically sorted order since bead_index_mapping is a Hashmap.

node/src/braid/mod.rs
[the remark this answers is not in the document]
ABD-AZE: Yes, this is indeed a better approach I think and by doing this we can eliminate the need to match the genesis as well, just a small doubt here, after considering that earliest cohort should I send all the cohorts from there or just traverse over the beads after that similar to what is currently done in the code?

node/src/braid/mod.rs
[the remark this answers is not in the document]
mcelrath: Use bead_index_mapping to find the starting position in the beads Vec, and then iterate. Obvs we can't iterate over the HashMap.

node/src/braid/mod.rs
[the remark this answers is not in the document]
mcelrath: Once you have a good starting point, just send everything. The receiver will get a few duplicate beads but they're small and there's really no benefit in trying to avoid sending a few duplicates.

node/src/braid/mod.rs
```diff
@@ -212,6 +218,7 @@ impl Braid {
                     AddBeadStatus::InvalidBead => {
                         continue;
                     }
+                    //Unecessary condition
                     AddBeadStatus::ParentsNotYetReceived => {
                         self.orphan_beads.push(orphan_bead);
```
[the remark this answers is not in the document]
Sansh2356: Comment is wrong .

node/src/ibd_manager/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+                            }
+                            Err(error) => {
+                                tracing::error!(
+                                error=?error, "Error while updating and sending it to request channel due to - "
+                                );
+                            }
+                        };
+                    } else {
+                        self.batch_mapping.insert(peer_id, IBD_BATCH_SIZE);
+                        match offset_sender.send(IBD_BATCH_SIZE) {
```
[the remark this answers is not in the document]
Sansh2356: This is 2nd batch first batch offset does not require us to fetch an offset .

node/src/main.rs
```diff
@@ -460,7 +478,19 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                                            info!("Insert command sent successfully to db handler after receiving bead from peer");
+                                        },
+                                        Err(error)=>{
+                                            error!(
+                                                source = ?message.source,
+                                                err = ?error.0,
+                                                "An error occurred while sending insert bead command received from peer"
+                                            );
+                                        }
+                                     };
```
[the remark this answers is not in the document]
Sansh2356: Retry mechanism is for future PRs .

node/src/braid/mod.rs
```diff
@@ -248,7 +255,65 @@ impl Braid {
...
+        for (idx, cohort) in self.cohorts.iter().enumerate() {
+            if cohort.0.contains(&smallest_index) {
+                //Finding the cohort for which the smallest index is a part of
+                smallest_cohort_index = idx;
+                break;
+            }
+        }
+        tracing::debug!(
+            smallest_index=?smallest_index,"Smallest possible cohort index for which the given smallest index is a part of - ",
+        );
```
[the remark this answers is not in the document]
Sansh2356: This can never happen if it does it ideally should be an error .

node/src/bead/mod.rs
```diff
@@ -53,11 +54,12 @@ impl Decodable for Bead {
 #[derive(Debug, Clone, PartialEq)]
 pub enum BeadRequest {
     // Request beads from a specific set of hashes
-    GetBeads(HashSet<BeadHash>),
+    GetBeads(Vec<BeadHash>),
     // Request the latest tips from a peer
     GetTips,
     GetGenesis,
     GetAllBeads,
+    GetBeadsAfter(Vec<BeadHash>),
```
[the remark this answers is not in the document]
mcelrath: Vec is marginally better here. Hashset can in principle have collisions though it's extremely unlikely.

node/src/bead/mod.rs
```diff
@@ -87,8 +88,47 @@ pub enum BeadResponse {
...
     // Error response
-    Error(String),
+    Error(BeadSyncError),
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum BeadSyncError {
+    GenesisMismatch,
+    Other(String),
```
mcelrath: Another error code should be UnknownBead -- if the peer requests a bead we haven't seen.

node/src/bead/mod.rs
```diff
@@ -11,13 +11,13 @@ use libp2p::futures::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 use libp2p::request_response::Codec;
 use libp2p::StreamProtocol;
 use serde::{Deserialize, Serialize};
-use std::collections::HashSet;
 
 const GET_BEADS: u8 = 0;
```
mcelrath: Remove these magic constants and use the enum instead. (see below)

node/src/bead/mod.rs
```diff
@@ -69,11 +69,12 @@ impl Decodable for Bead {
 #[derive(Debug, Clone, PartialEq)]
 pub enum BeadRequest {
     // Request beads from a specific set of hashes
-    GetBeads(HashSet<BeadHash>),
+    GetBeads(Vec<BeadHash>),
```
mcelrath: Put the magic constants here:
```
  #[repr(u8)]
  #[derive(Debug, Clone, PartialEq)]
  pub enum BeadRequest {
      GetBeads(Vec<BeadHash>) = 0,
      GetTips = 1,
      GetGenesis = 2,
      GetAllBeads = 3,
      GetBeadsAfter(Vec<BeadHash>) = 4,
  }
```

Need before the enum:
```
#[repr(u8)]
```

node/src/bead/mod.rs
```diff
@@ -87,8 +88,47 @@ pub enum BeadResponse {
...
+#[derive(Debug, Clone, PartialEq)]
+pub enum BeadSyncError {
+    GenesisMismatch,
+    Other(String),
+}
+
+impl Encodable for BeadSyncError {
+    fn consensus_encode<W: Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
+        match self {
+            BeadSyncError::GenesisMismatch => 0u8.consensus_encode(w),
```
mcelrath: You're hard coding 0u8 and 0u1 as the enum values here. These are magic protocol numbers and should be defined externally. Using the enum as our magic number repository instead, we can do:
```
            BeadSyncError::GenesisMismatch => (*self as u8).consensus_encode(w),
````

File: node/src/bead/mod.rs [hunk 12]
```diff
@@ -87,8 +88,47 @@ pub enum BeadResponse {
...
+            BeadSyncError::Other(message) => {
+                let mut written = 0;
+                written += 1u8.consensus_encode(w)?;
```
mcelrath: Using the enum idea:

```
                written += (*self as u8).consensus_encode(w)?;
```

node/src/bead/mod.rs
```diff
@@ -87,8 +88,47 @@ pub enum BeadResponse {
...
+            }
+        }
+    }
+}
+
+impl Decodable for BeadSyncError {
+    fn consensus_decode<D: BufRead + ?Sized>(d: &mut D) -> Result<Self, Error> {
+        let error_type = u8::consensus_decode(d)?;
+        match error_type {
+            0 => Ok(BeadSyncError::GenesisMismatch),
```
mcelrath: With the enum idea we can do:
```
        BeadSyncError::GenesisMismatch => Ok(BeadSyncError::GenesisMismatch)
```
which is much more readable than the bare `0`.

if you want you can write `from_u8(value: u8, d: &mut BufRead)` and `try_from(u8)` to make it a little more readable.

node/src/bead/mod.rs
```diff
@@ -123,16 +171,25 @@ impl Decodable for BeadRequest {
         match request_type {
             GET_BEADS => {
                 let count = u32::consensus_decode(d)?;
-                let mut hashes = HashSet::new();
+                let mut hashes = Vec::new();
```
mcelrath: Here's a one-liner to do the same.
```
hashes = (0..count).map(|_| BeadHash::consensus_decode(d)).collect()
```

node/src/braid/mod.rs
```diff
@@ -248,7 +254,70 @@ impl Braid {
             }
         }
     }
+    /// utility function for GetBeadsAfter request
+    pub fn get_beads_after(&self, old_tips: Vec<BeadHash>) -> Option<Vec<Bead>> {
+        let old_tips: HashSet<BeadHash> = old_tips.into_iter().collect();
+        tracing::warn!(
+            old_tips=?old_tips,"Tips received from the peer for which beads are requested for during IBD"
```
mcelrath, line 261: I don't understand what this means. How do you know we're in IBD? Are you saying a peer is requesting tips but we're in IBD?

node/src/braid/mod.rs
```diff
@@ -248,7 +255,65 @@ impl Braid {
...
+        );
+        //In case no tips are present i.e. the new braid-node has been initialized
+        if old_tips.len() == 0 {
+            return Some(self.beads.clone());
```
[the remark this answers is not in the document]
mcelrath: This is definitely a performance issue, how come you didn't want to change it?

node/src/braid/mod.rs
```diff
@@ -248,7 +254,70 @@ impl Braid {
...
+        }
+        let mut response_beads = Vec::new();
+        let mut found_start = false;
+        let mut smallest_index = usize::MAX;
+        //finding the starting index
+        for hash in &old_tips {
+            if let Some(&index) = self.bead_index_mapping.get(hash) {
```
mcelrath, line 271: If the hash is not in bead_index_mapping we should probably error.

node/src/braid/mod.rs
```diff
@@ -248,7 +254,70 @@ impl Braid {
...
+                if index < smallest_index {
+                    smallest_index = index;
+                }
+                found_start = true;
```
mcelrath: found_start doesn't do anything and isn't needed. If we error above on missing hashes, we should probably also error if old_tips is empty. Then after that smallest_index will always be the smallest in the following code without the found_start variable.

node/src/braid/mod.rs
```diff
@@ -248,7 +254,70 @@ impl Braid {
...
+            }
+        }
+        //If somehow no bead matched that can be due to possible latency/fork so send all the beads instead as fallback
+        if smallest_index == usize::MAX {
```
mcelrath, line 278: We can omit this by erroring above.

node/src/braid/mod.rs
```diff
@@ -248,7 +254,70 @@ impl Braid {
...
+            return Some(self.beads.clone());
+        }
+
+        tracing::debug!(
+            smallest_index=?smallest_index,"Smallest possible index from all the tips - ",
+
+        );
+        // just iterating over the vector of cohorts for now, this needs to be changed to use a more efficient retrieval of cohort index given bead hash
+        let mut smallest_cohort_index = usize::MAX;
+        for (idx, cohort) in self.cohorts.iter().enumerate() {
```
mcelrath, line 288: Hmmm. This is bad.

Add a reverse mapping `HashMap<bead_idx: usize, cohort_idx: usize>` in Braid so we can use it here.

node/src/braid/mod.rs
```diff
@@ -249,7 +258,34 @@ impl Braid {
...
+        for bead in &self.beads {
+            let bead_hash = bead.block_header.block_hash();
+            if flag {
+                response_beads.push(bead.clone());
+            } else {
+                if old_tips_and_genesis.contains(&bead_hash) {
+                    flag = true;
+                    continue;
+                }
+            }
```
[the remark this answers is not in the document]
mcelrath: Compiled for me?!?!?

node/src/braid/mod.rs
```diff
@@ -249,7 +258,33 @@ impl Braid {
...
+        }
```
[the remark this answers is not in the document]
mcelrath: Add the HashMap instead...

node/src/ibd_manager/mod.rs
```diff
@@ -0,0 +1,140 @@
+use std::collections::HashMap;
+
+use crate::utils::BeadHash;
+//Beads fetching will be done with batch size
+pub const IBD_BATCH_SIZE: usize = 100;
+
+//Storing tips mapping received from peers during `GetTips`
+//that will be flushed out after IBD is done hence no complete dependency
+#[derive(Debug)]
+pub enum IBDCommands {
```
mcelrath, line 14: Use the u8 enum idea above here too.

mcelrath (changes requested):

node/src/main.rs
```diff
@@ -51,9 +55,16 @@ use tokio::sync::{
...
-
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error>> {
+    let (mut ibd_manager, ibd_command_tx) = IBDManager::new();
+    //IBD cache handler
+    let _ibd_handler = tokio::spawn(async move {
+        ibd_manager.handle_ibd_command().await;
+    });
+
+    let ibd_or_not: AtomicBool = AtomicBool::new(false);
```
mcelrath: As discussed privately, this can't be a global like this because whether we're synced is a local evaluation, not global based on peers. All peer requests incur at least 2x the latency of floodsub, so we need to use our floodsub buffer and `incoming` as our source of truth on whether we're synced.

The Difficulty Adjustment Algorithm will measure a latency parameter $a$. For now, hard code it as 1 second. What we need to evaluate is how often we are have beads in `incoming` with respect to lambda. It is normal to have some beads in incoming, but not for a long time. We will have to develop a reasonable heuristic around this. 

How about: "average orphan lifetime" -- measure when we put a bead in `incoming` and when it becomes connected (because we've seen its parents). This measurement should be similar to lambda. Add timestamps for when we put beads into incoming and when they become connected. Keep statistics on this, and if over the last 10 x $a$ time period, beads spend less than 3 x $a$ in incoming, we consider ourselves synced and can begin mining. These numbers are arbitrary and we will have to calibrate them so make them parameters, but assume you will be able to obtain $a$ from the DAA.

node/src/main.rs
```diff
@@ -598,133 +632,358 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                            request,
+                            request_id: _,
+                            channel,
+                        } => {
+                            // Handle the bead sync request here
+                            match request {
+                                bead::BeadRequest::GetBeads(hashes) => {
+                                    let mut beads = Vec::new();
+                                    {
+                                        let braid_lock = braid.read().await;
```
mcelrath: This is potentially painful for keeping in sync, since it blocks writers to the Braid (incoming beads) if it takes more than $a$ ~ 1s to complete this request, we will sit with beads in `incoming` while we send IBD blocks, which causes our mined beads to mine old parents, cohorts to become large, and our payouts to be decreased or lost. Most of the time, peers will be requesting historic blocks not near the tip, and we don't want to block writing the braid.

Instead, consider using the DB directly and not locking the braid here unless the request is very close to the tips.

We also need to develop a strategy for syncing the DB to disk. We're not doing it at all right now. Might not matter in operation though, we can still read it without locks...

node/src/main.rs
```diff
@@ -598,133 +632,358 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                                            }
+                                        }
+                                    }
+                                    //Sending all the beads requested in the hashes supplied during `GetData` request
+                                    swarm.behaviour_mut().respond_with_beads(channel, beads);
+                                }
+                                bead::BeadRequest::GetTips => {
+                                    let tips;
+                                    {
+                                        let braid_lock = braid.read().await;
```
mcelrath: In contrast, this lock is acceptable because it is very quick. We're only reading a ~few beads, not 1000 beads which might be requested in GetBeads.

But it's really worthwhile thinking of ways to avoid this lock altogether. The time the lock is held is the time we're not syncing the braid, increasing our latency, and lowering our profits.

node/src/main.rs
```diff
@@ -598,133 +632,358 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                                            .cloned()
+                                            .map(|bead| bead.block_header.block_hash())
+                                            .collect();
+                                    }
+                                    swarm.behaviour_mut().respond_with_tips(channel, tips);
+                                }
+                                bead::BeadRequest::GetGenesis => {
+                                    let genesis;
+                                    {
+                                        let braid_lock = braid.read().await;
```
mcelrath: Don't necessarily need to lock the braid here. The genesis changes much more rarely. Maybe think of another way to do this. I think if we report accidentally a genesis from the previous epoch for $a$~1s during the (synchronous) switchover to a new epoch once every week or day, this is acceptable. The client ends up downloading the last day or week's beads but also will probably requesting the same info from other nodes, so this is extremely unlikely.

We can also insert heuristics to know whether we're near an epoch change, so the client knows and can react intelligently.

node/src/main.rs
```diff
@@ -598,133 +632,358 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                                            .cloned()
+                                            .map(|bead| bead.block_header.block_hash())
+                                            .collect();
+                                    }
+                                    swarm.behaviour_mut().respond_with_genesis(channel, genesis);
+                                }
+                                bead::BeadRequest::GetAllBeads => {
+                                    let all_beads;
+                                    {
+                                        let braid_lock = braid.read().await;
```
mcelrath: This lock is really bad.

node/src/main.rs
```diff
@@ -561,6 +565,21 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                                bead::BeadRequest::GetBeadsAfter(hashes) => {
+                                    let beads = braid.read().await.get_beads_after(hashes);
+                                    if let Some(response_beads) = beads {
+                                        swarm
+                                            .behaviour_mut()
+                                            .respond_with_beads(channel, response_beads);
+                                    } else {
+                                        swarm.behaviour_mut().respond_with_error(
+                                            channel,
+                                            BeadSyncError::Other(String::from(
```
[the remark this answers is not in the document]
mcelrath: agreed

node/src/ibd_manager/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+    UpdateIBDTipsCache {
+        received_tips: Vec<BeadHash>,
+        peer_id: String,
+    },
+    //Caching the received beadhashes received during `GetBead` request which will be used to fetch and extend beads in batches via `GetBead`
+    UpdateIBDGetBeadCache {
+        get_bead_response: Vec<BeadHash>,
+        peer_id: String,
+    },
+    //Update the batch offset and fetch newer batch offset window will become [offset*batchsize,(offset*batchsize)+batchsize]
```
mcelrath: This is very confusing.

node/src/ibd_manager/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+//that will be flushed out after IBD is done hence no complete dependency
+#[derive(Debug)]
+pub enum IBDCommands {
+    //Updating tips received from various sync peers that will act as the stopping window for IBD
+    UpdateIBDTipsCache {
+        received_tips: Vec<BeadHash>,
+        peer_id: String,
+    },
+    //Caching the received beadhashes received during `GetBead` request which will be used to fetch and extend beads in batches via `GetBead`
+    UpdateIBDGetBeadCache {
```
mcelrath: This naming is confusing. I'd just name it `UpdateIncoming`. It's not a "cache".

node/src/ibd_manager/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+        peer_id: String,
+    },
+    //Update the batch offset and fetch newer batch offset window will become [offset*batchsize,(offset*batchsize)+batchsize]
+    UpdateAndFetchBatchOffset {
+        peer_id: String,
+        offset_sender: tokio::sync::oneshot::Sender<usize>,
+        batch_size: usize,
+    },
+    //Fetching the cached Tips
+    FetchCachedTips {
```
mcelrath: There's no such thing as "cached" tips. There are only tips. Maybe peer tips at the beginning of IBD. But "cache" is the wrong word. Just `FetchTips` I'd say, and this has nothing to do with IBD. This should be a standard P2P message that is executed for whatever reason.  Asking for the tips of peers is good and important to know whether the network is fully connected.

node/src/ibd_manager/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+        peer_id: String,
+        tips_sender: tokio::sync::oneshot::Sender<Vec<BeadHash>>,
+    },
+    //Fetching cached beadshashes received during GetBeads
+    FetchGetBeadCache {
```
mcelrath: Again, need much clearer naming without the word "cache".

mcelrath (changes requested):

node/src/main.rs
```diff
@@ -598,133 +632,358 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                                        swarm.behaviour_mut().request_beads(peer, &pruned_beads[next_batch_offset..].to_vec());
+
+                                    }
+                                    else{
+                                        //IBD completed
+                                        tracing::info!(
+                                            peer = %peer,
+                                            "IBD has been completed with respect to peer"
+                                        );
+                                        ibd_spinlock.store(false, std::sync::atomic::Ordering::SeqCst);
```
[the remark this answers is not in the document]
Sansh2356: Current implementation is `single-sequential`  following the completion #292 another iteration for parallel sync nodes can be there after thorough discussions .

node/src/braid/mod.rs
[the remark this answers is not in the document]
OG1942: 🔥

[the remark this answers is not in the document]
OG1942: > 🔥

🔥⚡️

node/src/braid/mod.rs
Sansh2356: This means that the tips after which bead-hashes are to be sent during IBD .

node/src/braid/mod.rs
Sansh2356: The cloning is a very-very rare and edge case basically it means a fork is there so either we can stall or send all local beads .

Sansh2356: Again this will be a rare case of fork being created leading to no tips being found, so we can either error and stall or send all beads to revert the fork .

node/src/braid/mod.rs
Sansh2356: Can be discussed see above  comment .

Sansh2356: Can be discussed see the comment above .

node/src/bead/mod.rs
```diff
@@ -65,6 +57,31 @@ impl Decodable for Bead {
...
+    GetGenesis = 2,
+    GetAllBeads = 3,
+    GetBeadsAfter = 4,
+    BeadResponseError = 5,
+}
+//Converting from u8 to `RequestType`/ `ResponseType`
+impl BeadRequestType {
+    pub fn from_u8(v: u8) -> Option<Self> {
+        match v {
+            0 => Some(Self::GetBeads),
```
mcelrath: I thought I remember some helper or #[derive] that could do this for you automatically?

node/src/braid/mod.rs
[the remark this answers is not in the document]
mcelrath: This has been refactored in #299

node/src/bead/mod.rs
Sansh2356: `repr` is there but it is `unsafe` implementation as of now so i made another helper .

node/src/bead/mod.rs
```diff
@@ -65,15 +57,41 @@ impl Decodable for Bead {
     }
 }
 
+#[repr(u8)]
+#[derive(Debug, Clone, Copy)]
```
mcelrath: Please add `FromRepr` to the `derive` line, and `use strum::FromRepr`. We already have strum as a dependency (mstrr used it in the IPC code). This will auto-generate the `from_u8()` function below, which is fragile to keep in sync if we ever modify this.

node/src/bead/mod.rs
```diff
@@ -65,15 +57,41 @@ impl Decodable for Bead {
...
+pub enum BeadRequestType {
+    GetBeads = 0,
+    GetTips = 1,
+    GetGenesis = 2,
+    GetAllBeads = 3,
+    GetBeadsAfter = 4,
+    BeadResponseError = 5,
+}
```
mcelrath: Add `impl Encodable for BeadSyncError` and `impl Decodable for BeadSyncError`

Please do that for all enums that are used as network messages.

node/src/bead/mod.rs
```diff
@@ -87,52 +105,107 @@ pub enum BeadResponse {
...
+#[derive(Debug, Clone, PartialEq)]
+pub enum BeadSyncError {
+    GenesisMismatch,
+    Other(String),
+}
+
+impl Encodable for BeadSyncError {
+    fn consensus_encode<W: Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
+        match self {
+            BeadSyncError::GenesisMismatch => 0u8.consensus_encode(w),
```
mcelrath: Please use the enum symbolic names everywhere instead of 0u8 and 1u8...

`BeadSyncError::GenesisMismatch => BeadSyncErrorType::GenesisMismatch.consensus_encode(w)`

node/src/behaviour/tests.rs
```diff
@@ -184,9 +184,9 @@ async fn test_bead_request_handling() {
                                         .bead_sync
                                         .send_response(
                                             channel,
-                                            BeadResponse::Error(
+                                            BeadResponse::Error(BeadSyncError::Other(
```
mcelrath: But it IS implemented now! ;-) Please update this test.

node/src/bead/mod.rs
```diff
@@ -65,15 +57,41 @@ impl Decodable for Bead {
...
+pub enum BeadRequestType {
+    GetBeads = 0,
+    GetTips = 1,
+    GetGenesis = 2,
+    GetAllBeads = 3,
+    GetBeadsAfter = 4,
+    BeadResponseError = 5,
+}
+//Converting from u8 to `RequestType`/ `ResponseType`
+impl BeadRequestType {
```
mcelrath: Remove in favor of `#[derive(FromRepr)]` above

node/src/bead/tests.rs
```diff
@@ -274,3 +276,114 @@ fn test_codec_response_roundtrip() {
...
+        assert_eq!(response, decoded_response);
+    }
+}
+
+#[test]
+fn test_bead_sync_error_codec() {
+    // Test BeadSyncError encoding/decoding directly (not through codec)
+    let errors = vec![
+        BeadSyncError::GenesisMismatch,
+        BeadSyncError::Other("Network timeout".to_string()),
```
mcelrath: Remove the `BeadSyncError::Other` cases. Network timeout is handled by libp2p which will fire a `Event::OutboundFailure` so we don't need to handle this in BeadSyncError and we'd never be able to decode such a message anyway because it hits lip2p first.

node/src/bead/tests.rs
```diff
@@ -274,3 +276,114 @@ fn test_codec_response_roundtrip() {
...
+        BeadSyncError::Other("Invalid bead format".to_string()),
```
mcelrath: We don't need "invalid bead format" in the enum because `consensus_decode` will return an Error and that gets fired as an `io::Error` with `InvalidData`.

node/src/main.rs
```diff
@@ -598,133 +746,435 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                                                if let Some(bead) = braid_lock.beads.get(*index) {
+                                                    beads.push(bead.clone());
+                                                }
+                                            }
+                                        }
+                                    }
+                                    //Sending all the beads requested in the hashes supplied during `GetData` request
+                                    swarm.behaviour_mut().respond_with_beads(channel, beads);
+                                }
+                                bead::BeadRequest::GetTips => {
```
mcelrath: We need to check the `ibd_spinlock` here and refuse to respond with tips if we're in IBD. Add an error case `BeadSyncError::PeerSyncing` for this.

node/src/bead/mod.rs
```diff
@@ -87,52 +105,107 @@ pub enum BeadResponse {
...
+    GetBeadsAfter(Vec<BeadHash>),
     // Error response
-    Error(String),
+    Error(BeadSyncError),
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum BeadSyncError {
+    GenesisMismatch,
+    Other(String),
```
mcelrath: This `Other(String)` prevents us from using strum's `FromRepr` on this type. Instead, please:
1. Add a `BeadHashNotFound` error in this enum.
2. Use `FromRepr` and implement `consensus_encode` and `consensus_decode` for this type.

mcelrath (changes requested):

node/src/db/db_handlers.rs
```diff
@@ -88,7 +90,6 @@ impl DBHandler {
         txs_json: String,
         relative_json: String,
         parent_timestamp_json: String,
-        _ancestor_mapping: &HashMap<usize, HashSet<usize>>,
         bead_id: &usize,
```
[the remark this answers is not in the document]
Sansh2356: This was removed in favor of adapting the changes according to #299 and we were not using it for persisting anything .

node/src/main.rs
```diff
@@ -598,133 +749,478 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                                    if pruned_ref.len() <= IBD_BATCH_SIZE{
+                                        swarm.behaviour_mut().request_beads(peer, &pruned_ref);
+                                    }
+                                    else{
+                                        swarm.behaviour_mut().request_beads(peer, &pruned_ref[0..IBD_BATCH_SIZE].to_vec());
+                                    }
+
+                                }
+                                BeadResponse::Tips(tips) => {
+                                    info!(tips = ?tips, tip_count = %tips.len(), "Received braid tips");
```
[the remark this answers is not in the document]
Sansh2356: The conditions are required for splitting into appropriate batches .

node/src/main.rs
```diff
@@ -433,36 +467,151 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                                           }
+
+                                        }
+                                        else{
+                                            ibd_spinlock.store(false,Ordering::SeqCst);
+                                            continue;
+                                        }
+                                    }
+                                }
+                                else{
```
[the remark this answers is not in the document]
Sansh2356: `Beadannounce` will receive 1 bead at a request's response, and currently we are initiating with single peer so this will not be occurring leading to a potential deadlock.

node/src/main.rs
```diff
@@ -433,36 +467,151 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                                                   source = ?message.source,
+                                                   err = ?error.0,
+                                                   "An error occurred while sending insert bead command received from peer"
+                                               );
+                                           }
+                                        };
+                                        peer_manager.update_score(&message.source, 1.0);
+                                    }
+                                    for (sync_peer, ibd_ts) in timestamp_map.iter() {
+                                        let threshold = *ibd_ts + LATENCY_ALPHA * 10;
```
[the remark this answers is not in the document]
Sansh2356: It is a multiplicative constant, everything should not be a constant .

node/src/ibd_manager/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+                            Ok(_) => {
+                                tracing::info!("Sent newly initialized offset to request channel");
+                            }
+                            Err(error) => {
+                                tracing::error!(
+                                    error=?error,"Error while initiating batch offset and sending it to request channel"
+                                );
+                            }
+                        }
+                    }
```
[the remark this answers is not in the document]
Sansh2356: Irrespective of what is sent as an offset the first request is initialized without an offset hence response is not used.

node/src/main.rs
```diff
@@ -433,36 +467,151 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-                                     }
-                                 } else if let braid::AddBeadStatus::InvalidBead = status {
-                                     // update the peer manager about the invalid bead
-                                     peer_manager.penalize_for_invalid_bead(&message.source);
-                                 } else if let braid::AddBeadStatus::BeadAdded = status {
-                                     // update score of the peer
-                                     peer_manager.update_score(&message.source, 1.0);
-                                 }
+                                 if ibd_spinlock.load(Ordering::SeqCst){
+                                    let broadcast_ts = bead.uncommitted_metadata.broadcast_timestamp.clone().to_u32();
```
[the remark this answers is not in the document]
Sansh2356: This will never be the case as upperbound being u32 .

node/src/main.rs
```diff
@@ -598,133 +749,478 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                                            error!("An error occurred while sending ibd command to ibd_handler - {:?}, re-trying IBD",error.0);
+                                            match swarm_command_sender.send(SwarmCommand::InitiateIBD).await{
+                                                Ok(_)=>{
+                                                    info!("Reinitiating IBD command sent to swarm handler");
+                                                },
+                                                Err(error)=>{
+                                                    error!(error=?error,"Reinitiating IBD failed in GetAllBeads Response - ");
+                                                }
+                                            }
+                                            continue;
```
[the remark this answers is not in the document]
Sansh2356: No state update is required in case of retry and all the state updates related to retry are updated in the swarm_handler events, all the retries before the `handler` addition to mapping does only require the retry counter to be incremented other than that nothing more is required .

node/src/main.rs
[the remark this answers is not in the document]
mcelrath: I have another unfinished PR to attack timestamps...these are supposed to be all u64 microseconds since the epoch.

mcelrath:
Currently, this PR is not working to download any beads, because:
1. The boot node never exits IBD
2. Peer nodes don't serve what beads they have over IBD when they don't think they're synced.

I verified that a node in IBD correctly refuses Stratum connections.

To fix this:
1. When in IBD, a node should fire `SwarmCommand::InitiateIBD` and attempt to initiate IBD with any new incoming peer.
2. If we failed to find an IBD peer, and we *have* peers, we need to retry `SwarmCommand::InitiateIBD` on a timer.
3. Allow IBD to/from any peer, regardless of whether we're synced. If our beads are different than our peers, both peers need to IBD them in both directions.

The dynamics of the Braid are that it should never fork, and we never discard valid beads. If we had a fork, both sides of the fork need to share their beads. Then someone names both sides as parents and ties them up into a big cohort. The side with lower hashrate on the fork won't get paid, but this is expected behavior. But we have to evaluate that by actually getting those beads, computing the cohorts, and evaluating their work. We make payout decisions on beads we didn't receive so shouldn't be using any kind of timing decisions on whether we should ask for beads. We should always ask for beads.

I think we should remove the `ibd_or_not` boolean flag and the spinlock. With the above changes, it doesn't matter to peers whether we think we're synced and have decided to mine, because we'll serve beads to anyone. So we can remove the `BeadSyncError::PeerSyncing` error code I added...

So I propose:
1. Remove the P2P `ibd_spinlock` checks and simplify IBDManager:
    - Remove variables tied to the spinlock `timestamp_mapping`, `incoming_bead_mapping`
    - Remove logic in IBDManager about when to exit IBD and `ibd_or_not` (move to Stratum)
    - Remove `UpdateTimestampMapping`, `FetchTimestamp`, `FetchAllTimestamps`, `UpdateIncomingBeadMapping`, `GetIncomingBeadRetryCount`, `AbortWaitHandle`.
    - Remove related logic in main.rs lines 479-573, 1006-1041, 1241-1301
2. Remove `BeadSyncError::PeerSyncing` and every place it's returned, serve the request instead.
3. Move the `ibd_or_not` flag to stratum.rs, rename it to `ready_to_mine` or something.
    - Requires `Braid::orphanage_occupancy(10*LATENCY_ALPHA) < 0.5` or something like that.
4. Keep timestamps of when beads enter and exit `Braid::orphans`. I added an `orphan_index: HashSet<BeadHash>` to the Braid anyway in #299 because I needed to look up orphans by hash. I'll expand this to:
    - `pub orphanage: HashMap<BeadHash, (Bead, time::Instant)>` actual orphan store with O(1) lookup by hash.
    - `pub orphanage_history: VecDeque<(time::Instant, time::Instant)>` tracks entry/exit of orphans in the orphanage
    - `pub fn orphanage_occupancy(window: time::Duration) ->f64` returns the fractional occupancy of the orphanage over `window`. (includes beads currently in the orphanage)
    - I'll probably trigger pruning `orphanage_history` in extend(). Or we can set up a timer.
5. Move all the IBDManager fields into PeerInfo in `peer_manager/mod.rs`. Tracking what tips a peer has is a fundamentally peer-to-peer interaction, and I think belongs here. Add `sync_batch_offset`, `cached_tips`, and `pending_beadhashes` to PeerInfo and add the IBDManager methods to PeerInfo -- all these methods are with respect to a single peer anyway.

I will implement (4) as part of #299.

node/src/db/db_handlers.rs
```diff
@@ -288,6 +208,69 @@ impl DBHandler {
...
     }
 }
+pub fn prepare_bead_tuple_data(
+    beads: &Vec<Bead>,
+    bead_index_mapping: &HashMap<BeadHash, usize>,
+    bead: &Bead,
+) -> anyhow::Result<(String, String, String)> {
+    let mut parent_set: HashMap<usize, HashSet<usize>> = HashMap::new();
+
+    for (idx, b) in beads.iter().enumerate() {
```
mcelrath, line 218: This is incredibly expensive, and in the hot path while the braid lock is being held. 

After 299 you can just use `braid.parents` which has the data you need in O(1).

So I'm going to accept this, but we need to merge #240 then #299 then come back and fix this function.

node/src/lib.rs
```diff
@@ -370,11 +376,27 @@ impl SwarmHandler {
                     "Failed to extend Braid")
             }
         }
+
+        //Considering the index of the beads in braid will be same as the (insertion ids-1)
+        let bead_id = braid_data
```
mcelrath: This will panic if the bead is not added. Move this block of code up into the `AddBeadStatus::BeadAdded` match block.

mcelrath (changes requested):

node/src/db/db_handlers.rs
Sansh2356: Yes, agreed will replace it accordingly .

mcelrath (approved):
Looks good! Merging!

PR #253, opened 2025-09-20
DB Schema proposal

mcelrath:
All tables are append-only except Cohorts, which can be rewritten.


zaidmstrr (commented):
Concept ACK [b4cfecb](https://github.com/braidpool/braidpool/pull/253/commits/b4cfecb394fe8b916ea71340268d39c5249a9180)

Using SQLite as a database is not a bad choice. But I think you need to care about certain things in which SQLite might not work properly:
1) It handles multiple readers but only one writer at a time.
2) Not distributed: no clustering, replication, or sharding.

Other than that, it's a good choice if above limitations are not applicable to us.

mcelrath:
One writer and a single node is exactly how a Braidpool node works. It's
not hard to sequence writes of beads and this won't be a performance
bottleneck with a ~few beads/s, and there's absolutely no reason to
distribute this database. Each node creates it independently by downloading
beads.

node/src/db/schema.sql
mcelrath, line 1: -- 3. Cohorts

-- Auxiliary: positive cohort numbers (append a row for each new cohort)
CREATE TABLE CohortIds (
  id INTEGER PRIMARY KEY CHECK (id > 0)
);

-- Mapping: exactly one row per bead; cohort_id NULL means "unassigned"
CREATE TABLE Cohorts (
  bead_id   INTEGER PRIMARY KEY,           -- the bead’s id from Bead
  cohort_id INTEGER,                       -- NULL until known; else references CohortIds(id)
  FOREIGN KEY (bead_id)   REFERENCES Bead(id),
  FOREIGN KEY (cohort_id) REFERENCES CohortIds(id)
    ON UPDATE CASCADE                      -- safe if you ever renumber CohortIds
);

-- Helpful index: fast to pull all beads in a cohort and to scan for NULLs
CREATE INDEX cohorts_by_cohortid ON Cohorts(cohort_id);

-- Quick view of beads still awaiting a cohort assignment
CREATE VIEW Orphans AS
SELECT b.id
FROM Bead b
LEFT JOIN Cohorts c ON c.bead_id = b.id
WHERE c.cohort_id IS NULL;

Sansh2356:
Looks fine ACK .

mcelrath:
Merging with REGEXP support for now. @zaidmstrr 's analysis indicates this changes our inserts from ~150us to ~700us, both of which are far faster than the expected bead rate but might be a bottleneck during IBD. We may revisit the REGEXP requirement.

PR #254, opened 2025-09-24
small-fix:Updating address in config.rs

Sansh2356:


mcelrath (changes requested):
I wasn't sure what to do with this one for a while.

We need to document who has control over these wallets. For the testnet wallets it should be the node on french/dutch, so that we can share testnet coins if needed. For the mainnet one I don't know who controls either of those addresses. The original bitcoin address there was added by @zaidmstrr and has a couple transactions. Most of the testnet addresses are likely invalid though because they're the same address with the prefix changed, which will cause the checksum to fail.

Please use `bc1qrj5sqclvf877syf6wsf9qayzhcvufjcykuztlh` as the mainnet address, which I control and can be a Braidpool donation address, should we be lucky enough to get a donation or win a block with my bitaxe...

For the others, please add a documentation comment about the host and wallet name that control them.

PR #256, opened 2025-10-03
Support for segwit coinbase

Sansh2356:
- Removing the `witness` before notifying the downstream miner .
- Reconstructing `coinbase` before constructing the block with the witness commitment.

### DIFF
--- a/node/src/stratum.rs
+++ b/node/src/stratum.rs
@@ -4,13 +4,12 @@ use crate::{EXTRANONCE1_SIZE, EXTRANONCE2_SIZE, EXTRANONCE_SEPARATOR};
 use bitcoin::block::HeaderExt;
 use bitcoin::consensus::serialize;
 use bitcoin::io::Cursor;
-use bitcoin::pow::CompactTargetExt;
 use bitcoin::{
     absolute::{Decodable, Encodable},
     io::{self, Write},
     Transaction,
 };
-use bitcoin::{Block, BlockHeader, BlockTime, TxMerkleNode, Txid};
+use bitcoin::{BlockHeader, BlockTime, TxMerkleNode, Txid, Witness};
 use core::panic;
 use futures::{lock::Mutex, FutureExt};
 use rand::RngCore;
@@ -554,7 +553,7 @@ impl DownstreamClient {
         log::info!("Coinbase transaction hex: {}", hex::encode(&coinbase_bytes));
 
         let mut coinbase_cursor = Cursor::new(coinbase_bytes);
-        let coinbase_tx: Transaction =
+        let mut coinbase_tx: Transaction =
             bitcoin::Transaction::consensus_decode(&mut coinbase_cursor).unwrap();
 
         //computing merkle new merkle path due to updated coinbase transaction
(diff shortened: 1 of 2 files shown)


node/src/stratum.rs
```diff
@@ -1048,6 +1058,7 @@ pub struct JobNotification {
     pub nbits: String,
     pub ntime: String,
     pub clean_jobs: bool,
+    pub coinbase_witness: Option<Vec<Witness>>,
```
mcelrath: There's no need to store this since it's 0x00 * 32 by definition.

node/src/stratum.rs
```diff
@@ -674,11 +673,22 @@ impl DownstreamClient {
         log::info!("header.time: {}", time_be_hex);
         log::info!("header.bits: {}", bits_be_hex);
         log::info!("header.nonce: {}", nonce_be_hex);
-
-        // Construct and log the complete block using rust-bitcoin's Block struct
+        //Reconstruct our seg-wit coinbase by redoing the witness
```
mcelrath: what's the point of this block?

The coinbase has only one input with `prevout=0x00*32` and witness `0x00*32`.

node/src/stratum.rs
```diff
@@ -1102,6 +1114,20 @@ impl MiningJobMap {
         self.latest_job_id + 1
     }
 }
+///Removing witnesses from segwit transaction storing them in `JobDetails` in order to
+/// reconstruct segwit transaction while including in block as each of the downstream miner expects non-segwit coinbase only .
+fn strip_witnesses(tx: &mut Transaction) -> Vec<Witness> {
```
mcelrath: We don't need this function since the witness for the coinbase is trivial.

node/src/stratum.rs
```diff
@@ -1181,13 +1206,19 @@ impl Notifier {
             clean_job
         );
 
-        let coinbase_transaction = notified_template.transactions.get(0).unwrap();
-        let deserialized_coinbase = serialize::<Transaction>(coinbase_transaction);
+        let coinbase_transaction = notified_template.transactions.get_mut(0).unwrap();
```
mcelrath, line 1196: Instead do:
```
let mut nonsegwit_coinbase = coinbase_transaction.clone();
nonsegwit_coinbase.inputs[0].witness = Witness::new();
```

mcelrath (changes requested):

node/src/stratum.rs
Sansh2356: I did so that if we want to include more inputs in our coinbase having witness other than 0x000*32 till we have a stabilized coinbase structure .

node/src/stratum.rs
```diff
@@ -1048,6 +1054,7 @@ pub struct JobNotification {
     pub nbits: String,
     pub ntime: String,
     pub clean_jobs: bool,
+    pub coinbase_witness_commitment: Option<Vec<Witness>>,
```
mcelrath: Should be a single Witness, not a Vec.

mcelrath (approved):

PR #257, opened 2025-10-08
Forwarding shares to peers

Sansh2356:
- It provides the propagation mechanism for beads to be transferred from stratum service to the swarm via the `SwarmHandler` .
- Forwarding candidate beads to the peers after PoW validation .

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2521,11 +2521,10 @@ checksum = "241eaef5fd12c88705a01fc1066c48c4b36e0dd4377dcdc7ec3942cea7a69956"
 
 [[package]]
 name = "lock_api"
-version = "0.4.13"
+version = "0.4.14"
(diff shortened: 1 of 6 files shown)


mcelrath (changes requested):
I like this design a lot better, I think it's a lot more modular and logical.

I'd like you to go through the SwarmHandler and put in real data everywhere you can. Wherever you have to use fake data, add a FIXME or TODO marker so we can discuss how to make it not-fake. We can put those things in a new PR.

For example the miner pubkey is going to have to be a config option or stored in ~/.braidpool, and we can discuss how we want to do that. We might use the libp2p pubkey here and then we have to figure out how to sign with it.

Other things like timestamps, should be "now".

mcelrath (dismissed):
Tested, works.

PR #260, opened 2025-10-12
refactor: using rust-bitcoin functions in decode/encode_varint

zaidmstrr:
This PR uses the pre-implemented rust-bitcoin function in `decode_varint` and `encode_varint` rather than custom one.


mcelrath (approved):
Looks good!

PR #263, opened 2025-10-16
refactor:Changing the workspace structuring and pre-committed capnp rust bindings

Sansh2356:
- Removing pre-committed `capnp rust bindings` in favor build.rs .
- For accessing output or logs kindly check `/target/debug/build/node-<hash>/out` or can run `cargo build --verbose` .
- Removing older `zmq` implementations and dependencies which were not being used .
- Instantiating the workspace structure appropriately .
- Make sure `capnproto` compiler is pre-installed before building `node` crate check instructions here [capnproto_installation](https://capnproto.org/install.html) .


node/Cargo.toml
```diff
@@ -1,47 +1,41 @@
...
-num-bigint = "0.4.6"
-transitive = "1.1.0"
-secp256k1 = "0.30.0"
-libp2p = { version = "0.55.0", features = ["identify", "quic", "tokio","ping","macros","kad","dns","request-response","floodsub"] }
-toml = "0.8.22"
-jsonrpsee = {version = "0.25.1",features = ["full"]}
-async-trait = "0.1.88"
+hex = { workspace = true }
+capnp = { workspace = true }
+tokio = { workspace = true }
```
mcelrath, line 9: Why does this file remove all the version and features flags? Was that intentional? What's the consequence?

mcelrath (approved):
I love deleting code!

node/Cargo.toml
Sansh2356: Since it is the `workspace` member the deps that are native to the members i.e. share across the members are defined in the root `MANIFEST` which is `braidpool/Cargo.toml` where the versions and features flags are defined these are inherited by the workspace members, it helps in maintaining version compatibility across the sub-crates and also better organizing, if a dependency is not required by other then it can be added specifically and not inherited from `MANIFEST` .

zaidmstrr (commented):
tACK [d263b29](https://github.com/braidpool/braidpool/pull/263/commits/d263b290afcce9799dce881674510866fc1ccb79)

Tested locally, works well with the latest master. The capnp generated files also visible under `target/debug/build/node#hash/`.

Issue #262, opened 2025-10-14
Persistent state via SQlite based DB implementation


Sansh2356:
- init_db function in `init_db.rs` running before network swarm or another process to be initialized after pre-check condition of whether the db directory already exists on the desired path or not .
- Since we have all the relations constituting the braid, i have planned to keep a in memory cache for the storage of the braid status just before shutting down of the braid-node so that when the braid node is again boot up it is not needed to reconstruct the braid from the tuples that we have persisted on the disk in form of DB, since it will only be ran once during shutdown phase it will not be a problem event though serialization will be IO heavy in my opinion .
- Workers will be initialized supporting the process of version checking i.e. a snapshot of the current DB status with previous one that will help us to determine if the current DB has changed wrt any new update query and if it is the worker will then update the braid accordingly . 
- `db_handler.rs` that will be interface between `sqlite` schema connection and braid node updating and committing the changes accordingly .
- The 3rd point will also help us in efficient recovery via versioning check if any transaction fails midway and we can rollback to previous versions without going through entire log file whether it be immediate or deferred .

Sansh2356:
Refer PR #265 .

PR #265, opened 2025-10-18
Inital DB structure

Sansh2356:
- Initial DB implementation containing `DBHandler` as the query handler along with other query operations required .
- `init_db` contains initialization of disk storage of consensus data structures via `SQlite` using `sqlx` .
- `Query handler` along with extending braid appropriately.
- Addresses issue #262 .
- Schema changes namely removing `TEXT` in favor of `BLOB` along with autoincrement removal and addition of newer primary indexes for faster retrieval.
- Changing of `Bead` structure to accommodate enonce splitting . 
- Changing of trivial tests .


node/src/db/init_db.rs
```diff
@@ -0,0 +1,36 @@
+use anyhow::Result;
+use sqlx::{sqlite::SqliteConnectOptions, Executor, SqlitePool};
+use std::{env, fs, path::Path, str::FromStr};
+
+pub async fn init_db() -> Result<SqlitePool> {
+    let home_dir = env::var("HOME")?;
+    let db_dir = Path::new(&home_dir).join(".braidpool");
```
mcelrath, line 18: data path should be a config option.

node/src/db/mod.rs
```diff
@@ -0,0 +1,115 @@
+#![allow(non_snake_case)]
+pub mod db_handlers;
+pub mod init_db;
+
+///These are the types representing the tuple
+/// values for each of the relation defined under schema.sql which will be used
+/// while appending to each table
+#[derive(Debug, Clone, sqlx::FromRow, PartialEq)]
+pub struct BeadTuple {
```
mcelrath: We should just read this directly in to the Bead struct using queries rather than having this intermediate (and duplicate) type.

node/src/db/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+            weak_target: String::from(""),
+            miner_ip: String::from(""),
+            extra_nonce: String::from(""),
+            broadcast_timestamp: 1231,
+            signature: String::from(""),
+        }
+    }
+}
+#[derive(Debug, Clone, sqlx::FromRow, PartialEq)]
+pub struct TransactionTuple {
```
mcelrath: Don't make data types for everything in the db, just read in the data into our native types with queries.

node/src/error.rs
```diff
@@ -32,6 +32,43 @@ pub enum ErrorKind {
...
+impl fmt::Display for DBErrors {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            DBErrors::ConnectionToDBNotEstablished { error } => {
+                write!(f, "{:?}", error)
+            }
+            DBErrors::InsertionTransactionNotCommitted { error, query_name } => {
+                write!(
+                    f,
+                    "Insertion transaction of query {:?} failed due to - {:?}, therefore rolling-back the transaction",
```
mcelrath, line 129: Is this compatible with INSERT OR ROLLBACK?

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,861 @@
...
+        Ok((
+            Self {
+                receiver: db_handler_rx,
+                db_connection: Arc::new(Mutex::new(connection)),
+            },
+            db_handler_tx,
+        ))
+    }
+    //Insertion handlers private
+    async fn insert_bead(&self, bead_tuple: BeadTuple) -> Result<u64, DBErrors> {
```
mcelrath: Should take a Bead as an argument and below should bind to the fields of the bead.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,861 @@
...
+            .lock()
+            .await
+            .clone()
+            .begin()
+            .await
+            .unwrap();
+
+        match sqlx::query(
+            r#"
+            INSERT INTO bead (
```
mcelrath: This really needs to be a complex transaction with multiple statements and using INSERT OR ROLLBACK so that if the transaction fails, it fails atomically.

We need to simultaneously insert into the Transactions, Relatives, ParentTimestamps, and AncestorTimestamps in one transaction.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,861 @@
...
+            }
+            Err(error) => {
+                insert_parent_timestamp_tx.rollback().await.unwrap();
+                return Err(DBErrors::TupleNotInserted {
+                    error: error.to_string(),
+                });
+            }
+        };
+    }
+    async fn insert_ancestor_timestamp(
```
mcelrath: Having all these different functions will cause inserts of all the bead data (which includes transactions, timestamps, etc) to be possible to fall out of sync and corrupt the db. It needs to be all-or-nothing for all the data in a bead, its metadata, timestamps, etc.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,861 @@
...
+                insert_ancestor_timestamp_tx.rollback().await.unwrap();
+                return Err(DBErrors::TupleNotInserted {
+                    error: error.to_string(),
+                });
+            }
+        };
+    }
+    async fn insert_cohort_id(&self, cohort_id_tuple: CohortIdTuple) -> Result<u64, DBErrors> {
+        let mut insert_cohort_id_tx = self.db_connection.lock().await.begin().await.unwrap();
+        match sqlx::query("INSERT INTO CohortIds (id) VALUES (?)")
```
mcelrath: I think this isn't necessary? I think it gets automatically inserted when you insert into Cohorts? Need to check this.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,861 @@
...
+            }
+            Err(error) => {
+                insert_relative_tx.rollback().await.unwrap();
+                return Err(DBErrors::TupleNotInserted {
+                    error: error.to_string(),
+                });
+            }
+        }
+    }
+    pub async fn insert_query_handler(&mut self) {
```
mcelrath, line 160: What is this for?

Sansh2356: This will be the `query_handler` that will be handling the insertion operations, the interaction to `DB` via the `DBHandler` will be through the `sender` channel shared across threads which is thread safe (Sync+Send) and it will keep the handling in a centralized place .

node/src/db/db_handlers.rs
Sansh2356: Yes agreed i have added this as is since it is required to be a sequential insertion to be done before inserting into `Cohort` relation .

node/src/db/db_handlers.rs
Sansh2356: A sequential operation altering each table should be the ideal approach, i will alter this as in making it a serialized transaction that will eventually lead to ACID properties .

node/src/db/db_handlers.rs
Sansh2356: Same as below .

node/src/error.rs
Sansh2356: Yes .

node/src/db/mod.rs
Sansh2356: I had thought about this before doing, doing this will allow us to use inherit decode trait while fetching the tuple values otherwise we will either have to fetch attribute values one by one and that to will be `String` as the datatype which inherently satisfies the trait bounds required by `sqlx::SqliteRow` otherwise we will have to implement the decode on our own which is not a big task but it is not required in my opinion and therefore are required for the `fetch` operations, other than `fetch` write operations can be handled without these types being passed as the argument and i agree with that .

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1046 @@
...
+                    .bead_index_mapping
+                    .get(&*parent_bead_hash)
+                    .unwrap();
+                if let Some(value) = braid_parent_set.get_mut(&bead.0) {
+                    value.insert(*current_parent_bead_index);
+                }
+            }
+        }
+        let braid_children_set: HashMap<usize, HashSet<usize>> =
+            consensus_functions::reverse(&braid_data, &braid_parent_set);
```
mcelrath: This is dangerous... You're reversing the ENTIRE braid on every insert.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+
+#[derive(Debug)]
+pub struct DBHandler {
+    //Query reciever inherit to handler only
+    receiver: Receiver<BraidpoolDBTypes>,
+    //Shared across tasks for accessing DB after contention using `Mutex`
+    pub db_connection: Arc<Mutex<Pool<Sqlite>>>,
+    local_braid_arc: Arc<RwLock<Braid>>,
+}
+// The DATABASE_URL environment variable must be set at build time to a database which it can prepare queries against; the database does not have to contain any data but must be the same kind (MySQL, Postgres, etc.) and have the same schema as the database you will be connecting to at runtime.
```
mcelrath: DATABASE_URL does not seem to be used or referred to anywhere?

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+        let relatives_values = relative_tuples
+            .iter()
+            .map(|r| format!("({}, {})", r.child, r.parent))
+            .collect::<Vec<_>>()
+            .join(", ");
+
+        log::info!("{:?} ---- {:?}", relatives_values, parent_timestamps_values);
+
+        //Begin the insertion transaction
+        let sub_query_1 = "
```
mcelrath: Please merge the sub_queries into a single string query instead of assembling the query as a string on line 167. Easier to see what's going on but logically equivalent.

node/src/db/mod.rs
mcelrath: Please get rid of this struct and pass actual Beads instead.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+#[derive(Debug)]
+pub struct DBHandler {
+    //Query reciever inherit to handler only
+    receiver: Receiver<BraidpoolDBTypes>,
+    //Shared across tasks for accessing DB after contention using `Mutex`
+    pub db_connection: Arc<Mutex<Pool<Sqlite>>>,
+    local_braid_arc: Arc<RwLock<Braid>>,
+}
+// The DATABASE_URL environment variable must be set at build time to a database which it can prepare queries against; the database does not have to contain any data but must be the same kind (MySQL, Postgres, etc.) and have the same schema as the database you will be connecting to at runtime.
+//A transaction starts with a call to Pool::begin or Connection::begin.
```
mcelrath: This comment is now incorrect since we put the TRANSACTION statement in the SQL itself.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+            Self {
+                receiver: db_handler_rx,
+                db_connection: Arc::new(Mutex::new(connection)),
+                local_braid_arc: local_braid_arc,
+            },
+            db_handler_tx,
+        ))
+    }
+    //Insertion handlers private
+    pub async fn insert_sequential_insert_bead(&self, bead: Bead) -> Result<(), DBErrors> {
```
mcelrath: Please name this `insert`. It should be a simple function that simply adds a single bead to the db and nothing else.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+        let connection = match init_db().await {
+            Ok(conn) => conn,
+            Err(error) => {
+                log::error!("An error occurred while initializing and establishing connection with local DB {:?}.", error);
+                return Err(DBErrors::ConnectionToDBNotEstablished {
+                    error: error.to_string(),
+                });
+            }
+        };
+        let (db_handler_tx, db_handler_rx) = tokio::sync::mpsc::channel(1024);
```
mcelrath: Make this constant (1024) a file or crate-level constant.
```
pub const MSPC_DB_CHANNEL = 1024;
```

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+            db_handler_tx,
+        ))
+    }
+    //Insertion handlers private
+    pub async fn insert_sequential_insert_bead(&self, bead: Bead) -> Result<(), DBErrors> {
+        log::info!("Sequential insertion query received from the query handler");
+        let mut braid_parent_set: HashMap<usize, HashSet<usize>> = HashMap::new();
+        let braid_data = self.local_braid_arc.read().await;
+        //Constructing the parent set
+        for bead in braid_data.beads.iter().enumerate() {
```
mcelrath: We're passing a Bead in but actually iterating over parents by grabbing the braid.

This implies that the caller screwed up and didn't insert parents first.

Assume the caller will insert parents and instead fail if the parents aren't already in the DB.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+                    .bead_index_mapping
+                    .get(&*parent_bead_hash)
+                    .unwrap();
+                if let Some(value) = braid_parent_set.get_mut(&bead.0) {
+                    value.insert(*current_parent_bead_index);
+                }
+            }
+        }
+        //Constructing ancestor set, children set will be empty as it will become the next tip
+        let mut ancestor_mapping: HashMap<usize, HashSet<usize>> = HashMap::new();
```
mcelrath: Likewise this should have already been done by the caller. Don't do this here.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+            .get(&bead.block_header.block_hash())
+            .unwrap();
+        //Considering the index of the beads in braid will be same as the (insertion ids-1)
+        let current_bead_parent_set = braid_parent_set.get(&(bead_id)).unwrap();
+        let mut relative_tuples: Vec<RelativeTuple> = Vec::new();
+        let mut parent_timestamp_tuples: Vec<ParentTimestampTuple> = Vec::new();
+        //Constructing relatives and parent_timestamps
+        for parent_bead in current_bead_parent_set {
+            relative_tuples.push(RelativeTuple {
+                parent: (*parent_bead as i64) + 1,
```
mcelrath: Why are you adding 1 here?

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+            sub_query_4 = format!(
+        "INSERT INTO ParentTimestamps (parent, child, timestamp) VALUES {parent_timestamps_values};"
+    );
+        }
+        //Committing transaction
+        let sql_query = format!(
+            "{}{}{}{} COMMIT;",
+            sub_query_1, sub_query_2, sub_query_3, sub_query_4
+        );
+        let hex_converted_version =
```
mcelrath: Is there a reason to create these temporary, single use variables instead of putting `hex::encode(...)` directly in the bind statement below?

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+            .bind(bead.committed_metadata.comm_pub_key.to_string())
+            .bind(bead.committed_metadata.min_target.to_hex())
+            .bind(bead.committed_metadata.weak_target.to_hex())
+            .bind(bead.committed_metadata.miner_ip)
+            //TODO: Placeholder till splitting of `uncomitted_metadata` extranonce to extranonce1 + extranonce2
+            .bind("0101010101010101".to_string())
+            .bind(bead.uncommitted_metadata.broadcast_timestamp.to_string())
+            .bind(bead.uncommitted_metadata.signature.to_string())
+            .execute(&self.db_connection.lock().await.clone())
+            .await
```
mcelrath, line 128: The following block should be indented since it's an argument to `await`. I'd prefer a syntax like:
```
await { 
    log::error(!"Transaction failed to commit rolling back due to - {:?}", e); 
    [...]
}
```
Not sure why our linter or commit hooks didn't catch this. Is this indentation acceptable to rustfmt?

node/src/db/db_handlers.rs
mcelrath: Oh you have `insert_bead` below. Can we just delete this function then?

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+            }
+            Err(error) => {
+                insert_bead_tx.rollback().await.unwrap();
+                return Err(DBErrors::TupleNotInserted {
+                    error: error.to_string(),
+                });
+            }
+        };
+    }
+    async fn insert_transaction(
```
mcelrath: I don't think we have any need to insert a transaction that's NOT in a bead. It would make our db inconsistent with what's in the braid. Same for `insert_parent_timestamp` and `insert_ancestor_timestamp`.  I think we can delete these functions.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+            }
+            Err(error) => {
+                insert_cohort_tx.rollback().await.unwrap();
+                return Err(DBErrors::TupleNotInserted {
+                    error: error.to_string(),
+                });
+            }
+        };
+    }
+    async fn insert_relative_tuple(&self, relative_tuple: RelativeTuple) -> Result<u64, DBErrors> {
```
mcelrath: Again the only way this should happen is by inserting a bead. We don't want to make the db inconsistent by adding relatives when they're not specified directly in a bead. I think we can delete this function.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+                        let _res = self.insert_sequential_insert_bead(bead_to_insert).await;
+                    }
+                },
+            }
+        }
+    }
+}
+
+//Fetch handlers globally accesible
+pub async fn fetch_transactions_by_bead_id(
```
mcelrath: Why isn't this a class function DBHandler::transactions() ?

mcelrath (changes requested):

node/src/db/db_handlers.rs
Sansh2356: The reason for making those functions publicly accessible is that since i am handling the requests via a query_handler which are insert in nature that doesn't require the explicit break in the loop handling insertion events whereas the fetch loops requires the explicit return in order to return the fetch results back to the caller side that will lead to breaking of the loop, for solving this we will have to include a separate `oneshot::channel` for transferring the results back to caller side so in order to avoid this overhead i did this and also the cloning cost of connection is not much that is why i did that separation .
The read/fetch operations will be very very limited from what i have analyzed .

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+            let current_parent_timestamp = braid_data
+                .beads
+                .get(*parent_bead)
+                .unwrap()
+                .committed_metadata
+                .start_timestamp;
+            parent_timestamp_tuples.push(ParentTimestampTuple {
+                parent: (*parent_bead as i64) + 1,
+                child: (*bead_id as i64) + 1,
+                timestamp: current_parent_timestamp.to_u32().to_i64().unwrap(),
```
mcelrath: This timestamp should be a u64 (microseconds since unix epoch). Converting it to u32 will generally cause an overflow.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+            }
+        }
+    }
+}
+
+//Fetch handlers globally accesible
+pub async fn fetch_transactions_by_bead_id(
+    db_connection_arc: Arc<Mutex<Pool<Sqlite>>>,
+    bead_id: i64,
+) -> Result<Vec<TransactionTuple>, DBErrors> {
```
mcelrath: This should return `Vec<Txid>`

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+                error: error.to_string(),
+            });
+        }
+    };
+    Ok(rows)
+}
+pub async fn fetch_bead_by_bead_hash(
+    db_connection_arc: Arc<Mutex<Pool<Sqlite>>>,
+    bead_hash: String,
+) -> Result<Option<BeadTuple>, DBErrors> {
```
mcelrath: Should return `Bead`

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+            });
+        }
+    };
+
+    Ok(row)
+}
+pub async fn fetch_beads_by_cohort_id(
+    db_connection_arc: Arc<Mutex<Pool<Sqlite>>>,
+    cohort_id: i64,
+) -> Result<Vec<BeadTuple>, DBErrors> {
```
mcelrath: Should return `Vec<Bead>`

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+                error: error.to_string(),
+            });
+        }
+    };
+    Ok(rows)
+}
+pub async fn fetch_children_by_bead_id(
+    db_connection_arc: Arc<Mutex<Pool<Sqlite>>>,
+    parent_bead_id: i64,
+) -> Result<Vec<BeadTuple>, DBErrors> {
```
mcelrath: Should return `Vec<Bead>`

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,1140 @@
...
+            return Err(DBErrors::TupleNotFetched {
+                error: error.to_string(),
+            });
+        }
+    };
+    Ok(rows)
+}
+pub async fn fetch_bead_by_bead_hash(
+    db_connection_arc: Arc<Mutex<Pool<Sqlite>>>,
+    bead_hash: String,
```
mcelrath: Should be a `BeadHash` instead.

node/src/db/db_handlers.rs
Sansh2356: Yes, it passed the rustfmt` checks .

node/src/db/db_handlers.rs
Sansh2356: Yes, and additional let binding is required otherwise the intermediate result will get dropped .

node/src/db/db_handlers.rs
Sansh2356: Since we have ids from table index+1 manner .

mcelrath (changes requested):

node/src/db/db_handlers.rs
Sansh2356: This can be done although the channel capacity is still to be benchmarked .

node/src/db/db_handlers.rs
Sansh2356: Yes, agreed .

node/src/db/db_handlers.rs
mcelrath: Do we need that intermediate result? Logging? Ok

node/src/db/db_handlers.rs
Sansh2356: This will be moved under test module .

node/src/db/db_handlers.rs
Sansh2356: ACK, makes sense for this and above.

node/src/db/db_handlers.rs
Sansh2356: This will be moved under test module .

node/src/db/db_handlers.rs
Sansh2356: This will be moved under test module .

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,687 @@
...
+
+#[derive(Debug)]
+pub struct DBHandler {
+    //Query reciever inherit to handler only
+    receiver: Receiver<BraidpoolDBTypes>,
+    //Shared across tasks for accessing DB after contention using `Mutex`
+    pub db_connection: Arc<Mutex<Pool<Sqlite>>>,
+    local_braid_arc: Arc<RwLock<Braid>>,
+}
+// The DATABASE_URL environment variable must be set at build time to a database which it can prepare queries against; the database does not have to contain any data but must be the same kind (MySQL, Postgres, etc.) and have the same schema as the database you will be connecting to at runtime.
```
mcelrath: I don't think this is correct.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,687 @@
...
+//A transaction starts with a call to Pool::begin or Connection::begin.
```
mcelrath: Looks like this is copied from some documentation? We should probably remove in favor of a comment describing the DBHandler itself.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,687 @@
...
+        let connection = match init_db().await {
+            Ok(conn) => conn,
+            Err(error) => {
+                log::error!("An error occurred while initializing and establishing connection with local DB {:?}.", error);
+                return Err(DBErrors::ConnectionToDBNotEstablished {
+                    error: error.to_string(),
+                });
+            }
+        };
+        let (db_handler_tx, db_handler_rx) = tokio::sync::mpsc::channel(1024);
```
mcelrath: Make 1024 a constant like DB_MSPC_CHANNEL=1024 or something

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,687 @@
...
+            .await
+        {
+            Ok(rows) => rows,
+            Err(error) => {
+                return Err(DBErrors::TupleNotFetched {
+                    error: error.to_string(),
+                });
+            }
+        };
+        for tx_row in rows {
```
mcelrath: - High: node/src/db/db_handlers.rs:438 – fetch_bead_by_bead_hash rebuilds the bead from the in-memory braid instead of the rows it just
    fetched. On a cold start the braid is empty, so braid_data.beads.get((bead_id - 1) as usize).unwrap() will panic (and even for non-zero
    IDs the bead_id - 1 underflows for the first bead). Even if it doesn’t crash, the function breaks after the first transaction row and
    never assigns bead_txs back to fetched_bead, so you can’t hydrate stored transactions/parents from the database.

node/src/db/init_db.rs
```diff
@@ -0,0 +1,39 @@
...
+    let db_path = db_dir.join("braidpool.db");
+    let schema_path =
```
mcelrath: - High: node/src/db/init_db.rs:9 – The schema path relies on std::env::var("CARGO_MANIFEST_DIR").unwrap(). That variable is only
    guaranteed when running under cargo; binaries installed or run under systemd will panic before the DB even opens. The same function
    also uses several other unwrap()s (e.g. the SqliteConnectOptions::from_str and the first SqlitePool::connect_with) so any IO/config
    error will crash instead of surfacing a DBErrors::ConnectionToDBNotEstablished.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,687 @@
...
+            .bind(bead.block_header.bits.to_hex())
+            .bind(hex_converted_nonce)
+            .bind(bead.committed_metadata.payout_address)
+            .bind(bead.committed_metadata.start_timestamp.to_string())
+            .bind(bead.committed_metadata.comm_pub_key.to_string())
+            .bind(bead.committed_metadata.min_target.to_hex())
+            .bind(bead.committed_metadata.weak_target.to_hex())
+            .bind(bead.committed_metadata.miner_ip)
+            //TODO: Placeholder till splitting of `uncomitted_metadata` extranonce to extranonce1 + extranonce2
+            .bind("0101010101010101".to_string())
```
mcelrath: Hmmm...I guess we do nee extranonce1 and extranonce2 in uncommittedmetadata...

mcelrath (changes requested):

node/src/db/init_db.rs
Sansh2356: This is due to currently i have planned for not prod level release hence using it will be replaced afterwards .

node/src/db/db_handlers.rs
Sansh2356: Yes that is very much required as i had discussed during earlier conversation both are independent and for committing them individually we will have to alter the `commutted_metadata` stucture .

node/src/utils/test_utils.rs
```diff
@@ -197,4 +199,93 @@ pub mod test_utility_functions {
...
+            .broadcast_timestamp(time_val)
+            .extra_nonce(extra_nonce)
+            .signature(sig)
+            .build();
+        let bytes: [u8; 32] = [0u8; 32];
+
+        let test_block_header = BlockHeader {
+            version: BlockVersion::TWO,
+            prev_blockhash: BlockHash::from_byte_array(bytes),
+            bits: CompactTarget::from_consensus(439041101),
```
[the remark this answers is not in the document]
mcelrath: Please change this to minimum difficulty to avoid the inexplicable random number.

node/src/uncommitted_metadata/mod.rs
```diff
@@ -14,7 +15,20 @@ pub struct UnCommittedMetadata {
...
-
+impl Default for UnCommittedMetadata {
+    fn default() -> Self {
+        let hex = "3046022100839c1fbc5304de944f697c9f4b1d01d1faeba32d751c0f7acb21ac8a0f436a72022100e89bd46bb3a5a62adc679f659b7ce876d83ee297c7a5587b2011c4fcc72eab45";
+        let default_sig = Signature {
+            signature: secp256k1::ecdsa::Signature::from_str(hex).unwrap(),
+            sighash_type: EcdsaSighashType::All,
+        };
+        Self {
+            extra_nonce: 124562,
```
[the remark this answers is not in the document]
mcelrath: Just make this 0 to avoid the inexplicable random number.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,725 @@
...
```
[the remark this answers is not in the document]
mcelrath: Pleas just use 0 for this to avoid the inexplicable random number.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,725 @@
...
+            .bind(test_bead.block_header.bits.to_hex())
+            .bind(hex_converted_nonce)
+            .bind(test_bead.committed_metadata.payout_address)
+            .bind(test_bead.committed_metadata.start_timestamp.to_string())
+            .bind(test_bead.committed_metadata.comm_pub_key.to_string())
+            .bind(test_bead.committed_metadata.min_target.to_hex())
+            .bind(test_bead.committed_metadata.weak_target.to_hex())
+            .bind(test_bead.committed_metadata.miner_ip)
+            //TODO: Placeholder till splitting of `uncomitted_metadata` extranonce to extranonce1 + extranonce2
+            .bind("0000000000000000".to_string())
```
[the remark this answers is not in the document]
mcelrath: Please just use 0 for this to avoid the inexplicable random number.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,725 @@
...
+            .iter()
+            .map(|r| {
+                json!({
+                    "parent":r.1,
+                    "child":r.0
+                })
+            })
+            .collect::<Vec<_>>();
+
+        let final_test_query = r#"
```
mcelrath: Since this is identical to the actual query we use, please extract it to a string that can be used by both the main code query and this test.

mcelrath (changes requested):

node/src/db/init_db.rs
```diff
@@ -0,0 +1,99 @@
...
+                error: error.to_string(),
+                var: "{HOME} Directory".to_string(),
+            });
+        }
+    };
+    let db_dir = Path::new(&home_dir).join(".braidpool");
+    //Final db directory path
+    let db_path = db_dir.join("braidpool.db");
+    //Schema path to be initialized after creating db directory
+    let schema_path = Path::new(&std::env::current_dir().unwrap()).join("src/db/schema.sql");
```
mcelrath: This only works under cargo, and when the source tree is available. We couldn't distribute binaries.

Also the path is wrong, should be "node/src/db/schema.sql". 

How can we distribute the schema in a binary?

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,693 @@
...
+            .bind(bead.block_header.bits.to_hex())
+            .bind(hex_converted_nonce)
+            .bind(bead.committed_metadata.payout_address)
+            .bind(bead.committed_metadata.start_timestamp.to_string())
+            .bind(bead.committed_metadata.comm_pub_key.to_string())
+            .bind(bead.committed_metadata.min_target.to_hex())
+            .bind(bead.committed_metadata.weak_target.to_hex())
+            .bind(bead.committed_metadata.miner_ip)
+            //TODO: Placeholder till splitting of `uncomitted_metadata` extranonce to extranonce1 + extranonce2
+            .bind("0101010101010101".to_string())
```
mcelrath: This number is larger than i32::MAX so would cause a ParseIntError if ever saved to the db. Please make it 0.

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,693 @@
...
+            Err(error) => {
+                return Err(DBErrors::TupleNotFetched {
+                    error: error.to_string(),
+                });
+            }
+        };
+        for tx_row in rows {
+            let _txid = tx_row.get::<String, _>("txid");
+            let bead_id = tx_row.get::<u64, _>("bead_id");
+            let indexed_bead = braid_data.beads.get((bead_id) as usize).unwrap();
```
mcelrath: This will panic on restart when the braid is empty. We should be using fetched_bead here instead of indexed_bead.

Add a test for this scenario since this function isn't called (yet).

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,693 @@
...
+            .bind(test_bead.block_header.bits.to_hex())
+            .bind(hex_converted_nonce)
+            .bind(test_bead.committed_metadata.payout_address)
+            .bind(test_bead.committed_metadata.start_timestamp.to_string())
+            .bind(test_bead.committed_metadata.comm_pub_key.to_string())
+            .bind(test_bead.committed_metadata.min_target.to_hex())
+            .bind(test_bead.committed_metadata.weak_target.to_hex())
+            .bind(test_bead.committed_metadata.miner_ip)
+            //TODO: Placeholder till splitting of `uncomitted_metadata` extranonce to extranonce1 + extranonce2
+            .bind("0000000000000000".to_string())
```
mcelrath: Also unneccesarily large, it's just an int so please make it 0.

mcelrath (changes requested):
Please add a hook in main.rs to load the braid from disk on startup, before starting networking.

node/src/db/db_handlers.rs
Sansh2356: Length constraint has to be satisfied hence .

node/src/db/db_handlers.rs
Sansh2356: Test is already there and it will never be called if braid is empty so it will not panic as the braid will be constructed before initialising connection to db .

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,664 @@
...
+                })
+            })
+            .collect::<Vec<_>>();
+
+        let relatives_values = relative_tuples
+            .iter()
+            .map(|r| {
+                json!({
+                    "parent":r.1,
+                    "child":r.0
```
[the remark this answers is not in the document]
Sansh2356: No, it is right wrt schema .

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,664 @@
...
+            return Err(DBErrors::TupleNotFetched {
+                error: error.to_string(),
+            });
+        }
+    };
+    // println!("bead id {}", bead_id);
+    if bead_id != -1 {
+        //Fetching transactions from DB using unhex to get raw hex from blob
+        let rows = match sqlx::query(
+            "SELECT  unhex(txid) as txid,bead_id FROM Transactions WHERE bead_id = ?",
```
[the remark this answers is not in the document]
Sansh2356: No, this has been tested and mentioned clearly in Sqlite documentation .

node/src/db/db_handlers.rs
```diff
@@ -0,0 +1,664 @@
...
+            fetched_bead.committed_metadata.min_target = min_target;
+            fetched_bead.committed_metadata.start_timestamp = start_timestamp;
+            fetched_bead.committed_metadata.weak_target = weak_target;
+            fetched_bead.uncommitted_metadata.broadcast_timestamp = broadcast_timestamp;
+            fetched_bead.uncommitted_metadata.extra_nonce_1 = extranonce_1;
+            fetched_bead.uncommitted_metadata.extra_nonce_2 = extranonce_2;
+            fetched_bead.uncommitted_metadata.signature = signture;
+            Ok(())
+        })
+        .fetch_optional(&db_connection_arc.lock().await.clone())
```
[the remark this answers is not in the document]
Sansh2356: No, it requires to be owned .

Issue #275, opened 2025-11-03
Add a Shell Script for Easy Dashboard Boot-Up & Troubleshooting


its-me-abhishek:
**Is your feature request related to a problem?** 
Sometimes running the Braidpool Dashboard can feel a bit hectic for users who aren’t very familiar with web development setups. Managing multiple processes (API, Simulator, Frontend, etc.) manually can be error-prone and confusing if done without proper guidance.

**Describe the solution you'd like**
Add a shell script inside the dashboard/ folder that automatically boots up all the necessary services; similar to how Docker simplifies setups, but without actually using Docker.

1. The script can use tmux (or similar) to run multiple components in parallel terminal  windows.
2. It can perform environment variable checks before starting. Maybe even provide some option to create an env file and update variables 
3. It should support verbose logging to help troubleshoot issues like missing env vars, port conflicts, or build errors.

**Describe alternatives you've considered**
1. Have been Docker Compose to manage services (but that adds a Docker dependency, which might not be ideal for everyone ).
2. Running individual terminal sessions manually, which is tedious and inconsistent for new users.

**Additional context**
This script could make local development much smoother, especially for users who just want to test or visualize the dashboard without diving into setup details. It may even evolve later into a small CLI helper for Braidpool dashboard over the course.

Shivrajsoni:
hi @its-me-abhishek  can u assign me this issue ?

Shivrajsoni:
Hi @its-me-abhishek,

The PR is ready for review. I had a quick question regarding the Python version we’re using.
Python 3.14 was causing dependency compatibility issues on my system, so I switched to Python 3.11.9, which works correctly. I’ve also updated the script to explicitly use this version.

Please let me know if you have any suggestions or prefer a different approach regarding the Python version.

Thanks!

Shivrajsoni:
Hi @its-me-abhishek,
I’ve reviewed and implemented the changes you mentioned. I have a question regarding the Python version — it’s currently compatible with any 3.12.*. Could you please share your thoughts on this?

The PR is ready for merge.

Janmesh23:
hi @its-me-abhishek !
i would like to work on this issue , can you please assign me this issue

Abhishek-Punhani:
Hey @its-me-abhishek , Is this issue still open ?

Sansh2356:
@Abhishek-Punhani this is not the relevant issue now but we do required docker setup for dashboard side cc @priyashuu .

[...]

PR #279, opened 2025-11-04
enhancement: Adding a shell script for easy dashboard boot-up and troubleshooting

Shivrajsoni:
#275 
 
This pull request introduces a new shell script, dashboard/start-dev.sh, designed to dramatically simplify the local development setup for new contributors. It automates the entire process of dependency validation, environment setup, and service initialization, reducing the setup to a single command.

  What Was Done

  A new Bash script, start-dev.sh, was created to serve as a single entry point for running the Braidpool dashboard and its required
  simulator_api.py service.

  The script's start command performs the following actions:
   1. Tooling Validation: Checks if tmux, python3, and node are installed, providing helpful installation instructions if they are missing.
   2. Python Environment Setup:
       * Verifies that the python3 version is exactly 3.11.9 to ensure a consistent environment.
       * Creates a local Python virtual environment (.venv) in the tests/ directory if one doesn't already exist.
       * Installs all required Python packages from tests/requirements.txt into the virtual environment.
   3. Frontend Setup:
       * Installs all necessary frontend dependencies by running npm install in the dashboard/ directory.
   4. Service Initialization:
       * Starts a new tmux session named braidpool-dev.
       * Launches the Python simulator and the frontend development server in separate, named tmux windows.

  The script is idempotent, meaning it can be run multiple times without re-installing dependencies or re-creating environments that are already in place. It also provides stop and attach commands for easy session management

Approach and Rationale

  The primary goal was to create a user-friendly, "one-click" setup experience that minimizes manual configuration and potential errors for
   new developers.

   * Why a Shell Script? 
   *A shell script was chosen for its low dependency footprint (only requiring tmux, which is a common developer tool)
     and its ability to easily orchestrate commands across different language ecosystems (Python, Node.js).
   * Why `tmux`? tmux was used to run the multiple processes (simulator, frontend) in parallel while keeping their logs separate and easily
     accessible in a single terminal session. This provides a clean and organized way to monitor all moving parts of the application.
   * Why Virtual Environments? The script creates a Python virtual environment to encapsulate dependencies. This is a crucial best practice
     that prevents conflicts with the user's global packages and ensures the project has the exact dependencies it needs.
   * Why Idempotency? By making the setup steps idempotent, the script is fast and efficient on subsequent runs, only  performing actions that are strictly necessary.

### DIFF
--- a/dashboard/start-dev.sh
+++ b/dashboard/start-dev.sh
@@ -1,0 +1,159 @@
+#!/bin/bash
+
+# A script to manage the Braidpool development environment using tmux.
+#
+# This script provides a simple way to start, stop, and attach to all the
+# services needed for local Braidpool development. It automatically handles
+# checking dependencies and setting up the Python and Frontend environments.
+#
+# Usage:
+#   ./start-dev.sh start   - Sets up environment and starts all services.
+#   ./start-dev.sh stop    - Stops all services by killing the tmux session.
+#   ./start-dev.sh attach  - Attaches to the running tmux session.
+
+# --- Configuration ---
+set -e # Exit immediately if a command exits with a non-zero status.
+
+SESSION_NAME="braidpool-dev"
+REQUIRED_PYTHON_VERSION="3.11.9"
+TESTS_DIR="../tests"
+VENV_DIR="$TESTS_DIR/.venv"
+
+# Commands to run in each tmux window. These are defined after setup.
+SIMULATOR_CMD="$VENV_DIR/bin/python $TESTS_DIR/simulator_api.py"
+FRONTEND_CMD="npm run dev"
+
+# --- Setup and Validation Functions ---
+
+# Checks for all required command-line tools.
+check_tool_deps() {
+  echo "--- Checking for required tools ---"
+  for tool in tmux python3 node; do
+    if ! command -v $tool &>/dev/null; then
+      echo "❌ Error: '$tool' is not installed. Please install it to continue." >&2
+      if [ "$tool" == "tmux" ]; then
+        echo "   For macOS: brew install tmux" >&2
+        echo "   For Debian/Ubuntu: sudo apt-get install tmux" >&2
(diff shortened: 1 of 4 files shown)


dashboard/start-dev.sh
```diff
@@ -0,0 +1,159 @@
+#!/bin/bash
```
its-me-abhishek, line 1: The script does not have the commands to run the node `api`, please include that. Also maybe we should not hardcode the Python version (unsure on that though)

dashboard/start-dev.sh
```diff
@@ -0,0 +1,159 @@
...
+set -e # Exit immediately if a command exits with a non-zero status.
+
+SESSION_NAME="braidpool-dev"
+REQUIRED_PYTHON_VERSION="3.11.9"
+TESTS_DIR="../tests"
+VENV_DIR="$TESTS_DIR/.venv"
+
+# Commands to run in each tmux window. These are defined after setup.
+SIMULATOR_CMD="$VENV_DIR/bin/python $TESTS_DIR/simulator_api.py"
+FRONTEND_CMD="npm run dev"
```
its-me-abhishek, line 24: this is the command to run dev build, though users interested in self hosting shall run the prod build for better performance, maybe we can provide a prompt/option to choose it via the CLI of this script itself

its-me-abhishek (changes requested):
Please change the target branch to `dev` instead of main, and then update this, as the package-lock seems to be affected due to this issue

dashboard/start-dev.sh
Shivrajsoni: Please can you give me idea of about commands to run node api ?

its-me-abhishek: Do checkout https://github.com/braidpool/braidpool/tree/dev/dashboard/api

its-me-abhishek: The readme contains the commands, and .env.example has a sample .env file for fast setup!

PR #291, opened 2025-11-11
Switch logging to the tracing library

### LINKED ISSUE #165: Integer Overflow in flexbuffers::Reader Triggers Crash
### Public Disclosure
While writing a fuzz target for `Message::from_bytes`  in `node/src/protocol.rs` I found an Integer Overflow vulnerability in the FlexBuffers crate. I already disclosed this vulnerability in Discord communication [channel](https://discord.com/channels/1270418429412642957/1270423991068266672/1370828169342488718), but I didn't  talk about steps to reproduce it. So in this issue I will describe how to reproduce the issue in a controlled environment.

### Steps to Reproduce
1. Go to the file `node/src/protocol.rs`.
2. Inside unit tests add this new regression test there.

 ```sh
    #[test]
    fn test_flexbuffers_crash() {
        // The crash input found by the fuzzer
        let crash_input = &[247, 255, 255, 7, 1, 1, 255, 255, 1, 255, 255, 255, 255, 255, 255, 255, 
                        255, 255, 255, 255, 255, 255, 255, 255, 0, 0, 23, 2];
        
        println!("Testing crash input - this should cause a panic inside catch_unwind");
        
        // Wrap in catch_unwind to prevent the test from panicking
        let result = std::panic::catch_unwind(|| {
            Message::from_bytes(crash_input)
        });
        
        // Print what happened
        match &result {
            Ok(res) => println!("No panic occurred! Result: {:?}", res),
            Err(_) => println!("Confirmed: The input caused a panic as expected"),
        }
        
        // Verify that this input indeed causes a panic
        assert!(result.is_err(), "The crash input should cause a panic");
    }

```
3. Run this test as: `RUST_BACKTRACE=1 cargo test test_flexbuffers_crash -- --nocapture`

You all have the stack traces about the error for further investigation.

### PR DESCRIPTION
**Pull request by mcelrath**
Switch to the `tracing` library instead of `log` for systematic logging:
1. libp2p already uses it
2. Logs crate/module emitting the message
3. Replace all `println!()` statements with equivalent `tracing::info/error/debug` statements
4. Remove `env_logger` dependency
5. Removed `log` dependency in favor of `tracing-log` to intercept calls to `log::debug!` (etc) and forward them to the tracing module, for dependencies that use `log`.
6. Systematically use structured logging instead of string interpolation
7. Add `strum` as a dependency for better logging of errors involving enums (for `BitcoinRequest`, mostly in `ipc/client.rs`).
8. Propagate `template_id` and `block_hash` through the stratum and ipc interfaces for logging
9. Systematically use `template_id` as an integer instead of mixed string/int usage
10. Remove flexbuffers dependency (we removed it long ago) Fixes #165 
11. More clearly log Braid::extend and log failure to extend

Here's a list of direct dependencies that use `log`:
1. `sqlx` - Database connectivity (via sqlx-core, sqlx-sqlite)
2. `bitcoincore-rpc` - Bitcoin Core RPC client
3. `libp2p` - Peer-to-peer networking (via many sub-crates)
4. `jsonrpsee` - JSON-RPC client/server (via jsonrpsee-http-client)
5. `bitcoincore-zmq` - Bitcoin Core ZMQ bindings (via async_zmq → mio)

Here's a list of indirect dependencies that use `log`:
  - `async-io` - Used by sqlx-core
  - `async-std` - Used by sqlx
  - `kv-log-macro` - Used by async-std
  - `hyper-rustls` - HTTPS support for jsonrpsee
  - `if-watch` - Network interface monitoring for libp2p
  - `mio` (v0.6) - Async I/O for bitcoincore-zmq
  - `multistream-select` - Protocol negotiation for libp2p
  - `netlink-proto` & `netlink-sys` - Linux netlink for if-watch
  - `polling` - Polling mechanisms for async-io
  - `rtnetlink` - Linux routing for if-watch
  - `rustls` - TLS implementation (multiple dependents)
  - `rustls-platform-verifier` - Platform certificate verification
  - `soketto` - WebSocket for jsonrpsee
  - `tokio` - Core async runtime
  - `tracing-log` - The bridge itself (used by tracing-subscriber)
  - `tracing-subscriber` - tracing infrastructure

This PR is targeting the `15-10-25-db-integration` (#265) to minimize the diff. I will merge #265 first and change the target of this PR after merging it.

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -24,8 +24,6 @@ bitcoin_hashes = { workspace = true }
 bitcoincore-rpc = { workspace = true }
 bitcoincore-rpc-json = { workspace = true }
 bitcoincore-zmq = { workspace = true }
-env_logger = { workspace = true }
-log = { workspace = true }
 shellexpand = { workspace = true }
 num = { workspace = true }
 rand = { workspace = true }
@@ -35,6 +33,7 @@ toml = { workspace = true }
 jsonrpsee = { workspace = true }
 async-trait = { workspace = true }
 sqlx = { workspace = true}
+strum = { workspace = true }
 anyhow = "1.0.100"
 
 
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -35,6 +35,7 @@ async-trait = { workspace = true }
 sqlx = { workspace = true}
 strum = { workspace = true }
 anyhow = "1.0.100"
+if-addrs = "0.13"
 
 
 [build-dependencies]
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -16,7 +16,6 @@ tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
 futures = { workspace = true }
 bytes = { workspace = true }
-flexbuffers = { workspace = true }
 serde_json = { workspace = true }
 clap = { workspace = true }
 bitcoin = { workspace = true }
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -33,8 +33,8 @@ jsonrpsee = { workspace = true }
 async-trait = { workspace = true }
 sqlx = { workspace = true}
 strum = { workspace = true }
-anyhow = "1.0.100"
-if-addrs = "0.13"
+if-addrs = { workspace = true }
+anyhow = { workspace = true }
 
 
 [build-dependencies]
--- a/node/build.rs
+++ b/node/build.rs
@@ -21,7 +21,7 @@ fn main() {
                     Ok(entry) => {
                         let path = entry.path();
                         if path.extension().and_then(|ext| ext.to_str()) == Some("capnp") {
-                            println!("cargo:warning=Compiling schema file: {:?}", path);
+                            println!("cargo:info=Compiling schema file: {:?}", path);
                             //The output path for rust bindings is default `OUT_DIR` -
                             if let Err(e) = capnpc::CompilerCommand::new()
                                 .src_prefix("schema")
--- a/node/build.rs
+++ b/node/build.rs
@@ -21,7 +21,7 @@ fn main() {
                     Ok(entry) => {
                         let path = entry.path();
                         if path.extension().and_then(|ext| ext.to_str()) == Some("capnp") {
-                            println!("cargo:info=Compiling schema file: {:?}", path);
+                            println!("cargo:warn=Compiling schema file: {:?}", path);
                             //The output path for rust bindings is default `OUT_DIR` -
                             if let Err(e) = capnpc::CompilerCommand::new()
                                 .src_prefix("schema")
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -449,7 +449,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
                                              ),
                                          );
                                      } else {
-                                         warn!("No peers available to request parent beads");
+                                         warn!(parent_count = %bead.committed_metadata.parents.len(), "Insufficient peers for bead sync");
                                      }
                                  } else if let braid::AddBeadStatus::InvalidBead = status {
                                      // update the peer manager about the invalid bead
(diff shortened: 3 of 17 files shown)

### CODE REVIEW

**Review by zaidmstrr — COMMENTED** [review]
Looks good, things like `[instrument]` are really beneficial for debugging.

**Review by Sansh2356 — DISMISSED** [review]
LGTM @mcelrath .

File: node/src/main.rs [hunk 1]
```diff
@@ -65,22 +66,25 @@ async fn main() -> Result<(), Box<dyn Error>> {
     //Reconstructing local braid upon startup
     let db_connection_pool_ref = _db_handler.db_connection_pool.clone();
     let braid_ref = braid.clone();
+    // FIXME instead we should look 144 blocks back from the bitcoin tip (1 day) and load beads
```
Sansh2356, line 69: Can you elaborate further on this @mcelrath , since `IBD` is als currently structured to deal with no genesis bead .

node/src/stratum.rs
Sansh2356, line 1042: `DownstreamClient` must have a `new` implementation apart from initializing a new client interface from `default`, just noting for a future reference .

node/build.rs
```diff
@@ -21,7 +21,7 @@ fn main() {
                     Ok(entry) => {
                         let path = entry.path();
                         if path.extension().and_then(|ext| ext.to_str()) == Some("capnp") {
-                            println!("cargo:warning=Compiling schema file: {:?}", path);
+                            println!("cargo:info=Compiling schema file: {:?}", path);
```
Sansh2356: These logs are visible in the default `OUT_DIR` `env` variable  we can keep it at `warn` level also .

node/Cargo.toml
```diff
@@ -35,7 +32,9 @@ toml = { workspace = true }
 jsonrpsee = { workspace = true }
 async-trait = { workspace = true }
 sqlx = { workspace = true}
+strum = { workspace = true }
 anyhow = "1.0.100"
```
Sansh2356: This dependency can be removed .

node/Cargo.toml
```diff
@@ -35,7 +32,9 @@ toml = { workspace = true }
...
+if-addrs = "0.13"
```
Sansh2356: Can we add this by inheriting from  `workspace` instead ?

Sansh2356 (dismissed):
ACK @mcelrath  [`2ec04dd`](https://github.com/braidpool/braidpool/pull/291/commits/2ec04dd56f013259ba3b6bde84238f23b86d883a), just few comments perhaps for discussion purposes apart from that can you squash some of the commits into a parent commit ?

node/src/main.rs
mcelrath: This is about how many beads we load into memory. One day seems like a reasonable number to keep in memory.

Also this block needs to be moved down, below where we check/create the data directory, because that's where braidpool.db lives. Presumably this barfs if ~/.braidpool hasn't been created yet?

node/src/stratum.rs
mcelrath: You mean instead of using `DownstreamClient::default()` and then modifying it, you want to use a `::new()` method? That's a good idea.

node/build.rs
mcelrath: The warning was just annoying me, and it seemed it was recompiling with every run of cargo, violating the configuration we have "cargo:rerun-if-changed". It doesn't seem to be doing that anymore though, I'll put it back to warn.

node/Cargo.toml
mcelrath: This is used in ipc.rs to turn the `BitcoinRequest` enum into a string for logging purposes.

mcelrath:
> ACK @mcelrath [`2ec04dd`](https://github.com/braidpool/braidpool/pull/291/commits/2ec04dd56f013259ba3b6bde84238f23b86d883a), just few comments perhaps for discussion purposes apart from that can you squash some of the commits into a parent commit ?

I have github configured to do a squash & merge automatically.

PR #294, opened 2025-11-12
Feat: added shell script for automation

Shivrajsoni:
#275 
Shell Script for setup of dashboard ,
Screenshot 2025-11-12 at 8 40 42 PM

python 3.12.* required 
tmux required

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 #Virtual workspace not creating a root package along with workspace as not needed .
 [workspace]
-members = ["node"]
+members = ["node", "braidpool-cli"]
 resolver = "3"
 
 [workspace.package]
@@ -42,5 +42,6 @@ sqlx = {version="0.8.6",features=[ "sqlite",
 strum = { version = "0.26", features = ["derive"] }
 if-addrs = "0.13"
 anyhow = "1.0.100"
+reqwest = { version = "0.11", features = ["json"] }
 
 
--- a/braidpool-cli/Cargo.toml
+++ b/braidpool-cli/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "braidpool-cli"
+version.workspace = true
+edition.workspace = true
+
+[dependencies]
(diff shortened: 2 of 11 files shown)


zaidmstrr (commented):
Looks like the whole code is AI-generated.

mcelrath:
> Looks like the whole code is AI-generated.

I'm not necessarily opposed to that, but it requires heavier review. Fortunately the AI's are better at review than writing code. 

I'm not really happy about tying it to tmux though. Ultimately we want all this to run as a systemd service.

zaidmstrr:
> I'm not necessarily opposed to that, but it requires heavier review.

I'm also not opposed to that, but some of them write blindly without knowing its internal workings.

mcelrath (changes requested):
@Shivrajsoni thank you for your contribution!

I'd rather not merge anything to do with the simulator -- this was just for demos and development. I'd rather people spend their time adding the features we need to fully deprecate the simulator. See #298 for a list of RPCs we need -- which includes getting info about the braid which is what the simulator was doing.

If you'd like to get this merged, please:
1. Remove the tmux dependency.
2. Remove the python version check. There's no reason our code won't run with other versions of python. I use 3.13.
3. Remove the venv creation/check. Instead we should update `tests/README.md` if it is not accurate. Python venvs are a nightmare and people have different ways of handling it. I use direnv, others use uv/hatch/conda for venv management and we don't want to be telling people how they should be doing this.
4. Add arguments for the node, especially `--network` and default to `cpunet` for now.

I do think we need to add a script like this, because starting the front end is somewhat of a bear, and I'd love to add your contribution!

For comparison, I have a `bootnode.sh` script I've been using on our test servers to start the boot node (backend only). You can find it here:

https://github.com/mcelrath/braidpool/blob/sprint-1-cli-implementation/bootnode.sh

It's also written by AI, because never in my life will I ever be able to remember bash syntax. 🤣 I've been really reluctant to merge AI generated code so far, but I have been using it heavily.

PR #245, opened 2025-08-22
Add `descendant work` and `hwp` algorithm

Aritra8438:
This PR adds the `descendant work` and `hwp` algorithm.

PR #299, opened 2025-11-21
Braid refactor: Close the loop

mcelrath:
On auditing the braid code, I found much of it was stuck in a confusion we had a few months ago where we considered adding the braid algorithms directly to the braid class. For historical context, that's where I started, with a [python class implementing these algorithm](https://github.com/mcelrath/braidcoin). After getting confident that they were correct, I refactored them into the algorithms [you now find in tests/braid.py](https://github.com/braidpool/braidpool/blob/main/tests/braid.py).

Long story short, we decided to go in a direction that did not add these functions to the braid class (it requires it to work with hashes among other things, instead of integer indices) and decided to go with an approach that used indices into the `Vec<Bead>` held by the Braid, which in this PR I've added as a type alias `BeadIdx` for `usize`. This allowed us to keep the algorithmic functions basically unmodified and fast, but some confusion persisted in the code.

I have refactored a number of things here and fixed a few logic errors. A more complete changelog is in the git commit but here's what it is right now.

I started by looking at how to lock cohorts but discovered that there were a number of other things that needed to be fixed. I'm still (re)considering how braid locking works. and may move the COHORTS and THREADS globals into another PR upon further consideration.

Obsoletes and should close #241 and #245.

```
        - Remove hashet_to_vec and hashset_to_vec_deterministic in favor of in-line computation
        - Move BeadHash type into crate::bead
        - Add BeadIdx type alias for usize in crate::braid (this doesn't
          make sense in crate::bead because it's an index into a
          Vec<BeadIdx> held by the Braid.
        - Add Relatives = HashMap<BeadIdx, HashSet<BeadIdx>> alias and relatives!
            macro for convenient testing
        - Change Cohorts to be a simple type alias for HashSet<BeadIdx> instead of a one-element struct
        - Added parents, children to the Braid
            - Update extend() and some tests to maintain these
        - Removed cohort_tips, this seems to be a misunderstanding, cohorts don't have
            tips (they have heads and tails though, but there's no reason to
            record cohort tips globally)
        - Call algorithms::cohort() within extend()
        - Fix sub_braid (was remapping indices)
        - Fix extend (was popping all cohorts and recomputing *everything*)
        - Refactor consensus_functions into braid/algorithms.rs and all
          algorithmic tests (from python) into algorithm_tests.rs.
        - Remove the Braid parameter from all consensus_functions/algorithms
            This came from an intermediate state where we were considering putting
            these functions on the Braid class itself. Since we decided to go with
            index mappings in the Vec<Bead> in the braid, the consensus_functions can
            exactly match the original python and work on a parents dict instead.
            I think this caused some confusion.
        - Renamed lots of things to match the python code and be more concise:
            genesis_beads -> geneses
            bead_index_mapping -> index

        - Add COHORTS and THREADS as static globals across threads to isolate cohorts
            computation
```


mcelrath:
This PR is an example of what happens if you're not careful while vibe coding. While it contains some very important contributions, it's also a complete mess and far too large. Here's my own AI self-review on it:

● Review complete. The PR mixes protocol‑level timestamp changes, a major braid refactor, frontend redesign, a new miner API,
  and benchmarks; there are a few blocking issues plus a frontend bug.

   - Critical: Consensus encoding for timestamps switches from u32 seconds to u64 microseconds with no versioning/migration, 
  which will break wire compatibility for existing serialized data.
  node/src/committed_metadata/mod.rs   pub struct TimeVec(pub Vec<MicrosecondTimestamp>);
    ...
    len += time.as_micros().consensus_encode(w)?;
   
   node/src/uncommitted_metadata/mod.rs similarly encodes MicrosecondTimestamp.
   - High: DB layer still persists timestamps as seconds (to_u32) while in‑memory types are microseconds, so data is 
  truncated and won’t round‑trip correctly.
  node/src/db/db_handlers.rs   .bind(bead.committed_metadata.start_timestamp.to_u32())
    .bind(bead.uncommitted_metadata.broadcast_timestamp.to_u32())
   - High: Tests won’t compile: hashset_to_vec_deterministic was removed but is still imported, and tests still assume 
  MedianTimePast types.
  node/src/committed_metadata/tests.rs   use crate::utils::hashset_to_vec_deterministic;
    ...
    assert_eq!(metadata.start_timestamp, MedianTimePast::MIN);
   - Medium: Miner IDs become unstable when MAC is missing; using Date.now() causes the table/history to churn each refresh.
  dashboard/src/components/MinerInventory/MinerInventoryDashboard.tsx   id: data.mac || ip + '_' + Date.now(),

  Suggested PR split (well‑defined feature sets):

   1. Timestamp protocol upgrade: MicrosecondTimestamp, CommittedMetadata/UnCommittedMetadata, serialization changes, 
  schema/migration updates, and tests/data updates.
   2. Braid algorithm refactor: new braid/algorithms.rs, Braid structure/caches, algorithm tests, utils/test_utils macro 
  changes.
   3. RPC/Braid integration cleanups: Bead::hash() usage, RPC server tweaks, AddBeadStatus mapping updates.
   4. Miner API service: new miner_api/ FastAPI service, requirements, README.
   5. Dashboard miner inventory redesign: new MinerInventory components, analytics charts, NodeHealth tweaks, URLs update.
   6. Benchmarks & workspace changes: benchmarks/ crate additions, workspace Cargo.toml/Cargo.lock updates, ancillary 
  scripts.

Issue #301, opened 2025-11-28
Build failed with the error[E0658] in branch dev


zaidmstrr:
Currently building and compiling on top of the latest dev (58b4f8a) branch giving the error: 

```sh
error[E0658]: use of unstable library feature `inherent_str_constructors`
   --> node/src/db/db_handlers.rs:494:34
    |
494 |             let payout_address = str::from_utf8(&row.get::<Vec<u8>, _>("payout_address"))
    |                                  ^^^^^^^^^^^^^^
    |
    = note: see issue #131114 <https://github.com/rust-lang/rust/issues/131114> for more information
```

A simple fix is in `db_handlers`:

```diff
- let payout_address = str::from_utf8(&row.get::<Vec<u8>, _>("payout_address"))
-     .unwrap()
-     .to_string();
+ let payout_address = String::from_utf8(row.get::<Vec<u8>, _>("payout_address"))
+           .map_err(|_| DBErrors::TupleAttributeParsingError {
+                error: "Invalid payout_address UTF-8".into(),
+               attribute: "payout_address".into(),
+            })?;
```
I think the new version is appropriate and will not panic and log a clear and concise message.

Rustix69:
@zaidmstrr Can I look into it ???

zaidmstrr:
Yes, you can.

mcelrath:
How did this happen? I was able to build before merge.

zaidmstrr:
> How did this happen? I was able to build before merge.

In @Sansh2356 system this is also not reproducible. But I'm facing the above error while building the source code. Maybe some particular OS system or versions are triggering this.

zaidmstrr:
I found the exact problem about this issue and it was mentioned in [#131114](https://github.com/rust-lang/rust/issues/131114) of the official rust issue tracker. In some Rust versions, calling `str::from_utf8` without the `std:: `prefix and without importing `std::str `triggers an instability error `(inherent_str_constructors)`. The function itself is stable, but the shorthand `str::from_utf8` only works when `std::str` is in scope or in newer Rust versions where this was stabilized.

Sansh2356:
Nice, @zaidmstrr i have also thought about `MSRV` and will do this it in upcoming PRs after proper discussions that will stabilize the usage across users and we can then perform periodic bumps in versioning that will help in dealing with these unwanted bugs .

PR #308, opened 2025-12-06
fix: stabilize str::from_utf8 by adding std:: prefix

zaidmstrr:
This fixes issue [#301](https://github.com/braidpool/braidpool/issues/301) by prefixing `std::` before `str` to make it stable across all Rust versions.


Sansh2356 (dismissed):
ACK @zaidmstrr .

Sansh2356 (approved):

PR #309, opened 2025-12-06
Orphan bead handling wrt DB 

### LINKED ISSUE #306: Refactor IBD
Our IBD PR #240 got a bit long and it took me a long time to see a better way, but here's a proposal.

The dynamics of the Braid are that it should never fork, and we never discard valid beads. If we had a fork, both sides of the fork need to share their beads. Then someone names both sides as parents and ties them up into a big cohort. The side with lower hashrate on the fork won't get paid, but this is expected behavior. But we have to evaluate that by actually getting those beads, computing the cohorts, and evaluating their work. We can't make payout decisions on beads we didn't receive so shouldn't be using any kind of timing decisions on whether we should ask for beads. We should always ask for beads.

I think we should remove the ibd_or_not boolean flag and the spinlock. With the above changes, it doesn't matter to peers whether we think we're synced and have decided to mine, because we'll serve beads to anyone. So we can remove the BeadSyncError::PeerSyncing error code I added...

So I propose:

1.    Remove the P2P ibd_spinlock checks and simplify IBDManager:
    -    Remove variables tied to the spinlock timestamp_mapping, incoming_bead_mapping
    -    Remove logic in IBDManager about when to exit IBD and ibd_or_not (move to Stratum)
    -    Remove UpdateTimestampMapping, FetchTimestamp, FetchAllTimestamps, UpdateIncomingBeadMapping, GetIncomingBeadRetryCount, AbortWaitHandle.
    -    Remove related logic in main.rs lines 479-573, 1006-1041, 1241-1301
 2.   Remove BeadSyncError::PeerSyncing and every place it's returned, serve the request instead.
 3.   Move the ibd_or_not flag to stratum.rs, rename it to ready_to_mine or something.
    -    Requires Braid::orphanage_occupancy(10*LATENCY_ALPHA) < 0.5 or something like that.
 4.   Keep timestamps of when beads enter and exit Braid::orphans. I added an orphan_index: HashSet<BeadHash> to the Braid anyway in #299 because I needed to look up orphans by hash. I'll expand this to:
        - pub orphanage: HashMap<BeadHash, (Bead, time::Instant)> actual orphan store with O(1) lookup by hash.
        - pub orphanage_history: VecDeque<(time::Instant, time::Instant)> tracks entry/exit of orphans in the orphanage
        - pub fn orphanage_occupancy(window: time::Duration) ->f64 returns the fractional occupancy of the orphanage over window. (includes beads currently in the orphanage)
     -   I'll probably trigger pruning orphanage_history in extend(). Or we can set up a timer.
5.  Move all the IBDManager fields into PeerInfo in peer_manager/mod.rs. Tracking what tips a peer has is a fundamentally peer-to-peer interaction, and I think belongs here. Add sync_batch_offset, cached_tips, and pending_beadhashes to PeerInfo and add the IBDManager methods to PeerInfo -- all these methods are with respect to a single peer anyway.

I will implement (4) as part of https://github.com/braidpool/braidpool/pull/299.

This should substantially simplify the IBD code, which was largely based on an "Am I synced" decision which is inappropriate in an asynchronous network. You're never synced in an asynchronous network, but always synced, there's always something in flight you don't know about. Bitcoin intentionally slowed this down to 10 minutes and created a synchronous blockchain data structure. We need to be faster and the more-correct question is "are we seeing everything and can we connect everything we see in newly broadcast beads" which is the purpose of the orphanage concept, and occupancy thereof. It's a measure of "am I missing broadcasts". And "seeing all broadcasts" is the "I am synced" signal.

After discussing with @Sansh2356 he wants to add the following minor features in the corresponding PR as well:
1. Move `prepare_bead_tuple_data()` out of main and into the DBHandler.
2. Consolidate the DB lock around main.rs:475-478: only hold the lock through extend and then examine what beads were added to the braid by looking at `braid_data.beads.len()` and copying them out to pass to the DBHandler, releasing the lock earlier.

Sansh2356:
- Currently the `orphan` beads were not being persisted in DB leading to foreign key violation and causing error in subsequent bead propagation .
- This PR adds the functionality for connecting the `Braid` with the `db_handler` using `db_tx` and thus persisting the beads if any present in `orphan buffer` as soon as all of its parents have been received . 
- It also removes request in `BeadSync` for the beads whose parents have not yet been received i.e. it is orphan bead which was redundant in nature causing unnecessary utilization of bandwidth .
- Changes verbosity of logs and updating the tests accordingly .

The above was initial though but over the discussions me @mcelrath arrived at a conclusion that tight coupling b/w DB and `Braid` is not required and must be kept as separated as possible .

- Removal of locks passed on for `Braid` into main thread only reducing the causes as well as checking of potential race conditions arising eventually leading to deadlocks. 
- Orphan persistence using the simple length check instead of anything else .
- Removal of `Timestamp` based checking for `IBD` for the purpose of connecting DAG after receiving beads from sync nodes as explained under #306 and `ibd_manager` moving into `peer_manager`.
- The mining condition is to be moved into `stratum` component instead depending dynamically over `Orphan-Occupancy` above certain threshold `alpha` defined as latency param which will eventually done in forth-coming PR  and adding batched insertions during IBD.
- This also adds IBD retry to be initiated per request basis for respective timeouts to switch between sync peers to continue IBD from the previous state.
- Adding ibd_flag for preventing downstream connections.
- Replacing the `Arc<Mutex<T>>` wrapping around db connection pool to avoid unnecessary locking along with the clearing IBD buffers in case of stale or dropped and retried requests and clearing buffer after ibd is completed.
- It also adds batched insertions along with sequential insertions for both orphan beads as well as the parent  beads.


mcelrath (changes requested):
This PR creates a circular dependency, Braid <-> DBHandler.

In main.rs:517 around `AddBeadStatus::BeadAdded` we are writing new beads to the DB on a successful extend (which will automatically also add orphans to the braid if possible). So all we really need to do is update main.rs to save not only the new bead, but instead all new beads in the braid since the last save. 

For instance, Before calling `braid_data.extend(&bead)` at main.rs:477, save `let braid_prev_len = braid_data.beads.len()` and then when writing to the DB, write all beads between `braid_prev_len` and the new `braid_data.beads.len()`.

node/src/db/db_handlers.rs
```diff
@@ -208,69 +344,6 @@ impl DBHandler {
         }
     }
 }
-pub fn prepare_bead_tuple_data(
```
mcelrath, line 211: This function has moved in the file, making comparison difficult. Unless it's really important that it appear at this place in the file, please move your modified implementation back to this location.

node/src/db/mod.rs
```diff
@@ -8,10 +10,9 @@ pub mod init_db;
 pub enum InsertTupleTypes {
     InsertBeadSequentially {
         bead_to_insert: Bead,
-        txs_json: String,
-        relative_json: String,
-        parent_timestamp_json: String,
-        bead_id: usize,
+        curr_beads: Vec<Bead>,
```
mcelrath: Why do we need `bead_to_insert`, `curr_beads` and `removed_orphans` too? Why not just one `Vec<Bead>`?

node/src/db/mod.rs
```diff
@@ -8,10 +10,9 @@ pub mod init_db;
...
+        bead_index_mapping: HashMap<bitcoin::BlockHash, usize>,
```
mcelrath: Please use the `BeadHash` and `BeadIdx` types.

mcelrath (changes requested):
Overall, this seems overly complex to me. I don't understand why `prepare_bead_tuple_data` takes so many arguments, its job is just to make json tuples for insert. Why does it take so many arguments?

Likewise InsertBeadSequentially should probably just be called `InsertBeads` and it should take a `Vec<Bead>`. Where they need to get things like BeadIdx, why just not pass in a reference to the Braid?

node/src/db/mod.rs
Sansh2356: `bead_to_insert` is required for bead_id as it is not necessary that bead_id will always be equal to last_idx of  `Vec<Bead>` of `Braid`, we can directly pass on `bead_id` instead .

node/src/peer_manager/mod.rs
```diff
@@ -26,6 +32,17 @@ pub struct PeerInfo {
     pub connected: bool,
     /// The peer's IP address
     ip_addr: Option<IpAddr>,
+    /// Tracks the most recent tips received from each peer during IBD.
+    /// (peer_id --> Vec<BeadHash>)
+    tips_mapping: Vec<BeadHash>,
```
zaidmstrr: The names here are confusing and don't represent the true nature of the variable use. Same for `batch_mapping`, `get_bead_mapping`, choose a name which truly states that these variables are used for IBD and for the specific behaviour.

node/src/peer_manager/mod.rs
```diff
@@ -1,8 +1,14 @@
+use crate::utils::BeadHash;
 use libp2p::PeerId;
 use std::collections::{HashMap, HashSet};
 use std::net::IpAddr;
 use std::time::{Duration, Instant};
 
+pub const IBD_BATCH_SIZE: usize = 500;
+pub const MAX_IBD_RETRIES: u64 = 10;
+pub const IBD_TRIGGER_AFTER: u64 = 20;
```
zaidmstrr: Why are we waiting for 20 sec in `IBD_TRIGGER_AFTER`? I think instead of waiting for a fixed period of time we can use some ready signal which will allow us to trigger the IBD. Example signal ready after peer discovery, etc.

node/src/db/db_handlers.rs
```diff
@@ -169,12 +214,25 @@ impl DBHandler {
                 BraidpoolDBTypes::InsertTupleTypes { query } => match query {
                     InsertTupleTypes::InsertBeadSequentially {
```
zaidmstrr, line 623: Here the beads are added sequentially to the persistence db even when we are fetching the beads in batches during IBD. What if we add another type which allows us to add the beads in batches to the db? We can do this by first validating all the beads using `braid.extend()` and then using a single transaction for the db to add all the beads at once. And elsewhere (without IBD), sequential bead insert is sufficient.

zaidmstrr (commented):
I reviewed this PR and found some areas of improvement and gave some suggestions. You are free to discuss them below.

node/src/peer_manager/mod.rs
Sansh2356: Will do the changes .

node/src/peer_manager/mod.rs
Sansh2356: Yeah, we can do minimum connected peer and stall until that threshold is not achieved .

node/src/db/db_handlers.rs
```diff
@@ -126,39 +219,399 @@ impl DBHandler {
...
+            Err(err) => {
+                error!("Failed to begin DB batch transaction: {}", err);
+                return Err(DBErrors::ConnectionToSQlitePoolFailed {
+                    error: err.to_string(),
+                });
+            }
+        };
+
+        let mut inserted_count = 0u32;
+        let beads_ref = beads.clone();
```
zaidmstrr: The variable `beads_ref` is computationally expensive with no real benefits and needs to be completely removed from the database. The only purpose it serves is logging.

node/src/db/db_handlers.rs
```diff
@@ -126,39 +219,399 @@ impl DBHandler {
...
+                        }
+                        if let Ok(relatives) =
+                            serde_json::from_str::<Vec<serde_json::Value>>(&relatives_json)
+                        {
+                            all_relatives_json_parts.extend(relatives);
+                        }
+                        if let Ok(parent_ts) =
+                            serde_json::from_str::<Vec<serde_json::Value>>(&parent_ts_json)
+                        {
+                            // warn!("PARENTS JSON STRING - {:?}",parent_ts);
```
zaidmstrr: This line needs to be removed.

node/src/db/db_handlers.rs
```diff
@@ -126,39 +219,399 @@ impl DBHandler {
...
+        Ok(())
+    }
+    fn prepare_bead_tuple_data(
+        &self,
+        bead_index_mapping: &HashMap<BeadHash, (usize, u32)>,
+        bead: &Bead,
+    ) -> anyhow::Result<(String, String, String)> {
+        let bead_id = bead_index_mapping
+            .get(&bead.block_header.block_hash())
+            .unwrap()
```
zaidmstrr: It's good to not use `unwrap` here.

node/src/db/db_handlers.rs
```diff
@@ -126,39 +219,399 @@ impl DBHandler {
...
+            .0;
+        let mut current_parents = Vec::new();
+        let mut relatives = Vec::new();
+        let mut parent_ts = Vec::new();
+        for parent in &bead.committed_metadata.parents {
+            let (parent_index, parent_timestamp) = bead_index_mapping.get(parent).unwrap();
```
zaidmstrr: You can also remove the `unwrap` from here.

node/src/db/db_handlers.rs
```diff
@@ -126,39 +219,399 @@ impl DBHandler {
...
+                    debug!(
+                        "Json relatives - {:?} \n",
+                        all_relatives_json_parts.get(original_bead_id.0)
+                    );
+                    debug!(
+                        "Original bead  -- {:?} \n Fetched bead by its bead hash from DB- {:?}\n",
+                        original_bead, bead
+                    );
+                    for parent_hash in &original_bead.committed_metadata.parents {
+                        let id = bead_index_mapping.get(parent_hash).unwrap();
```
zaidmstrr: Same here you can remove `unwrap()`.

zaidmstrr (commented):
I have reviewed the PR and found some issues. You can look at them and resolve them and discuss them.

node/src/db/db_handlers.rs
Sansh2356: I have addressed all the unwraps in `docker-test-env` branch already, i will pick that commit here .

mcelrath:
● Blockers (detail)

   1. IBD queue/tips are unbounded and never reset  
    - Where: node/src/peer_manager/mod.rs around handle_update_ibd_* and handle_fetch_ibd_* (lines ~149–188).  
    - What happens: ibd_bead_queue and ibd_peer_tips are only extended and then cloned for fetch; they’re not cleared or
   capped.  
    - Why this is a blocker: A peer can keep sending large batches, causing memory growth and duplicate bead hashes. 
  Retries then re‑request already‑seen ranges, and queues keep growing—classic DoS surface and correctness risk.  
    - Fix direction: Replace vs extend (or drain after fetch), cap at IBD_BATCH_SIZE, and reset per‑peer IBD state on 
  retry/complete.
   2. Batch insert can re‑insert resolved orphans  
    - Where: node/src/main.rs around batch IBD insert (lines ~951–955).  
    - What happens: InsertBeadsBatch is called with all received beads plus removed_orphans. Orphans that were resolved 
  earlier in the batch can appear twice.  
    - Why this is a blocker: DB has UNIQUE/PK on bead hash; duplicates can trigger a rollback and stall IBD.  
    - Fix direction: Only include AddBeadStatus::BeadAdded in the batch list and dedupe against removed_orphans.
   3. Partial tuple prep leaves DB inconsistent  
    - Where: node/src/db/db_handlers.rs around the bulk insert path (lines ~331–371).  
    - What happens: The Bead row is inserted even if prepare_bead_tuple_data fails; related rows 
  (Relatives/ParentTimestamps/Transactions) are skipped.  
    - Why this is a blocker: Leaves orphaned Bead rows without relations—breaks invariants and later queries.  
    - Fix direction: Validate tuple prep first or roll back the bead row; keep insert and tuple creation in a single 
  transaction that fails atomically.
   4. HWP comparator errors masked  
    - Where: node/src/braid/mod.rs around highest_work_path (line ~1068).  
    - What happens: comparator errors are treated as equal, which can make HWP nondeterministic.  
    - Why this is a blocker: HWP is consensus‑critical; nondeterminism risks divergence.  
    - Fix direction: propagate/handle comparator errors explicitly.
   5. IBD pruning order risks skipping beads  
    - Where: node/src/main.rs around IBD pruning/queue creation (line ~1059).  
    - What happens: pruning stops at tips and iterates cohorts in an unordered way; hashes may be truncated and beads 
  skipped.  
    - Why this is a blocker: IBD can miss beads, causing inconsistent DAG state.

  ---------------------------------------------------------------------------------------------------------------------

  Failed tests (detail)

  cargo test -p node → 95 passed, 8 failed.

  AddrInUse failures (port already bound):

   - behaviour::tests::test_bead_request_handling (node/src/behaviour/tests.rs:94)
   - behaviour::tests::test_floodsub_message_propagation (node/src/behaviour/tests.rs:289)
   - rpc_server::test_extend_rpc (node/src/rpc_server.rs:1044)
   - rpc_server::test_get_mining_info_rpc (node/src/rpc_server.rs:1044)
   - rpc_server::test_get_ipc_stats_rpc_simple (node/src/rpc_server.rs:1044)
   - rpc_server::test_get_tips_cli_flow (node/src/rpc_server.rs:1044)
   - rpc_server::test_unstage_transactions_rpc_simple (node/src/rpc_server.rs:1044)

  Meaning: These tests try to bind fixed ports and collided with existing listeners. This is environmental flakiness,
  not necessarily caused by PR code. We can rerun with isolated ports or ensure no lingering services are running.

  Logic failure:

   - stratum::test::server_start_test (node/src/stratum.rs:2119)
  Assertion expected 3, got 6. This looks like a behavioral change from PR #309 (likely more connections/events 
  recorded), and needs code‑level investigation.

Sansh2356:
> ● Blockers (detail)
> 
>     1. IBD queue/tips are unbounded and never reset
>        - Where: node/src/peer_manager/mod.rs around handle_update_ibd_* and handle_fetch_ibd_* (lines ~149–188).
>        - What happens: ibd_bead_queue and ibd_peer_tips are only extended and then cloned for fetch; they’re not cleared or
>        capped.
>        - Why this is a blocker: A peer can keep sending large batches, causing memory growth and duplicate bead hashes.
>        Retries then re‑request already‑seen ranges, and queues keep growing—classic DoS surface and correctness risk.
>        - Fix direction: Replace vs extend (or drain after fetch), cap at IBD_BATCH_SIZE, and reset per‑peer IBD state on
>        retry/complete.
> 
>     2. Batch insert can re‑insert resolved orphans
>        - Where: node/src/main.rs around batch IBD insert (lines ~951–955).
>        - What happens: InsertBeadsBatch is called with all received beads plus removed_orphans. Orphans that were resolved
>        earlier in the batch can appear twice.
>        - Why this is a blocker: DB has UNIQUE/PK on bead hash; duplicates can trigger a rollback and stall IBD.
>        - Fix direction: Only include AddBeadStatus::BeadAdded in the batch list and dedupe against removed_orphans.
> 
>     3. Partial tuple prep leaves DB inconsistent
>        - Where: node/src/db/db_handlers.rs around the bulk insert path (lines ~331–371).
>        - What happens: The Bead row is inserted even if prepare_bead_tuple_data fails; related rows
>        (Relatives/ParentTimestamps/Transactions) are skipped.
>        - Why this is a blocker: Leaves orphaned Bead rows without relations—breaks invariants and later queries.
>        - Fix direction: Validate tuple prep first or roll back the bead row; keep insert and tuple creation in a single
>        transaction that fails atomically.
> 
>     4. HWP comparator errors masked
>        - Where: node/src/braid/mod.rs around highest_work_path (line ~1068).
>        - What happens: comparator errors are treated as equal, which can make HWP nondeterministic.
>        - Why this is a blocker: HWP is consensus‑critical; nondeterminism risks divergence.
>        - Fix direction: propagate/handle comparator errors explicitly.
> 
>     5. IBD pruning order risks skipping beads
>        - Where: node/src/main.rs around IBD pruning/queue creation (line ~1059).
>        - What happens: pruning stops at tips and iterates cohorts in an unordered way; hashes may be truncated and beads
>        skipped.
>        - Why this is a blocker: IBD can miss beads, causing inconsistent DAG state.
> 
> 
> Failed tests (detail)
> 
> cargo test -p node → 95 passed, 8 failed.
> 
> AddrInUse failures (port already bound):
> 
>     * behaviour::tests::test_bead_request_handling (node/src/behaviour/tests.rs:94)
> 
>     * behaviour::tests::test_floodsub_message_propagation (node/src/behaviour/tests.rs:289)
> 
>     * rpc_server::test_extend_rpc (node/src/rpc_server.rs:1044)
> 
>     * rpc_server::test_get_mining_info_rpc (node/src/rpc_server.rs:1044)
> 
>     * rpc_server::test_get_ipc_stats_rpc_simple (node/src/rpc_server.rs:1044)
> 
>     * rpc_server::test_get_tips_cli_flow (node/src/rpc_server.rs:1044)
> 
>     * rpc_server::test_unstage_transactions_rpc_simple (node/src/rpc_server.rs:1044)
> 
> 
> Meaning: These tests try to bind fixed ports and collided with existing listeners. This is environmental flakiness, not necessarily caused by PR code. We can rerun with isolated ports or ensure no lingering services are running.
> 
> Logic failure:
> 
>     * stratum::test::server_start_test (node/src/stratum.rs:2119)
>       Assertion expected 3, got 6. This looks like a behavioral change from PR [Orphan bead handling wrt DB  #309](https://github.com/braidpool/braidpool/pull/309) (likely more connections/events
>       recorded), and needs code‑level investigation.

All the tests are passing can checkout to branch and run `cargo test -p node`. 

Apart from that the provided suggestions 1 i already had mind for draining all the required states after IBD has been completed, apart from that in 3 i agree that `continue` should be replaced with hard-errors leading to db-component being shut-down since any disorder will lead to non-consistent state for every forward state so this will be replace by `DBErrors` instead and 4 the error must be replaced.


Apart from the above 3 remaining 2 are not called for, respectively because the `removed_orphan` is for a single batch only hence cannot occur twice for the same batch and in the other received beads are in order so pruning them according to the tips difference will and should be handled by the `cohort` as well as `extend` functionalities respectively .

node/src/peer_manager/mod.rs
```diff
@@ -91,23 +122,154 @@ impl PeerManager {
...
+            peer_info.ibd_peer_tips.clear();
+            peer_info.ibd_bead_queue.clear();
+            peer_info.ibd_batch_offset = IBD_BATCH_SIZE;
+            peer_info.ibd_inflight_request = None;
+        }
+    }
+
+    /// Record an in-flight IBD request to `peer_id`. Called immediately after
+    /// `bead_sync.send_request(...)` returns its `OutboundRequestId`.
+    pub fn set_ibd_inflight(&mut self, peer_id: PeerId, request_id: OutboundRequestId) {
```
zaidmstrr, line 163: The function has no guard to hold the correct invariant. Thus, can include the possible race conditions. Also, this is called in many places inside `main.rs` under independent code blocks.

node/src/braid/mod.rs
zaidmstrr, line 99: The `orphan_beads` Vec is an unbounded buffer and has no checks on the capacity limits. Any malicious peer can send any number of malformed beads with unknown parents and can crash the buffer with OOM.

node/src/main.rs
```diff
@@ -152,28 +150,20 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-                error!(error=?error,"An error occurred while initiating IBD after waiting for peer discovery - ");
-            }
-        };
-    });
+
+    // IBD will be triggered when peer count threshold is reached
+    let mut ibd_initiated = false;
+
+    // Flag to indicate IBD completion - prevents miner connections until synced
+    let ibd_complete = Arc::new(AtomicBool::new(false));
```
zaidmstrr, line 158: The variable is never set to `false` after being used or setting it to true. Which can introduce unexpected behaviour later in the code.

zaidmstrr (commented):
I found some possible issues that needs to taken a look. While the reviewing process is still going on from my side. Until then you can look at the issues I have found.

mcelrath:
This report summarizes the open findings from the re-review after rebasing onto the updated upstream.

1. **Security | High | `node/src/main.rs:901` – Unbounded `GetAllBeads` response.** Any peer can request the full DAG, which clones and serializes all beads. **Impact:** peer-triggered CPU/RAM exhaustion and outbound bandwidth spikes. **Fix:** remove or gate `GetAllBeads`, use paged `GetBeadsAfter` only, and enforce a hard response-size cap before encoding.

Bob's take: valid. Use `GetBeadsAfter` only with paging. Add a config setting which limits the number of simultaneous peers we will serve IBD to. (Bitcoin uses 8 "full relay" peers)

2. **Database | High | `node/src/main.rs:1007` – Duplicate inserts in batch.** `removed_orphans` and `beads_to_insert` are concatenated without de-duplication, so resolved orphans can appear twice. **Impact:** Bead PK/UNIQUE violations and full batch rollback, stalling IBD. **Fix:** de-dup by `BeadHash` before batching (preferred) or use `INSERT OR IGNORE` with guarded dependent inserts.

Bob's take: valid. IMHO the whole flow is overly complex. You have three data structures representing "beads to be added" (`beads_to_insert`, `removed_orphans`, and InsertBeadSequentially from floodsub). Ideally there should be just ONE interface, `InsertBeads(Vec<Bead>)` Also consider using an automatically de-duped data structure like HashSet or HashMap.

3. **Performance | Critical | `node/src/main.rs:607/1007` – Orphan-tail cloning under braid write lock.** Per-bead cloning of orphan tail and bead-index mapping occurs under the write lock, producing O(n^2) work. **Impact:** long lock holds and latency spikes as DAG grows. **Fix:** avoid tail cloning per bead; have braid return an explicit resolved-orphans set and process outside the write lock.

Bob's take: valid. The simplest fix to this is to have `extend()` return the list of beads added **including** any resolved orphans, rather than trying to guess about orphans after the fact. (main.rs:589 and main.rs:988)

4. **Performance | High | `node/src/braid/mod.rs:197` – O(n^2) orphan processing.** `process_orphan_beads` removes from a Vec and recurses, repeatedly scanning/shifting. **Impact:** quadratic time on large orphan backlogs. **Fix:** use a queue (`VecDeque`) plus lookup (`HashSet`/index map), or an iterative retain/partition pass.

Bob's take: This is #299 territory, those algorithmic improvements need to be a separate PR and should be fixed there.

5. **Performance | Medium | `node/src/db/db_handlers.rs:539` – N+1 orphan reinserts.** `InsertBeadSequentially` reinserts removed orphans one-by-one with serialization and DB I/O. **Impact:** sync lag for large orphan tails. **Fix:** batch orphan reinserts using the bulk insert path.

Bob's take: valid. Build a vector and use `InsertBeadsBatch` instead. Sequential inserts in SQL is very slow compared to batch inserts.

6. **Rust | High | `node/src/main.rs:282` – `expect` on behaviour creation.** `BraidPoolBehaviour::new(...).expect(...)` panics on startup failures. **Impact:** non-recoverable crash instead of error propagation. **Fix:** return/propagate the error (e.g., map to `io::Error`).

Bob's take: valid. Simple fix.

7. **Rust | Medium | `node/src/main.rs:1469` – `expect` on hardcoded targets.** Parsing `"1d00ffff"` uses `expect`. **Impact:** runtime panic if parsing changes. **Fix:** use `CompactTarget::from_consensus(0x1d00ffff)` or handle the `Result`.

Bob's take: valid. Just switch to `CompactTarget::from_consensus()` instead. Simple fix. Correct fix is correctly deferred with the explicit TODO.

8. **Rust | Medium | `node/src/uncommitted_metadata/mod.rs:23` – Default signature `expect`.** Default metadata parses a hardcoded signature with `expect`. **Impact:** runtime panic if constant changes or parse fails. **Fix:** use pre-validated constant bytes or a fallible constructor.

Bob's take: this is explicitly deferred functionality. Just add a TODO here too.

9. **Rust | Medium | `node/src/committed_metadata/mod.rs:95` – Default pubkey `expect`.** Default metadata parses a hardcoded pubkey with `expect`. **Impact:** runtime panic in default construction. **Fix:** use pre-validated constant bytes or a fallible constructor.

Bob's take: same as above, just add a TODO here.

10. **Rust | Medium | `node/src/utils/mod.rs:84` – `create_test_bead` in production module.** Public function uses test-only constants and `expect`. **Impact:** accidental prod usage with panic paths. **Fix:** move under `#[cfg(test)]` or into `utils::test_utils`.

Bob's take: valid. This should be part of the test framework under `#[cfg(test)]`

11. **Architecture | Medium | `node/src/main.rs:607` – Orphan persistence inferred via slice.** Resolved orphans are inferred from `beads[bead_id+1..]`, coupling main to braid internals. **Impact:** misidentified resolved orphans and fragile persistence logic. **Fix:** have braid return an explicit `resolved_orphans` set.

Bob's take: same as #3.

12. **Architecture | Medium | `node/src/db/db_handlers.rs:599` – DB insert failures not propagated.** Insert errors are logged and dropped. **Impact:** in-memory braid can diverge from DB silently. **Fix:** propagate errors to sync logic and retry/backpressure or fail the batch.

Bob's take: valid. Return `Err` from the handler and have the caller apply backpressure or retry or crash loudly. (e.g. pause IBD if the error is transient). This is a correctness issue. We can't allow the db to diverge from the in-memory representation under any circumstances. It becomes an immediate consensus failure.

13. **Architecture | Medium | `node/src/main.rs:980` – IBD orchestration split across main and PeerManager.** Sync logic lives in main while state lives in PeerManager. **Impact:** tighter coupling and harder testing of IBD flow. **Fix:** reintroduce an IBD manager/state machine with explicit APIs.

Bob's take: This is probably the biggest issue with this PR. PeerManager holds IBD‑specific state (cached bead queues, inflight request tracking, offsets, reset logic), while main.rs drives the actual IBD flow: initiating IBD, validating responses, computing next batch offsets, sending next requests, and handling error paths (disconnect, reset, retry). This means you can't reason about the IBD lifecycle in a single place, and it makes unit testing almost impossible (requiring a full swarm event loop). This has implicit coupling: main has to know how PeerManager's internal IBD fields behave and what resets are required. (more below)

14. **Cryptography | Low | `node/src/main.rs:1469` (and metadata defaults) – Placeholder pubkey/signature/targets.** Mining broadcast still uses hardcoded pubkey/signature and fixed targets. **Impact:** not spec-compliant; unsafe to treat as production-ready. **Fix:** wire to configured miner pubkey, real signatures, and targets derived from difficulty.

Bob's take: explicit future work. Add TODO's where appropriate so we find the right sites when this is implemented in a follow-on PR.

I think I may have made the suggestion to deprecate `ibd_manager` and I think it was a mistake (sorry) and the AI is right on this. Here's some concrete suggestions:
1. IBDManager should own only paging/requests/retries/inflight state. It should not parse/validate/insert beads. Add systematic unit tests for IBDManager. This will remove most of the issues around requiring the entire Swarm to test and make it easier to review and merge.
2. Create a new `ingest_beads()` function that handles validation, braid extension, orphan resolution, DB insert, and scoring. It becomes a one-shot, if it doesn't error, memory and DB are consistent and we're in consensus. If it does, it needs either retry logic or a loud failure and a crash. The application cannot reliably continue.
3. Delete `GetAllBeads` and use only `GetBeadsAfter` with paging. Delete `InsertBeadSequentially` and use only `InsertBeadsBatch` that takes a vector.
4. `extend()` should return the inserted beads.
5. Centralize peer scoring inside `ingest_beads()`. Right now it's split across floodsub and IBD handlers. Compute invalid/added counts once in the ingest path and update the score there.
6. Do DB and scoring work outside the Braid lock.

PR #312, opened 2025-12-08
Remove template retry logic in IPC listener

### LINKED ISSUE #300: Remove template retry logic in `ipc.rs`
We don't need a retry mechanism for the templates for now. The current code retries for the template if it contains bytes less than the `MIN_TEMPLATE_SIZE` and makes 3 attempts to call the `getblocktemplate` function; if the size is still < `MIN_TEMPLATE_SIZE` it chooses the last template. For the new behaviour, completely remove these reattempts and choose the template in the first try. You can contain `MIN_TEMPLATE_SIZE` variant (for correctness) but don't forget to reduce its size, maybe to 256 bytes?

s373nZ:
This PR attempts to close #300 by removing the template retry logic from the IPC listener.

- Renames `get_template_with_retry()` => `get_template()` and removes the `max_attempts` parameter.
- Reduces the `MIN_TEMPLATE_SIZE` constant from `512` to `256`.
- Emits warning if template is lower than `MIN_TEMPLATE_SIZE`, but accepts the block anyway.

Closes #300

Disclaimer: First time contributor here and don't know much about Braidpool architecture (yet?), but saw #300 as a "good first issue" and thought to take a crack at it.

### DIFF
--- a/node/src/ipc.rs
+++ b/node/src/ipc.rs
@@ -114,13 +114,11 @@ pub async fn ipc_block_listener(
 
             // Only try to get initial template if node is synced
             if is_synced {
-                match get_template_with_retry(
+                match get_template(
                     &mut shared_client,
-                    3,
                     RequestPriority::High,
                     "initial template",
                     tip_height,
-                    0,
                     network,
                 ).await {
                     Ok(template) => {
@@ -176,13 +174,11 @@ pub async fn ipc_block_listener(
                                 match shared_client.is_initial_block_download(Some(RequestPriority::High)).await {
                                     Ok(in_ibd) => {
                                         if !in_ibd { // Node is synced (not in IBD)
-                                            match get_template_with_retry(
+                                            match get_template(
                                                 &mut shared_client,
-                                                2,
                                                 RequestPriority::High,
                                                 &format!("block {}", height),
                                                 height,
-                                                0,
                                                 network,
                                             ).await {
                                                 Ok(template) => {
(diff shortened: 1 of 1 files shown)


node/src/ipc.rs
```diff
@@ -385,144 +383,77 @@ pub async fn ipc_block_listener(
...
-                                warn!(
-                                    context = %context,
-                                    size_bytes = %hex.len(),
-                                    attempt = %attempt,
-                                    max_attempts = %max_attempts,
-                                    "Template too small - retrying"
-                                );
-                            }
-                        }
+    match client
```
zaidmstrr: You can replace these `match` statement with  `?` operator, this will propagate any error better when `get_block_template_componets` return with some error.

node/src/ipc.rs
```diff
@@ -385,144 +383,77 @@ pub async fn ipc_block_listener(
...
+        .get_block_template_components(None, Some(priority))
+        .await
+    {
+        Ok(components) => {
+            match create_braidpool_template(
```
zaidmstrr: Same here use the `?` operator. So we can catch all the errors given by `create_braidpool_template`

zaidmstrr (commented):
Thanks for working on this issue. The changes seem good, but I think there was some issue in error propagation. The current function only returns two types of error and handles or looks for some specific type of error (`ErrorKind::ConnectionBroken` is an example in this case) and ignores the rest of the errors that come from the functions `get_block_template_components` and `create_braidpool_template` and fallback to the two default errors. This way its hard to debug when one of the internal functions fails. So I think replacing `match` with `?` is good here.

s373nZ:
@zaidmstrr Thanks for the good feedback. Removed the `match` statements and replaced with error propagation using the `?` operator.

mcelrath (changes requested):
Thanks for your contribution @s373nZ ! Please address or dismiss the Copilot suggestions. This looks good to me. A couple of questions:

1. Why do we need `initial_nonce` at all? It's basically irrelevant for the block template, we can just set it to zero. Can we simplify the interface by removing this parameter or does it cause any problem?
2. Why do we have `MIN_TEMPLATE_SIZE` at all? What are we guarding against with that? All we should really care about is that `create_braidpool_template` is successful and doesn't error. (This is probably a question for @zaidmstrr since he added this logic)

s373nZ:
> Please address or dismiss the Copilot suggestions.

Addressed in the latest push.

> Why do we need initial_nonce at all? It's basically irrelevant for the block template, we can just set it to zero. Can we simplify the interface by removing this parameter or does it cause any problem?

Looks like this is a good catch. `initial_nonce` is only ever set to `0` in practice, the variations being used previously during retries. IMO should be ok to remove.

I'll wait for @zaidmstrr to opine and confirm on both questions before making any additional changes.

zaidmstrr:
> 1. Why do we need `initial_nonce` at all? It's basically irrelevant for the block template, we can just set it to zero. Can we simplify the interface by removing this parameter or does it cause any problem?

We can remove this `nonce` parameter from `get_template` only and we need to pass `0` to `create_braidpool_template` explicitly then. Because internally `create_block_template` uses this nonce to create a valid block, which is a part of `template_creator.rs`

> 2\. Why do we have `MIN_TEMPLATE_SIZE` at all? What are we guarding against with that? All we should really care about is that `create_braidpool_template` is successful and doesn't error. (This is probably a question for @zaidmstrr since he added this logic)

The reason for this is good logging to miners' consoles. If miners receive a very minimal template, it's just logged, and they know about that they received an almost empty block to mine. Otherwise there is no way to find out that the block we received is minimal or not. If you don't like it, we can remove it.

zaidmstrr (commented):
Code review ACK [4925306](https://github.com/braidpool/braidpool/pull/312/commits/49253063cdec2b32a16bff2749f33d46d6fe79f4)

There are some small nits:

The commit named [doc: Update IPC listener documentation block for get_template().](https://github.com/braidpool/braidpool/pull/312/commits/7f3dc24b7507031a8d269293d5d32b37b985b4e2) has the wrong prefix name `doc`, as `doc` is only prefixed with the changes related to documentation. Same for commit [chore: Remove template retry logic in IPC listener.](https://github.com/braidpool/braidpool/pull/312/commits/49253063cdec2b32a16bff2749f33d46d6fe79f4) here `chore` is not a valid prefix defined under [CONTRIBUTING.doc](https://github.com/braidpool/braidpool/blob/dev/CONTRIBUTING.md), please refer to this doc for styling and naming conventions. And I recommend appending both commits into one.

mcelrath:
> The reason for this is good logging to miners' consoles. If miners receive a very minimal template, it's just logged, and they know about that they received an almost empty block to mine. Otherwise there is no way to find out that the block we received is minimal or not. If you don't like it, we can remove it.

I think a better way to do this is to log the number of transactions. Size is a poor proxy for the number of transactions, which is what we really care about.

zaidmstrr:
> I think a better way to do this is to log the number of transactions. Size is a poor proxy for the number of transactions, which is what we really care about.

Yes, that's good. I will create an issue for this.

s373nZ:
- Removed `initial_nonce` parameter from `get_template()` and replaced with a function-local `NONCE` constant initialized to `0`.
- Squashed commits and used prefix `refactor:`.

zaidmstrr (dismissed):
Code review ACK [22d11c3](https://github.com/braidpool/braidpool/pull/312/commits/22d11c34f7f180edfcf3fd0b4d8dade62a0ad39b)
Thanks for addressing suggested nits.

s373nZ:
Addressed the Copilot suggestion to simplify redundant `match` statements by checking `complete_block_bytes.len()` directly and moving it above  the `processed_template` assignments to satisfy the borrow checker.

zaidmstrr (approved):
Code ACK [f72e3b3](https://github.com/braidpool/braidpool/pull/312/commits/f72e3b37630e7a8c5e23b7874f7493eaa13ad5cc)
Just a small copilot suggestion addressed since the last review.

s373nZ:
Rebased against `dev` and signed commit https://github.com/braidpool/braidpool/pull/312/commits/d8ee85d2c60b20be026827351ccdd3a69551b7a8.

mcelrath (approved):
Looks good! Thanks!

PR #320, opened 2025-12-19
Change IPC listener block size heuristic to transaction count

### LINKED ISSUE #314: Change size huristic to number of transactions inside `ipc.rs`
Mentioned by @mcelrath in [PR#312](https://github.com/braidpool/braidpool/pull/312#issuecomment-3641573448), currently we are tracking for blocks (mostly empty) which contain minimal transactions, like 1 or 2, by looking at the block size defined under [ipc.rs](https://github.com/braidpool/braidpool/blob/dev/node/src/ipc.rs) as `MIN_TEMPLATE_SIZE`. We want to change that heuristic from size to number of transactions.

s373nZ:
This PR is a follow up to #312 to change the block size heuristic from `256` bytes to number (`1`) of transactions.
- Introduces a new accessor method named `block_transaction_count()` in `FinalTemplate` which peeks at the first `varint` after the block header to return the number of transactions in the block template.
- Checks the transaction count against `MIN_TRANSACTION_COUNT=2` instead of the old `MIN_TEMPLATE_SIZE`.
- Refactors `FinalTemplate` accessors to use a constant `BLOCK_HEADER_LENGTH` instead of hardcoding `80` everywhere for better readability.

Questions:
- Contributor guidelines read like there should only be one commit per PR. Including the refactor commit along with this PR felt right to me - let me know if you prefer a separate PR for that.
- I didn't see existing unit tests for the `FinalTemplate` accessors or I would have included one for `block_transaction_count()`. If there is somewhere else to look for tests, please LMK. Can also work on adding one if required.

Closes #314

### DIFF
--- a/node/src/ipc.rs
+++ b/node/src/ipc.rs
@@ -401,7 +401,7 @@ async fn get_template(
     block_height: u32,
     network: Network,
 ) -> Result<client::BlockTemplate, Box<dyn std::error::Error>> {
-    const MIN_TEMPLATE_SIZE: usize = 256;
+    const MIN_TRANSACTION_COUNT: u64 = 1;
     const NONCE: u32 = 0;
     let config = CoinbaseConfig::for_network(network);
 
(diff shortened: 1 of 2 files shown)


mcelrath:
MIN_TRANSACTION_COUNT should be 1 (only the coinbase tx). If you could please change that @s373nZ I will review and merge. Looks correct at first glance.

s373nZ:
@mcelrath
- Changed `MIN_TRANSACTION_COUNT` to `1`.
- Rebased against `dev`.

mcelrath (approved):
Looks correct to me!

PR #324, opened 2025-12-25
Optimizing query for fetching beads on the basis of benchmarking

### LINKED ISSUE #322: Optimizing fetch beads
- Bench-marking using `criterion` about the existing query wrt newer ones, possibly using CTEs or `JOINS` to include the results in single query itself avoiding `OFFSET`.
- Using different paging procedures such as `keyset` etc for optimizing.
- Parallelizing bead fetching as in by worker threads fetching separate batches since once fetched they will be sorted in order to maintain the topological order of `extend` being done, but ultimately the bottle-neck will remain the optimization of `extend` functionality or parallelizing both processes of fetching and simultaneously extending to local braid .
- The above can be done using `tokio::mpsc` but it would require a tight coupling between `DB` and `Braid` which is not necessarily the best choice, however this can be done as a part of another sub-issue .

Sansh2356:
- As per issue proposed in #322 .

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -670,7 +670,7 @@ name = "braidpool-cli"
 version = "0.1.0"
 dependencies = [
  "clap",
- "reqwest 0.12.28",
+ "reqwest",
  "serde",
  "serde_json",
  "tokio",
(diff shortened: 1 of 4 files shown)


Sansh2356:
[benchmark comparison](https://github.com/user-attachments/files/24336321/criterion.zip), though this is on the basis of 5000 beads currently, but the upwards of this is also approximately the same, in this offset is not avoided as using the `key-setting` paging will also lead to similar overhead, perhaps we can utilize `CTE` also to reduce the memory overhead of executing each time during offset batching and the other optimization is mentioned in #322 .

mcelrath:
Please fix the conflicts and will merge.

mcelrath (approved):

Issue #298, opened 2025-11-18
RPC enhancements and braidpool-cli


mcelrath:
Create a new rust binary `braidpool-cli` which operates similarly to `bitcoin-cli` and wraps access to our RPC on `http://localhost:6682`. 

We currently have the following RPCs:
- `getbead` gets a specific bead by hash
- `addbead` adds a bead by JSON
- `getcohortcount` gets the count of the number of cohorts
- `getbeadcount` gets the total number of beads in this epoch
- `gettips` gets the current bead tips

# Immediate TODO list:
1. I've seen getbead fail when requesting tips, not sure why and can't reproduce right now.
2. `gettips` and `getbead` return the result as a string that contains a JSON result. Return the JSON result directly without escaped quotes so I can do `braidpool-cli getbead 00000000291f60a3b42fb75466ffbcbb8e3686492b01833e90ea19569c2bd419 | jq` to pretty print it.
3. I don't think `getcohortcount` is correct. On my node it's showing 249 where there's 608 beads right now. Given the way we've been mining, I strongly doubt there are any cohorts with more than one bead. This needs investigation.

# New RPC methods 

See also #200 which is the primary consumer of these RPCs and #246  @its-me-abhishek 

1. `getcohort` which gets a list of bead hashes in the cohort.
2. `getgenesis` which should return the genesis bead hash for this epoch. It must be a bitcoin block and there should be exactly one genesis.
3. `getminerinfo` get a list of connected Stratum miners
4. `getminer` get detailed statistics about a specific miner identified by `connection_id`/`extraonce1`.
5. `getmininginfo` get detailed statistics about beads mined by us, expected payout, etc.
6. `getparents` get the parent hashes of a bead by bead hash
7. `getchildren` get the children hashes of a bead by bead hash
8. `gethwpath` get the list of beads in the highest work path (should take a `limit` argument)
9. `getipcstats` get statistics about the IPC connection: templates and log information printed like `IPC queue statistics failed=0 avg_ms=1 critical=0 high=0 normal=0 low=0`
10. `getbraidinfo` similar to `getblockchaininfo` in bitcoin-cli but with an aggregation of the above braid information
11. `getnodeinfo` should show: libp2p PeerID, payout address, miner pubkey, etc.
12. `getpeerinfo` should show the IP/PeerID/libp2p address of connected peers, and any statistics we can gather. (Note: I really want latency statistics and dug into this a while back, but libp2p+QUIC does not expose them at all)
[...]

PR #333, opened 2026-01-04
rpc:enhancement and braidpool-cli addition

Shivrajsoni:
#298 

## RPC Server and Braidpool-cli implementation 

Adds JSON-RPC server for the Braidpool node with endpoints for querying braid data, mining info, peer management, and Bitcoin RPC proxying.

### Implemented Methods

Most RPC methods are implemented by reading from shared state (braid, peer manager, stratum connections) and returning the requested data. Methods like `getbead`, `getbeadcount`, `gettips`, `getmininginfo`, `getpeerinfo` etc. query the in-memory braid structure and return JSON responses. The `bitcoinproxy` method forwards requests directly to Bitcoin Core via HTTP JSON-RPC.

### Methods Not Implemented in this PR 
`getminer` get detailed statistics about a specific miner identified by connection_id/extraonce1.
`committedtransactions` Transactions committed in beads but not yet mined (might use cmempool IPC) [See this](https://x.com/braidpool/status/1929678658010337392)
`blockstagedtransactions` Transactions in the blocktemplate we're currently mining (might use cmempool IPC) [See this](https://x.com/braidpool/status/1929678658010337392)
`stagetransaction` Add a transaction to our staged list. Can take a txid (from bitcoind) or a full tx.
`cmempool:<method>` proxy IPC calls to cmempoold using @zaidmstrr's prototype https://github.com/zaidmstrr/bitcoin-rpc-multiprocess
 Note that cmempool is fast-changing (we have to add and remove txs every time we see a new bead to create block  templates), so this may be of limited usefulness.

### DIFF
--- a/braidpool-cli/src/main.rs
+++ b/braidpool-cli/src/main.rs
@@ -1,0 +1,197 @@
+use clap::{Parser, Subcommand};
+use serde::{Deserialize, Serialize};
+use serde_json::json;
+
+/// Braidpool CLI - Command line interface for interacting with Braidpool node
+#[derive(Parser, Debug)]
+#[command(name = "braidpool-cli", version, about, long_about = None)]
+struct Cli {
+    /// RPC server URL (default: http://127.0.0.1:6682)
+    #[arg(long, default_value = "http://127.0.0.1:6682")]
+    rpc_url: String,
+
+    #[command(subcommand)]
+    commands: Commands,
+}
+
+#[derive(Debug, Subcommand)]
+#[command(rename_all = "PascalCase")]
+enum Commands {
+    /// Get a bead by hash
+    GetBead {
+        /// The bead hash (as a hex string)
+        bead_hash: String,
+    },
+
+    /// Add a bead via serialized JSON string
+    AddBead {
+        /// JSON-formatted bead
+        bead_data: String,
+    },
+
+    /// Get total number of beads
+    GetBeadCount,
+
+    /// Get total number of cohorts
+    GetCohortCount,
+
+    /// Get current DAG tips
+    GetTips,
+
+    /// Get a list of bead hashes in the cohort
+    GetCohort {
+        /// The id of the Cohort
+        cohort_id: u64,
+    },
+
+    /// Get the genesis bead hash for this epoch
+    GetGenesis,
+
+    /// Get a list of connected Stratum miners
+    GetMinerInfo,
+
+    ///  get detailed statistics about beads mined by us, expected payout, etc.
+    GetMiningInfo,
+
+    /// Get the parent hashes of a bead by bead_hash
+    GetParents {
+        /// The bead hash (as a hex string)
+        bead_hash: String,
+    },
+
+    /// Get the children hashes of a bead by bead hash
+    GetChildren {
+        /// The bead hash (as a hex string)
+        bead_hash: String,
+    },
+
+    /// Get the list of beads in the highest work path
+    GetHwPath {
+        /// Limit the number of results
+        limit: u8,
+    },
+
+    /// Get statistics about the IPC connection
+    GetIpcStats,
+
+    /// Get braid information (similar to getblockchaininfo in bitcoin-cli)
+    GetBraidInfo,
+
+    /// Get node information (libp2p PeerID, payout address, miner pubkey, etc.)
+    GetNodeInfo {
+        /// The node identifier (bead hash)
+        node: String,
+    },
+
+    /// Get peer information (IP/PeerID/libp2p address of connected peers)
+    GetPeerInfo,
+
+    /// Get the list of transactions staged for the next bead we mine
+    StagedTransactions,
+
+    /// Remove a transaction from our stage list by txid
+    UnstageTransactions {
+        /// Transaction ID to remove
+        tx_id: String,
+    },
+
+    /// Proxy a Bitcoin RPC call to bitcoind
+    /// Example: braidpool-cli bitcoin getblockchaininfo
+    Bitcoin {
+        /// Bitcoin RPC method name (e.g., "getblockchaininfo", "getblockhash", etc.)
+        method: String,
+        /// JSON array of parameters (optional, defaults to empty array)
+        #[arg(long, default_value = "[]")]
+        params: String,
+    },
+}
+
+#[derive(Serialize, Debug)]
+struct JsonRpcRequest {
+    jsonrpc: &'static str,
+    method: String,
+    params: serde_json::Value,
+    id: u64,
+}
+
+#[derive(Deserialize, Debug)]
+struct JsonRpcResponse {
+    result: serde_json::Value,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    error: Option<serde_json::Value>,
+}
+
+/// Client-side RPC call function
+async fn call_rpc(
+    rpc_url: &str,
+    method: &str,
+    params: serde_json::Value,
+) -> Result<serde_json::Value, Box<dyn std::error::Error>> {
+    let rpc_request = JsonRpcRequest {
+        jsonrpc: "2.0",
+        method: method.to_string(),
+        params,
+        id: 1,
+    };
+
+    let client = reqwest::Client::new();
+    let res = client.post(rpc_url).json(&rpc_request).send().await?;
+
+    if res.status().is_success() {
+        let rpc_response: JsonRpcResponse = res.json().await?;
+        if let Some(error) = rpc_response.error {
+            return Err(format!("RPC error: {}", error).into());
+        }
+        Ok(rpc_response.result)
+    } else {
+        let status = res.status();
+        let text = res
+            .text()
+            .await
+            .unwrap_or_else(|_| "Could not read error body".to_string());
+        Err(format!("HTTP error: {}\nResponse: {}", status, text).into())
+    }
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let cli = Cli::parse();
+
+    let (method, params) = match &cli.commands {
+        Commands::GetBead { bead_hash } => ("getbead", json!([bead_hash])),
+        Commands::AddBead { bead_data } => ("addbead", json!([bead_data])),
+        Commands::GetBeadCount => ("getbeadcount", json!([])),
+        Commands::GetCohortCount => ("getcohortcount", json!([])),
+        Commands::GetTips => ("gettips", json!([])),
+        Commands::GetCohort { cohort_id } => ("getcohort", json!([cohort_id])),
+        Commands::GetGenesis => ("getgenesis", json!([])),
+        Commands::GetMinerInfo => ("getminerinfo", json!([])),
+        Commands::GetMiningInfo => ("getmininginfo", json!([])),
+        Commands::GetParents { bead_hash } => ("getparents", json!([bead_hash])),
+        Commands::GetChildren { bead_hash } => ("getchildren", json!([bead_hash])),
+        Commands::GetHwPath { limit } => ("gethwpath", json!([limit])),
+        Commands::GetIpcStats => ("getipcstats", json!([])),
+        Commands::GetBraidInfo => ("getbraidinfo", json!([])),
+        Commands::GetNodeInfo { node } => ("getnodeinfo", json!([node])),
+        Commands::GetPeerInfo => ("getpeerinfo", json!([])),
+        Commands::StagedTransactions => ("stagedtransactions", json!([])),
+        Commands::UnstageTransactions { tx_id } => ("unstagetransactions", json!([tx_id])),
+        Commands::Bitcoin { method, params } => {
+            let params_value: serde_json::Value =
+                serde_json::from_str(params).unwrap_or_else(|_| json!([]));
+            ("bitcoinproxy", json!([method, params_value]))
+        }
+    };
+
+    match call_rpc(&cli.rpc_url, method, params).await {
+        Ok(result) => {
+            let pretty_response = serde_json::to_string_pretty(&result)?;
+            println!("{}", pretty_response);
+            Ok(())
+        }
+        Err(e) => {
+            eprintln!("Error: {}", e);
+            std::process::exit(1);
+        }
+    }
+}
--- a/braidpool-cli/src/main.rs
+++ b/braidpool-cli/src/main.rs
@@ -15,43 +15,51 @@ struct Cli {
 }
 
 #[derive(Debug, Subcommand)]
-#[command(rename_all = "snakecase")]
 enum Commands {
     /// Get a bead by hash
+    #[command(name = "getbead")]
     GetBead {
         /// The bead hash (as a hex string)
         bead_hash: String,
     },
 
     /// Add a bead via serialized JSON string
+    #[command(name = "addbead")]
     AddBead {
         /// JSON-formatted bead
         bead_data: String,
     },
 
     /// Get total number of beads
+    #[command(name = "getbeadcount")]
     GetBeadCount,
 
     /// Get total number of cohorts
+    #[command(name = "getcohortcount")]
     GetCohortCount,
 
     /// Get current DAG tips
+    #[command(name = "gettips")]
     GetTips,
 
     /// Get a list of bead hashes in a cohort by its ID
+    #[command(name = "getcohortbyid")]
     GetCohortById {
         /// The ID of the cohort
         cohort_id: u64,
     },
 
     /// Get the genesis bead hash for this epoch
+    #[command(name = "getgenesis")]
     GetGenesis,
 
     /// Get a list of connected Stratum miners
+    #[command(name = "getminerinfo")]
     GetMinerInfo,
 
     /// Get detailed statistics about beads mined by us, expected payout, etc.
     /// Requires at least one filter: public_keys or miner_ips
+    #[command(name = "getmininginfo")]
     GetMiningInfo {
         /// List of public keys (hex-encoded) to filter beads by.
         /// Supports multiple keys for key rotation scenarios.
@@ -67,32 +75,38 @@ enum Commands {
     },
 
     /// Get the parent hashes of a bead by bead_hash
+    #[command(name = "getparents")]
     GetParents {
         /// The bead hash (as a hex string)
         bead_hash: String,
     },
 
     /// Get the children hashes of a bead by bead hash
+    #[command(name = "getchildren")]
     GetChildren {
         /// The bead hash (as a hex string)
         bead_hash: String,
     },
 
     /// Get the list of beads in the highest work path, limited by count
+    #[command(name = "gethighestworkpathbycount")]
     GetHighestWorkPathByCount {
         /// Limit the number of results returned
         limit: u8,
     },
 
     /// Get statistics about the IPC connection
+    #[command(name = "getipcstats")]
     GetIpcStats,
 
     /// Get braid information (returns bead_count , tip_count , tips, cohort_count, orphan_count, genesis_beads, total_work)
+    #[command(name = "getbraidinfo")]
     GetBraidInfo,
 
     /// Get node information for the node that created a specific bead
     /// Returns information about the node that created the specified bead, including:
     /// common_pubkey (libp2p public key), miner_ip, payout_address, and minimum_target.
+    #[command(name = "getnodeinfo")]
     GetNodeInfo {
         /// The bead hash (block hash) as a 64-character hex-encoded string representing the bead's block hash
         /// This identifies which bead's creator node information you want to retrieve
@@ -100,12 +114,15 @@ enum Commands {
     },
 
     /// Get peer information (IP/PeerID/libp2p address of connected peers)
+    #[command(name = "getpeerinfo")]
     GetPeerInfo,
 
     /// Get the list of transactions staged for the next bead we mine
+    #[command(name = "stagedtransactions")]
     StagedTransactions,
 
     /// Remove a transaction from our stage list by txid
+    #[command(name = "unstagetransactions")]
     UnstageTransactions {
         /// Transaction ID to remove
         tx_id: String,
@@ -113,6 +130,7 @@ enum Commands {
 
     /// Proxy a Bitcoin RPC call to bitcoind
     /// Example: braidpool-cli bitcoin getblockchaininfo
+    #[command(name = "bitcoin")]
     Bitcoin {
         /// Bitcoin RPC method name (e.g., "getblockchaininfo", "getblockhash", etc.)
         method: String,
@@ -199,6 +217,7 @@ impl std::error::Error for RpcCallError {
 }
 
 async fn call_rpc(
+    client: &reqwest::Client,
     rpc_url: &str,
     method: &str,
     params: serde_json::Value,
@@ -210,7 +229,6 @@ async fn call_rpc(
         id: 1,
     };
 
-    let client = reqwest::Client::new();
     let res = client
         .post(rpc_url)
         .json(&rpc_request)
@@ -245,6 +263,10 @@ async fn call_rpc(
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let cli = Cli::parse();
+
+    let client = reqwest::Client::builder()
+        .timeout(std::time::Duration::from_secs(30))
+        .build()?;
 
     let (method, params) = match &cli.commands {
         Commands::GetBead { bead_hash } => ("getbead", json!([bead_hash])),
@@ -305,7 +327,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         }
     };
 
-    match call_rpc(&cli.rpc_url, method, params).await {
+    match call_rpc(&client, &cli.rpc_url, method, params).await {
         Ok(result) => {
             let pretty_response = serde_json::to_string_pretty(&result).map_err(|e| {
                 Box::new(std::io::Error::new(
--- a/braidpool-cli/src/main.rs
+++ b/braidpool-cli/src/main.rs
@@ -15,7 +15,7 @@ struct Cli {
 }
 
 #[derive(Debug, Subcommand)]
-#[command(rename_all = "PascalCase")]
+#[command(rename_all = "snakecase")]
 enum Commands {
     /// Get a bead by hash
     GetBead {
@@ -38,9 +38,9 @@ enum Commands {
     /// Get current DAG tips
     GetTips,
 
-    /// Get a list of bead hashes in the cohort
-    GetCohort {
-        /// The id of the Cohort
+    /// Get a list of bead hashes in a cohort by its ID
+    GetCohortById {
+        /// The ID of the cohort
         cohort_id: u64,
     },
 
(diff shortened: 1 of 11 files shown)


Sansh2356:
@Shivrajsoni kindly do `rustfmt` and test suite is failing for `rpc_server.rs` test make sure since tests run in parallel and not in sequential manner so port must be kept distinct for each server spawning, also the `todo!` implementations can be removed entirely and can be added in next sub-PRs and update PR description also .

Shivrajsoni:
hi @Sansh2356  Thanks for the review!
I’ve applied rustfmt, fixed the failing rpc_server.rs tests by assigning distinct ports to each spawned server (keeping parallel test execution in mind), and removed the todo! implementations as suggested to be handled in follow-up sub-PRs.

The PR description has also been updated accordingly. Please let me know if anything else needs adjustment.

Sansh2356:
ut-ACK have gone through code looks fine, will test it asap.

braidpool-cli/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+    #[arg(long, default_value = "http://127.0.0.1:6682")]
+    rpc_url: String,
+
+    #[command(subcommand)]
+    commands: Commands,
+}
+
+#[derive(Debug, Subcommand)]
+#[command(rename_all = "PascalCase")]
+enum Commands {
```
zaidmstrr, line 18: The RPC names here are in `PascalCase` which is generally not good to have as a JSON-RPC method name. Change these to simple `lower_snake_case`.

node/src/rpc_server.rs
```diff
@@ -240,6 +290,493 @@ impl RpcServer for RpcServerImpl {
...
+        serde_json::to_value(&braid_info)
+            .map_err(|_| ErrorObjectOwned::owned(2, "Internal Server Error", None::<()>))
+    }
+
+    async fn get_node_info(&self, node: String) -> Result<Value, ErrorObjectOwned> {
+        info!("Get Node Info request received");
+
+        let hash = node
+            .parse::<BeadHash>()
+            .map_err(|_| ErrorObjectOwned::owned(1, "Invalid bead hash format", None::<()>))?;
```
zaidmstrr: This error is not propagating to the end user CLI upon passing invalid hex, instead of this the CLI is returning `Error: error decoding response body` as a common message to all other invalid arguments. This behaviour is  same in other RPC methods also.

node/src/rpc_server.rs
```diff
@@ -137,36 +44,188 @@ pub async fn handle_request(
...
         }
     }
 }
 #[async_trait]
 impl RpcServer for RpcServerImpl {
-    async fn get_bead(&self, bead_hash: String) -> Result<String, ErrorObjectOwned> {
+    async fn get_bead(&self, bead_hash: String) -> Result<Bead, ErrorObjectOwned> {
         let hash = bead_hash
             .parse::<BeadHash>()
             .map_err(|_| ErrorObjectOwned::owned(1, "Invalid bead hash format", None::<()>))?;
```
zaidmstrr, line 280: Same here this is not propagating to the end user CLI upon passing invalid args.

braidpool-cli/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let cli = Cli::parse();
+
+    let (method, params) = match &cli.commands {
+        Commands::GetBead { bead_hash } => ("getbead", json!([bead_hash])),
+        Commands::AddBead { bead_data } => ("addbead", json!([bead_data])),
+        Commands::GetBeadCount => ("getbeadcount", json!([])),
+        Commands::GetCohortCount => ("getcohortcount", json!([])),
```
zaidmstrr, line 275: This command should return `int` instead of `String` and this is not related to your change but I think you can change this behaviour in `rpc_server.rs::get_cohort_count`

braidpool-cli/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+        Commands::GetTips => ("gettips", json!([])),
+        Commands::GetCohort { cohort_id } => ("getcohort", json!([cohort_id])),
```
zaidmstrr: The name `GetCohort` is confusing this should be `getcohortbyid` or similar.

node/src/rpc_server.rs
```diff
@@ -240,6 +290,493 @@ impl RpcServer for RpcServerImpl {
...
+        let miner_ips: Vec<String> = connection_map
+            .downstream_channel_mapping
+            .keys()
+            .cloned()
+            .collect();
+
+        Ok(miner_ips)
+    }
+
+    async fn get_mining_info(&self) -> Result<Value, ErrorObjectOwned> {
```
zaidmstrr: Currently `GetMiningInfo` is showing some stats about the work, and I'm unsure about how beads mined by me and total work counted by me are calculated. The current heuristic it follows is checking with `comm_pub_key`, but do you think it's enough to get this info right? What if miners use different `pub_keys` or change in the middle? Then how will this give us the correct output?

braidpool-cli/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+        Commands::GetGenesis => ("getgenesis", json!([])),
+        Commands::GetMinerInfo => ("getminerinfo", json!([])),
+        Commands::GetMiningInfo => ("getmininginfo", json!([])),
+        Commands::GetParents { bead_hash } => ("getparents", json!([bead_hash])),
+        Commands::GetChildren { bead_hash } => ("getchildren", json!([bead_hash])),
+        Commands::GetHwPath { limit } => ("gethwpath", json!([limit])),
```
zaidmstrr: This name doesn't seem clear `GetHwPath`. Try to make it more clear or use full spelling, consider using `GetHighestWorkPathByCount` or similar.

node/src/rpc_server.rs
```diff
@@ -240,6 +290,493 @@ impl RpcServer for RpcServerImpl {
...
+                        .block_hash()
+                        .to_string()
+                })
+                .collect(),
+            None => Vec::new(), // This case is unlikely if the parent exists, but it's safe to handle.
+        };
+        Ok(children_hashes)
+    }
+
+    async fn get_hwpath(&self, limit: u8) -> Result<String, ErrorObjectOwned> {
```
zaidmstrr: Currently typing the value greater than the available beads doesn't show some error or message to the user; it just shows all the hashes. Consider showing some kind of message to the user, like `"The max available count is x"` or `"The provided count seems too high and does not exist"` plus add some more info for the user so they can figure out what to do next.

node/src/rpc_server.rs
```diff
@@ -240,6 +290,493 @@ impl RpcServer for RpcServerImpl {
...
+            )),
+            Err(_) => Err(ErrorObjectOwned::owned(
+                5,
+                "IPC proxy handler channel closed - response receiver dropped",
+                None::<()>,
+            )),
+        }
+    }
+
+    async fn get_braid_info(&self) -> Result<Value, ErrorObjectOwned> {
```
zaidmstrr, line 802: `GetBraidInfo` showing me `orphan_count` `5920`, which doesn't seem a true value here, but maybe this bug is not related to this code and part of the other file. But please debug and try to find out.

braidpool-cli/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+    },
+
+    /// Get statistics about the IPC connection
+    GetIpcStats,
+
+    /// Get braid information (similar to getblockchaininfo in bitcoin-cli)
+    GetBraidInfo,
+
+    /// Get node information (libp2p PeerID, payout address, miner pubkey, etc.)
+    GetNodeInfo {
```
zaidmstrr, line 110: `GetNodeInfo` is very unclear in passing arguments; it doesn't state what I need to pass along with this method. Please clarify clearly to user what they need to pass with this RPC method.

braidpool-cli/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// Get the list of beads in the highest work path
+    GetHwPath {
+        /// Limit the number of results
+        limit: u8,
+    },
+
+    /// Get statistics about the IPC connection
+    GetIpcStats,
+
+    /// Get braid information (similar to getblockchaininfo in bitcoin-cli)
```
zaidmstrr: Remove `similar to getblockchaininfo in bitcoin-cli` from here it's unnecessary. You can replace this with some other information that this RPC can return.

zaidmstrr (changes requested):
Thanks for working on this. I reviewed this and found some noticeable points to look at. You can review, implement and further discuss the points below. I give the opinions based on the high-level overview of the PR. More technical go-through might be discussed later on review follow ups.

node/src/rpc_server.rs
Sansh2356: This is related to #299 and #309 should be solved after these patches are merged .

braidpool-cli/src/main.rs
Sansh2356: Agreed this should be changed .

braidpool-cli/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+struct JsonRpcRequest {
+    jsonrpc: &'static str,
+    method: String,
+    params: serde_json::Value,
+    id: u64,
+}
+
+#[derive(Deserialize, Debug)]
+struct JsonRpcResponse {
+    result: serde_json::Value,
```
Sansh2356: Here since you are using `reqwest` it will wrap the RPC response into Http response, either you can directly parse the `res.text` that is the http body that will be containing the rpc response or you can make the parsing of result also optional here and then propagate error to the cli user, currently no error is being propagated to the user as mentioned by @zaidmstrr  also you can choose from above or any other way you might like .

braidpool-cli/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    error: Option<serde_json::Value>,
```
Sansh2356: Instead of  `Value` can directly make a `JsonRpcError` type here referring directly as per [Json-rpc 2.0](https://www.jsonrpc.org/specification#error_object) here 
making it more verbose if you want to .

braidpool-cli/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+        jsonrpc: "2.0",
+        method: method.to_string(),
+        params,
+        id: 1,
+    };
+
+    let client = reqwest::Client::new();
+    let res = client.post(rpc_url).json(&rpc_request).send().await?;
+
+    if res.status().is_success() {
```
Sansh2356: This condition is always true so no check is required since [`jsonrpsee`](https://docs.rs/jsonrpsee-core/0.26.0/jsonrpsee_core/middleware/trait.RpcServiceT.html) server side responses are `Infallible` hence even if there is an error at rpc side it will still be 200 status code at http client .

node/src/rpc_server.rs
```diff
@@ -137,36 +44,188 @@ pub async fn handle_request(
...
+    async fn get_hwpath(&self, limit: u8) -> Result<String, ErrorObjectOwned>;
+
+    #[method(name = "getipcstats")]
+    async fn get_ipc_stats(&self) -> Result<String, ErrorObjectOwned>;
+
+    #[method(name = "getbraidinfo")]
+    async fn get_braid_info(&self) -> Result<Value, ErrorObjectOwned>;
+
+    #[method(name = "getnodeinfo")]
+    async fn get_node_info(&self, node: String) -> Result<Value, ErrorObjectOwned>;
```
Sansh2356: This method as it is currently implemented doesn't make any sense to me as also mentioned by @zaidmstrr .

node/src/rpc_server.rs
```diff
@@ -137,36 +44,188 @@ pub async fn handle_request(
...
+    method: String,
+    params: serde_json::Value,
+    id: u64,
+}
+
+/// JSON-RPC response structure
+#[derive(Deserialize)]
+struct JsonRpcResponse {
+    result: Option<serde_json::Value>,
+    error: Option<serde_json::Value>,
```
Sansh2356, line 174: Same as above `JsonRpcError` instead of `Value` .

Sansh2356 (changes requested):
Small nits required kindly check @Shivrajsoni .

Shivrajsoni:
hi @Sansh2356 @zaidmstrr Thanks for the review!
I have applied the requested changes

node/src/rpc_server.rs
Shivrajsoni: To avoid ambiguity with comm_pub_key, the updated approach for GetMiningInfo allows the client to explicitly provide either the miner’s public key(s) or miner IP(s).

Work and bead statistics are then calculated by filtering contributions based on these user-supplied identifiers rather than relying on a single inferred key. This makes the accounting deterministic even if a miner rotates public keys or switches identities mid-session.

braidpool-cli/src/main.rs
Shivrajsoni: Updated the RPC method names to use lower_snake_case instead of PascalCase

node/src/rpc_server.rs
Shivrajsoni: Updated to handle this edge case.

When the requested count exceeds the available beads, the RPC now returns a clear error, and the CLI surfaces a user-friendly message (e.g., indicating the maximum available count and why the request failed). This helps users understand what went wrong and how to adjust their input.

braidpool-cli/src/main.rs
Shivrajsoni: Added documentation and clearer argument naming, specifically bead_hash, with inline comments explaining that it identifies which bead’s creator node information is being requested. This should make the expected input and usage explicit to users.

braidpool-cli/src/main.rs
Shivrajsoni: Updated this to use a dedicated JsonRpcError type instead of a generic serde_json::Value, aligning with the JSON-RPC 2.0 specification. This makes the error structure more explicit and easier to extend

braidpool-cli/src/main.rs
Shivrajsoni: Updated this by making result optional and properly propagating RPC errors through the HTTP response to the CLI.

braidpool-cli/src/main.rs
Shivrajsoni: updated

node/src/rpc_server.rs
Shivrajsoni: Fixed by properly propagating JSON-RPC errors to the CLI using a structured RpcCallError enum, avoiding generic decode errors for cases like invalid hex across all RPC methods.

Sansh2356 (dismissed):
t-ACK @Shivrajsoni  you can checkout the co-pilot review if you find any useful otherwise looks good to me .

braidpool-cli/src/main.rs
```diff
@@ -0,0 +1,338 @@
...
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let cli = Cli::parse();
+
+    let (method, params) = match &cli.commands {
+        Commands::GetBead { bead_hash } => ("getbead", json!([bead_hash])),
+        Commands::AddBead { bead_data } => ("addbead", json!([bead_data])),
+        Commands::GetBeadCount => ("getbeadcount", json!([])),
```
zaidmstrr, line 274: It will be good if this `Commands::GetBeadCount` return an int instead of String, just do the same thing you did with `Commands::GetCohortCount`

node/src/rpc_server.rs
```diff
@@ -233,12 +314,640 @@ impl RpcServer for RpcServerImpl {
...
+                        .block_hash()
+                        .to_string()
+                })
+                .collect(),
+            None => Vec::new(), // This case is unlikely if the parent exists, but it's safe to handle.
+        };
+        Ok(children_hashes)
+    }
+
+    async fn get_highest_work_path_by_count(&self, limit: u8) -> Result<String, ErrorObjectOwned> {
```
zaidmstrr: There are some inconsistent Stringly Typed return types which are creating some structural problems while delivering the response to the user. Currently a client needs to manually double-decode the JSON string inside the JSON response as a result, it shows the output like this: 

```
./target/debug/braidpool-cli get_highest_work_path_by_count 10
"[\"000000000b3ad6bf987cf52981945c72a3c9db445d224d3eb87449ed700262d0\",\"00000000035ee15c78a7a2a588f5ce3df1ed081d83811831d934643b602c2555\",\"000000004d6ef66c16eaa64a297d3dbd91bf21574bb60919f8076bb42c44e913\",\"00000000d4dd3010388ea45036a3a16aaf3831136e918716e469a5d7b31b5d2c\",\"00000000c7368ce0bd30f56dc296750c451e433606eb74d2a1708c9f30130e73\",\"00000000405833a00c55acd14e60009e8a23e698b713abc034395e152a8df293\",\"000000006cb457f27806c757f449aa7d24c2f4d7411718b4c7fa89fc5ca08a9a\",\"00000000f647585406045342e7c23789a7d7f962676033f7441d7b4631ba659b\",\"000000003459c46e416cd4d87adeff3c20de1ce4b92cfc393f48057d9ddd93e8\",\"000000000080b772d9a49dc7f404ea0aba5aae1a99422fc60b2f53f649526cf2\"]" 
```
while the desired way of showing is like this:
```
./target/debug/braidpool-cli get_highest_work_path_by_count 10
[
  "000000000b3ad6bf987cf52981945c72a3c9db445d224d3eb87449ed700262d0",
  "00000000035ee15c78a7a2a588f5ce3df1ed081d83811831d934643b602c2555",
  "000000004d6ef66c16eaa64a297d3dbd91bf21574bb60919f8076bb42c44e913",
  "00000000d4dd3010388ea45036a3a16aaf3831136e918716e469a5d7b31b5d2c",
  "00000000c7368ce0bd30f56dc296750c451e433606eb74d2a1708c9f30130e73",
  "00000000405833a00c55acd14e60009e8a23e698b713abc034395e152a8df293",
  "000000006cb457f27806c757f449aa7d24c2f4d7411718b4c7fa89fc5ca08a9a",
  "00000000f647585406045342e7c23789a7d7f962676033f7441d7b4631ba659b",
  "000000003459c46e416cd4d87adeff3c20de1ce4b92cfc393f48057d9ddd93e8",
  "000000000080b772d9a49dc7f404ea0aba5aae1a99422fc60b2f53f649526cf2"
]
```

node/src/rpc_server.rs
```diff
@@ -233,12 +314,640 @@ impl RpcServer for RpcServerImpl {
...
+    async fn get_cohort_count(&self) -> Result<u64, ErrorObjectOwned> {
         let braid_data = self.braid_arc.read().await;
         let count = braid_data.cohorts.len();
         info!(count = %count, "Get cohort count request received");
 
-        Ok(count.to_string())
+        Ok(count as u64)
+    }
+
+    async fn get_cohort_by_id(&self, cohort_id: u64) -> Result<String, ErrorObjectOwned> {
```
zaidmstrr: This function also has the inconsistent Stringly Typed return types problem. The current behaviour is like this:
```
./target/debug/braidpool-cli get_cohort_by_id 10
"[\"000000002ae201f1836157be9d9334d57f9264bfde07ee4cb5a27ae9564153f2\"]"
```

node/src/rpc_server.rs
```diff
@@ -233,12 +314,640 @@ impl RpcServer for RpcServerImpl {
...
+            our_work_share_percent,
+            payout_address,
+            filter_info,
+        };
+
+        serde_json::to_value(&mining_info)
+            .map_err(|e| ErrorObjectOwned::owned(2, format!("Internal Error: {}", e), None::<()>))
+    }
+
+    async fn get_parents(&self, bead_hash: String) -> Result<String, ErrorObjectOwned> {
```
zaidmstrr: This function also has the problem of inconsistent Stringly Typed return types. The current behaviour is like this:
``` 
 ./target/debug/braidpool-cli get_parents 000000003459c46e416cd4d8
7adeff3c20de1ce4b92cfc393f48057d9ddd93e8
"[\"00000000f647585406045342e7c23789a7d7f962676033f7441d7b4631ba659b\"]"
```

braidpool-cli/src/main.rs
zaidmstrr: You can also remove the underscore from the names. For example `getparents` `getmininginfo` etc. As these are easier to type.

node/src/rpc_server.rs
```diff
@@ -317,17 +1038,97 @@ pub async fn run_rpc_server(
...
+    let url = config.rpc_url();
+    let request = JsonRpcRequest {
+        jsonrpc: "2.0",
+        method: method.to_string(),
+        params: params.clone(),
+        id: 1,
+    };
+
+    // Build HTTP client with timeout
+    let client = reqwest::Client::builder()
```
zaidmstrr: In `bitcoin_proxy` and `call_bitcoin_rpc_direct`, you are creating a new HTTP client for every single request which is not optimal, try to create a client a single time and reuse that for the rest of the session.

zaidmstrr (changes requested):
Thanks for implementing and fixing the previously marked issues. Most of the code structure seems solid but there are some more potential issues I found related to performance, structure and user interface.

braidpool-cli/src/main.rs
Shivrajsoni: fixed it , GetBeadCount and GetCohortCount returns int now instead of string

node/src/rpc_server.rs
Shivrajsoni: fixed the inconsistent stringly typed return types by updating the CLI to stop doing the extra decode and to pretty-print the response. That should give the formatted list you described.

node/src/rpc_server.rs
Shivrajsoni: created a bitcoin_proxy http client once at service/session started  ,stored it in BitcoinRpcConfig , and reused it , also done same for braidpool_cli http_client , also added timeout of 30 sec

node/src/rpc_server.rs
Shivrajsoni: fixed

braidpool-cli/src/main.rs
Shivrajsoni: removed it

zaidmstrr: Some of the RPC method names still contain an underscore with them.

zaidmstrr (commented):
Thanks for fixing most of the suggested things. But I saw a change which is partially implemented, you can take a look and fix that.

Shivrajsoni:
thnx for the review @zaidmstrr , i have implemented the changes

node/src/peer_manager/mod.rs
```diff
@@ -258,6 +258,65 @@ impl PeerManager {
...
+            self.peers.values().filter(|p| p.connected).collect();
+        let total_peers = self.peers.len();
+        let connected = connected_peers_info.len();
+
+        let mut avg_latency_ms = 0.0;
+        let mut latency_count = 0;
+        let mut geo_groups = HashSet::new();
+        let mut inbound_count = 0;
+
+        for info in &connected_peers_info {
```
zaidmstrr: You're iterating through connected_peers_info twice first to collect the stats and then to build the JSON list.

node/src/peer_manager/mod.rs
```diff
@@ -258,6 +258,65 @@ impl PeerManager {
...
+            if let Some(latency) = info.latency {
+                avg_latency_ms += latency.as_millis() as f64;
+                latency_count += 1;
+            }
+            if let Some(group) = &info.geo_group {
+                geo_groups.insert(group.clone());
+            }
+        }
+
+        if latency_count > 0 {
```
zaidmstrr, line 296: This is redundant and used here and below inside json `avg_latency_ms`.

node/src/peer_manager/mod.rs
```diff
@@ -258,6 +258,65 @@ impl PeerManager {
         self.connected_peers.len()
     }
 
+    /// Get peer information as JSON value for RPC responses
+    pub fn get_peers_json(&self) -> serde_json::Value {
+        use serde_json::json;
+
+        let connected_peers_info: Vec<&PeerInfo> =
```
zaidmstrr: Instead of allotting this here can you use the direct `self.peers.values()`?

zaidmstrr (commented):
I found few performance optimizations. It will be good if you look at them.

node/src/peer_manager/mod.rs
Shivrajsoni: I've refactored get_peers_json to make a single pass over self.peers.values().filter(|p| p.connected). In this single loop, both the aggregate statistics (inbound count, total latency for average, geo groups) and the individual peer JSON objects are collected simultaneously.

node/src/peer_manager/mod.rs
Shivrajsoni: removed the logic from json

node/src/main.rs
```diff
@@ -238,25 +238,15 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-    // Initializing the peer manager
-    let mut peer_manager = PeerManager::new(8);
+    // Initializing the peer manager (shared between swarm and RPC server)
+    // Using RwLock to allow concurrent reads (RPC server) while swarm handler can write
+    let peer_manager_arc = Arc::new(tokio::sync::RwLock::new(PeerManager::new(8)));
     //For local testing uncomment this keypair peer since it running to process will
     //result in same peerID leading to OutgoingConnectionError
 
     // let keypair = identity::Keypair::generate_ed25519();
+
```
zaidmstrr: nit: redundant newline.

node/src/main.rs
```diff
@@ -330,6 +320,29 @@ async fn main() -> Result<(), Box<dyn Error>> {
     let latest_template_for_ipc = latest_template.clone();
     let latest_template_merkle_branch_for_ipc = latest_template_merkle_branch.clone();
 
+    // Create RPC proxy command channel - sender goes to RPC server, receiver goes to IPC handler
+    let (rpc_proxy_tx, rpc_proxy_rx) = tokio::sync::mpsc::unbounded_channel::<RpcProxyCommand>();
+
+    // peer_manager_arc is created above and shared between swarm and RPC server
+
```
zaidmstrr: nit: redundant newline.

zaidmstrr (commented):
Overall looks good, but there are some small nits. Also please resolve the conflict and push again.

node/src/main.rs
Shivrajsoni: Thanks for the feedback! I’ve resolved the conflict and addressed the nits. Please take another look.

node/src/rpc_server.rs
```diff
@@ -222,23 +322,669 @@ impl RpcServer for RpcServerImpl {
...
+        }
+        let genesis_bead_index = braid_data.genesis_beads.iter().next().unwrap();
+        let genesis_bead = &braid_data.beads[*genesis_bead_index];
+
+        Ok(genesis_bead.block_header.block_hash().to_string())
+    }
+
+    async fn get_miner_info(&self) -> Result<Vec<String>, ErrorObjectOwned> {
+        info!("Get Miner Info Request Received");
+        let connection_map = self.stratum_connection_mapping.lock().await;
```
zaidmstrr, line 390: Here this is locking the main stratum connection mapping on each RPC call; this can block the critical tasks when someone uses this RPC and stratum tries to update the state on connection mapping, for e.g. adding the new miner, removing the disconnected miner or sending messages to miners. This can be more impactful when someone uses the `getminerinfo` RPC aggressively. The solution which might be here is to wrap the `connection_mapping` with `RwLock`. And I guess this change is out of scope for this PR, so I'm proceeding with this, but this can be a potential future optimization. I think using this lock on `connection_mapping` is also useful throughout the codebase in many places.

zaidmstrr (approved):
Code ACK [29f04fe](https://github.com/braidpool/braidpool/pull/333/changes/29f04fed44ea873be5d164a56e14cff8d35f30e1)

Thanks for considering the previous suggestions. This looks good to me, although I have suggested one future optimization which we can do, but that is not related to this PR. So you or anyone else can take a look at that.

node/src/rpc_server.rs
Shivrajsoni: Thanks for the review @zaidmstrr ,  I’ll open a follow-up issue and work on it separately

PR #371, opened 2026-02-04
fix(DAA): Implement corrected Braid Consensus DAA per spec

Harshit2405-2004:
### Summary
Implemented the corrected Braid Consensus DAA with proper distinction between:
- `x̄` (all beads) → hashrate estimation
- `x̄₁` (parents) → damping formula

### Test Results
-  Stubborn Mining: Suppressed (9.81% with 33% hashrate)
-  Stability: Excellent (N_B/N_C = 1.78 ± 0.09)

### Key Finding
DAA converges to N_B/N_C ≈  1.73 due to more accurate lambda.
This appears to be the true equilibrium

### Questions for Review
1. Accept 1.78 as new target, or tune to reach 2.42?
2. Is selfish mining success rate acceptable?
3. Approve a_min = 0.10 or change to NETWORK_SIZE?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,68 +1,57 @@
-# Welcome to BraidPool contributing guide
+# Contributing to Braidpool
 
-Thank you for investing your time in contributing to our project!
+Contributing to open source is one of the best things you can do. It not only benefits you but many others and projects like Braidpool, which aims to solve one of the critical problems in Bitcoin about centralisation; this becomes even more important. Braidpool is a pure open-source project, meaning anyone can run, modify and use this implementation in its own way. Braidpool is designed to remove centralisation of mining pools and provides a way to overcome this issue.
+<br/> <br>
+With this responsibility, we also want to put the right efforts into development so everything goes smoothly and in a collaborative way. For this, there are some guidelines designed that will help us to work together so we can read, write and understand each other's code in an efficient way.
 
-In this guide, you will get an overview of a contribution workflow from opening an issue, creating a PR, reviewing, and merging the PR.
+## Initials
 
-## New contributor guide
+We deeply value each and every contributions made to Braidpool, we take this as a huge step to make Bitcoin safer and sustainable in future. So, new contributors are always needed and appreciated. <br><br>
+Before moving further it's best to familiarize yourself with the Braidpool codebase structure for more information you can refer to 
 
-To get an overview of the project, read the [documentations of BraidPool](./docs/).
+[overview](https://github.com/braidpool/braidpool/blob/main/docs/overview.md), [specification](https://github.com/braidpool/braidpool/blob/main/docs/braidpool_spec.md), 
+and [consensus](https://github.com/braidpool/braidpool/blob/main/docs/braid_consensus.md). This will give you a brief idea about working and its implementation. Braidpool is implemented using Rust, so some familiarity with Rust is also required. The general idea to understand a project is looking into their `tests` directory; the same applies here.
 
-### Issues
+## Structuring a Pull Request
 
-#### Create a new issue
+This covers how you should approach a PR from creating to merging. 
 
-If you spot a problem with the project, [search if an issue already exists](https://github.com/braidpool/braidpool/issues). If a related issue doesn't exist, you can open a new issue using a relevant issue form.
++ Use a concise title that summarizes the change, prefixed by the affected area (e.g., `bead`, `dag`, `rpc`, `doc`, `test`, `feat`, `style`) <br>
+ ``` e.g., bead: Add dynamic difficulty validation. ```
++ Explain what the PR does and why it’s needed and reference related issues `(e.g., Fixes #123)` or discussions `(e.g., Discord thread)`.
++ Limit the PR to a single feature, bug fix, or refactor. Avoid combining unrelated changes. If ignored, this will lead to several other problems. e.g., hard to review, delay feedback, may block others work.
++ Ensure each commit represents a single logical change (e.g., one commit for adding a function, another for tests) to maintain a clean history, make code reviews easier, and allow for efficient debugging or reverting specific changes if needed.
++ Before merging, squash related commits into one; multiple commits for a single logical change seem irrelevant.
 
-#### Solve an issue
+  ```sh
+  git rebase -i HEAD~3
+  # Squash three commits into one with message:
+  bead: Add dynamic difficulty validation
 
-Scan through our [existing issues](https://github.com/braidpool/braidpool/issues) to find one that interests you. You can narrow down the search using `labels` as filters. If you find an issue to work on, and after discussion, it comes out to be a valid issue, you are welcome to open a PR with a fix.
+  Implement `validate_difficulty` in `src/bead.rs` to check bead hashes.
+  Add unit tests in `tests/bead_validation.rs`. Fixes #123.
+  ```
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -4,32 +4,6 @@ Contributing to open source is one of the best things you can do. It not only be
 <br/> <br>
 With this responsibility, we also want to put the right efforts into development so everything goes smoothly and in a collaborative way. For this, there are some guidelines designed that will help us to work together so we can read, write and understand each other's code in an efficient way.
 
-## Key Concepts (Quick Reference)
```
priyashuu, line 7: Hey @Harshit2405-2004, thank you for raising the PR. I noticed that the `Key Concepts (Quick Reference)` section was removed from CONTRIBUTING.md. Could you clarify the reason behind its removal?

CONTRIBUTING.md
```diff
@@ -80,4 +54,5 @@ Here are few steps you can follow while reviewing PRs.
 
 ## Communication Channels
 
+>>>>>>> origin/main
```
priyashuu: Looks like an unresolved merge conflict made it into the commit `(>>>>>>> origin/main in CONTRIBUTING.md).`
Please clean this up and ensure the file is conflict-free.

CONTRIBUTING.md
Harshit2405-2004: Thanks for spotting that! It was an accidental deletion. I will revert the original file.

CONTRIBUTING.md
Harshit2405-2004: Thanks for spotting that! I just resolved these accidental conflicts

CONTRIBUTING.md
```diff
@@ -4,7 +4,7 @@ Contributing to open source is one of the best things you can do. It not only be
 <br/> <br>
 With this responsibility, we also want to put the right efforts into development so everything goes smoothly and in a collaborative way. For this, there are some guidelines designed that will help us to work together so we can read, write and understand each other's code in an efficient way.
 
-## Key Concepts (Quick Reference)
+## Key Concepts (Quick Reference)Expand commentComment on line L7Resolved
```
priyashuu: The section header formatting looks off here.
 `## Key Concepts (Quick Reference)` isn’t rendering cleanly and is showing GitHub UI text like `Expand commentComment on line L7Resolved`.
 This usually happens when the header isn’t properly separated. Could you fix the Markdown so the header renders correctly?

mcelrath:
It appears you implemented the algorithm in my doc below the line explicitly marked "CONTENT BELOW HERE IS OUTDATED" in #91 's braid_consensus.md. This algorithm is definitely wrong for a number of reasons. You should look more carefully at the EMA and SMA algorithms in my jupyter notebook. This PR should come with that document, and a very careful explanation and treatment of how it works and why it works. It DEFINITELY should target 2.42 unless you think my analysis is wrong or have a good argument. Targeting N_B/N_C is the timestamp-free, extremely difficult to attack measure that is extremely important here, and a new feature for any blockchain. None of the DAG-based blockchains have done this before, it's truly new, so much be approached very carefully.

Dunno if this was implemented with AI (that file marker is exactly the kind of thing AI might ignore), and it's okay if you did, but we also just merged a comprehensive AI PR review workflow -- try it. For this PR I want to see:
1. Careful exposition in braid_consensus.md, with figures and math
2. Careful AI review that the implementation matches the spec (I'll run this myself, but you should run it too before submitting the PR)
3. A significant battery of tests (more below)

There are a number of things we need to prove about the DAA to make sure it's acceptable.
1. Wait for the N_B/N_C to stabilize at 2.42, and then suddenly double the hashrate. Prove or show by simulation that the DAA doesn't exhibit "ringing" -- overshooting the new target, then coming back down below, then above, etc. Do the same with halving the hashrate. The transition between two constant hashrate domains should be "critically damped" -- it should smoothly and asymptotically approach the new target defined by the hashrate.
2. N_B = N_C is a singular point in the Lambert W function. If you see this, the algorithm has no idea how "far away" it is from the target, and has nothing to use to gauge how much it should adjust the target. In such a case we need to adjust the lookback window. If you're using 17/7, double it and go to 34/14, or choose a larger N_C that has a nice integer ratio close to 2.41. If you still have N_B=N_C, do it again. As I mentioned, N_C=7 is probably too "short" and doesn't give you enough averaging time to get a good measurement. 7 blockchain-like beads in a row will be a common occurrence (multiple times per day) just due to statistics).
3. Show that the N_B/N_C -> infinity limit behaves sensibly and doesn't have computational DoS problems. I implemented an O(N) cohorts algorithm, this should be largely fixed, and #299 contains some preliminary benchmarks, but we need to make sure the node doesn't get locked up if you increase the hashrate by 100x and cohort computation becomes expensive. This is an actual likely scenario when we launch. On day 1 we have 3 bitaxes. On day 2 we have a 10,000x S21-Pro farm join, and the system must behave sensibly and not crash.
4. In the same vein, show that the N_B/N_C -> 1 limit behaves sensibly. The lambertW function blows up and a naive implementation will fail. This is partially in (1) above, but it's more than just halving the hashrate. This needs to be recursive. If we have to recurse all the way back to genesis to figure out how to adjust the hashrate, then that's what we have to do. If it's blockchain-like all the way back to genesis, then we still have to do something sensible. (e.g. halve the target, and keep halving it every 7 cohorts until we start to see N_B/N_C > 1).  It's still possible that while halving the target we NEVER see cohorts with N_B > 1, because there's only one miner. Something sensible has to happen here too. Driving the difficulty to zero or mindiff will have negative consequences when anyone plugs in an S21 and generates 100,000 beads/s.
5. Show by simulation or math that the variance induced by the DAA is not significantly larger than the inherent variance at constant difficulty. The formula is in braid_consensus.md. It should be statistically impossible for the DAA-modified variance to be *less* than the natural variance at constant hashrate, but the difference should be as small as possible. If you have parameters in your DAA, a good thing to do would be to minimize those parameters with respect to this variance. This requires careful treatment. Such an analysis will drive N_C to infinity, which is obviously dumb.
6. An analysis of how bitcoin's hashrate has changed over time would be valuable here. What's the natural variance? Does our DAA behave reasonably over it's natural variance? Of course, this is a *pool* not the entire bitcoin blockchain, and miners can and do switch pools, so pools see higher variance than the main chain. Nonetheless, proving our DAA is "sensible" given bitcoin's "natural" variance is important.
7. Compare the distribution of N_B/N_C using NO DAA and constant hashrate and difficulty to the one with your DAA using a Kolmogorov-Smirnov test. Seek to minimize this. (this overlaps with (5) but variance is the second moment of the probability distribution, K-S will capture higher moments too) by targeting 2.42 you can easily hit only the FIRST moment of the distribution (the mean). We can't "fix" latency through the DAA. The distribution WILL be modified by the DAA, but we want that to be as small as possible, simulating/mathing at constant difficulty.

Did I mention that this is the **hardest** remaining problem in Braidpool? None of the above problems are insurmountable. But it will be challenging. Take the challenge. A lot of really, really dumb things have been done with shitcoins over the years on this topic. I've put over a decade of thought into this and eliminated a bunch of dumb options for you (and timestamps!) 

I would LOVE to see mathematical proofs of how the DAA behaves, if you can do it. If  you're using AI, ask the AI to create a proof. In either case I'm going to beat the ever loving shit out of this code, because I don't want Braidpool to die on day 2, and this algorithm is one of our biggest attack surfaces.

mcelrath:
Closing due to inactivity. @Harshit2405-2004 if you've had a chance to work more on this, please re-open or make a new PR.

Issue #302, opened 2025-11-28
Removing unwanted/unsafe unwraps with suitable handler


Sansh2356:
- The current code coverage has many potential unsafe `unwrap()` around  which are not suitable, audit each of those existing unwraps and replace those with appropriate result handling .
- Update corresponding errors in `error.rs` and handle the unwrap safely, can use `anyhow` or other crate but after suitable consideration. 
- Replace all possible `unwrap()` safely .

Rustix69:
@Sansh2356 Can I pick up this issue ??

Sansh2356:
Assigned @Rustix69 .

Sansh2356:
@Rustix69 any updates wrt to this ?

Rustix69:
@Sansh2356 Will raise teh PR by tonight ...

Rustix69:
@Sansh2356 This change will be too much larger so should we focus on the files priority wise ?

The `dev` branch has 335 unwrap() + 31 expect() = 366 total calls across 16 files.

Sansh2356:
@Rustix69 not all unwraps are to be taken care of we don't want to be extremely timid `wrt` to removing unwraps in those cases also where it is impossible. You can opt for multiple PRs but i don't think that is required handling will be spread over almost all the files but resolving those won't be requiring much larger LOC . If you want to submit multiple PRs go ahead .

Sansh2356:
@Rustix69 still working over this ?

Rustix69:
@Sansh2356 I am having my sem papers i had worked on it will raise 2 PRs by tonight for sure ...

Sansh2356:
@Rustix69 still working on this ?

parthdude07:
Hi! @Sansh2356  I noticed the recent check-in regarding the status of this issue.

If the current assignee is no longer working on it or is unavailable, I would be happy to take this over. I am available to work on this actively and can pick it up immediately. Thanks!

Sansh2356:
@parthdude07 let's wait for @Rustix69 to reply he was already working on this .

Rustix69:
> [@Rustix69](https://github.com/Rustix69) not all unwraps are to be taken care of we don't want to be extremely timid `wrt` to removing unwraps in those cases also where it is impossible. You can opt for multiple PRs but i don't think that is required handling will be spread over almost all the files but resolving those won't be requiring much larger LOC . If you want to submit multiple PRs go ahead .

Hey @Sansh2356 Should I work on Next PR ?

[...]

PR #393, opened 2026-02-18
Replace unsafe unwrap/expect calls with proper error handling in Phase 2 modules

Rustix69:
Fixes : #302 

## Summary

Phase 2 of the unwrap/expect cleanup: replaces unsafe `unwrap()` and `expect()` calls with proper `Result` handling in production code.

## Files Changed

- `node/src/db/db_handlers.rs` - DB row parsing, index lookups, `MedianTimePast`, `Signature`, extranonce
- `node/src/braid/mod.rs` - `reverse()` HashMap lookup, `highest_work_path` bead selection
- `node/src/rpc_server.rs` - HTTP client build, `ArrayParams::insert`, server build/`local_addr`, `get_bead` serialization
- `node/src/behaviour/mod.rs` - Response channel sends (`respond_with_beads`, `respond_with_beadhashes`, etc.)
- `node/src/template_creator.rs` - `encode_varint` (replaced `expect` with manual varint encoding)


Sansh2356 (commented):
t-ACK .

mcelrath (changes requested):
Hey @Rustix69 do you think you could fix the merge conflicts and address the following issues?

● Rust Review: Replace unsafe unwrap/expect calls with proper error handling in Phase 2 modules (PR #393)

  Summary

  Review complete. There’s a build‑blocking edit in rpc_server.rs, plus a new silent‑failure path in DB fetch logic.

  Findings

  🔴 Critical

   1. parse_arguments is syntactically incomplete (won’t compile).
  The new function’s match never closes, and the function never closes before the next #[derive] block, so rpc_server.rs 
  doesn’t parse.
  File: node/src/rpc_server.rs (around lines ~112–145)

  🟠 High

   1. fetch_bead_by_bead_hash now silently drops row‑parse errors.
  The map(|row| { ... Ok(()) }) produces Option<Result<_, DBErrors>>, but the inner Err is never checked. This converts parse
   failures into “success” with partially populated fetched_bead.
  File: node/src/db/db_handlers.rs

  🟡 Medium

   1. get_bead added unused JSON serialization with wrong return type.
  The method returns Bead, but now constructs String via serde_json::to_string and discards the Result, adding a must_use 
  warning and wasted work.
  File: node/src/rpc_server.rs
   2. Missing parent mapping now silently drops edges.
  Replacing unwrap() with if let Some avoids panic but hides invariant violations; you likely want to surface/log and abort 
  to avoid corrupting HWP calculations.
  File: node/src/braid/mod.rs

  Code Samples

```rust

   // rpc_server.rs (parse_arguments is incomplete)
   pub async fn parse_arguments(...) -> () {
       ...
       let (rpc_method, method_params) = match cli_command {
           RpcCommand::AddBead { ... } => { ... }
           RpcCommand::GetBead { ... } => { ... }
   // <-- missing closing braces for match + function

   // db_handlers.rs: inner Result is ignored
   match sqlx::query("SELECT * FROM bead WHERE hash = ?")
       .bind(...)
       .map(|row| { ... Ok(()) })
       .fetch_optional(...)
       .await
   {
       Ok(_rows) => { /* _rows: Option<Result<_, DBErrors>> ignored */ }
       Err(e) => return Err(DBErrors::TupleNotFetched { ... }),
   }

   // rpc_server.rs: unused JSON serialization with wrong return type
   match bead {
       Some(bead) => {
           let json = serde_json::to_string(&bead)?; // String
           Ok(json) // Result<String, _> (unused)
       }
       None => Err(...)
   }
   bead.ok_or_else(...) // returns Result<Bead, _>
```

Rustix69:
Hey @mcelrath I need 2 days i will push the fixes. Thanks

Atishyy27:
Rustix69, are you still planning to pick this back up? If not, would you mind if I carried just the behaviour/mod.rs hunk (the five send_response expects) as a separate small PR against dev? It is the only part of this change a remote peer can trigger: if the peer times out or disconnects before we answer, send_response returns Err and the expect panics the swarm task. Your if-let-warn conversion there is exactly right and I would credit this PR for it. No rush either way, happy to leave it alone if you are continuing.

PR #397, opened 2026-02-26
node: Replace custom rust-bitcoin fork with official crate

### LINKED ISSUE #396: Removing customized `rust-bitcoin` dependency
Currently we are using a custom version of [rust-bitcoin](https://github.com/braidpool/rust-bitcoin/tree/cpunet) for our cpunet but in reality we need very few things to be customized from the rust-bitcoin; otherwise all other functionalities are needed the same as the original. But these small customisations have their own drawbacks, like maintaining a whole rust-bitcoin dependency and not being able to use the same functions for the other behaviours. For example, we are using the customized `validate_pow` function from the rust-bitcoin but we can't use the same function to compute the actual pow in any other places of the codebase because it was intended to be used as a customized one. This will lead to using the two rust-bitcoin dependencies (original and customized) which is more worse. The solution for this is to carve out the functions which need to be modified and implement them in the Braidpool under some file `util.rs` and use these functions for customized behaviour.

Rustix69:
Remove dependency on braidpool/rust-bitcoin cpunet fork and switch to the official rust-bitcoin/rust-bitcoin at the same upstream commit.
- Fixes #396 

The three customizations from the fork (cpunet block hash with "cpunet\0" suffix, Network::CPUNet variant, and tc1 address prefix) are now implemented locally in a new node/src/cpunet.rs module:

- Add BraidpoolNetwork enum wrapping bitcoin::Network with CPUNet
- Add cpunet_block_hash() and cpunet_validate_pow() functions
- Add Bead::bead_hash() as single dispatch point for bead identification
- Replace all block_header.block_hash() calls with bead_hash()
- Update config, main, ipc, stratum, and template_creator for new types
- CPUNet maps to Regtest for standard bitcoin library operations

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -385,10 +385,10 @@ dependencies = [
 [[package]]
 name = "base58ck"
 version = "0.2.0"
-source = "git+https://github.com/braidpool/rust-bitcoin.git?branch=cpunet#156f317b6a90d2ddfb9c7b74ac8ff36b705c52b8"
+source = "git+https://github.com/rust-bitcoin/rust-bitcoin.git?rev=a419fc9aa689b576e4d605aa00b77a5298d2b185#a419fc9aa689b576e4d605aa00b77a5298d2b185"
 dependencies = [
  "bitcoin-internals 0.4.0",
- "bitcoin_hashes 0.16.0 (git+https://github.com/braidpool/rust-bitcoin.git?branch=cpunet)",
+ "bitcoin_hashes 0.16.0 (git+https://github.com/rust-bitcoin/rust-bitcoin.git?rev=a419fc9aa689b576e4d605aa00b77a5298d2b185)",
 ]
 
 [[package]]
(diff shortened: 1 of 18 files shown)


Sansh2356 (changes requested):
Concept NACK [a928d49](https://github.com/braidpool/braidpool/pull/397/changes/a928d490de86da94d6b950d3d62ab3d1c57eee98)

The intention of #396 is to utilize wrappers around the base types being provided from the `rust-bitcoin` but not only for the `cpunet` since braidpool can operate on any specific network type whether it is `mainnet`, `testnet3/4` ,`signet` etc . The way of calculating `PoW` is different only for `cpunet` and not for others so on the basis of cli argument passed during runtime in `cli.rs` which is the selected network construct the wrapper accordingly that will handle both the cases of `cpunet` and other network types, so kindly restructure the PR accordingly .

zaidmstrr (commented):
Concept NACK [a928d49](https://github.com/braidpool/braidpool/pull/397/changes/a928d490de86da94d6b950d3d62ab3d1c57eee98)

The PR doesn't seem to look like it's achieving the desired goal and this changes the consensus functions in a hard, non-backward-compatible way. Also, this PR introduces the restrictions of moving from the testing to deployment phase by introducing the hardcoded custom function which will not change when moving to mainnet. Additionally, the PR looks like a change generated by some coding agent without any review, customizations or proper analysis of the internal working. Request you to close this PR.

Rustix69:
Thanks for the review, acknowledged.

I’ll rework this to match #396 by:
- using network-agnostic wrappers over rust-bitcoin types,
- selecting behavior at runtime from `cli.rs`,
- keeping non-cpunet consensus behavior unchanged,
- applying custom PoW/hash only for `cpunet`,
- adding tests for backward compatibility + cpunet-specific behavior.

If this PR is too noisy after rework, I’ll open a clean superseding PR.

**CC :** @zaidmstrr

zaidmstrr:
> If this PR is too noisy after rework, I’ll open a clean superseding PR.
> 
> **CC :** @zaidmstrr

Try opening the new PR considering all the above facts. Closing this for now.

PR #398, opened 2026-03-01
Adding extension for primitive types required and exported from rust-…

Sansh2356:
- Adding extensions for `Network`, `Hrp` and `Params` to reduce the issue of maintaining both the bech32 and rust-bitcoin fork as per our deps that will also hamper the SV2 integration due to mixed version of same deps and also will include the extensions required to support `cpunet`.
- There were other inclusion to these in our forked repository to accommodate the full inclusion of `cpunet` such as genesis block hash and  Magic which is required for establishing the p2p message  communication but were not in use or would not be used at least in the near changes so have not included those.
- Addressing #396 .


Sansh2356:
Updating this PR instead of closing with newer adaptation of implementing the `Cpunet` struct itself for handling all the functionalities related to `Hrp`,`Params` and parsing etc .

zaidmstrr (commented):
Approach ACK [dc1fd3e](https://github.com/braidpool/braidpool/pull/398/changes/dc1fd3eef7f9953f7bd37e0f9562ca3058777b86)

Thanks for taking this up. The approach used here is exactly what I thought initially about creating a separate utility file which serves the purpose of custom cpunet. This PR does the exact same. I haven't reviewed the code yet.

zaidmstrr (commented):
As the goal of this PR is to remove the cpunet dependency from the custom `rust-bitcoin`. But I still see the dependency of custom `rust-bitcoin` in `Cargo.toml`. I think it would be good if you replaced that with the non custom version of it.

node/src/cpunet.rs
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    #[inline]
+    pub fn is_cpunet_hrp(hrp: &str) -> bool {
+        hrp.eq_ignore_ascii_case(CPUNET_HRP)
+    }
+
+    /// Encodes a witness program as a cpunet bech32m or bech32 address depending upon `WitnessVersion` for non-taproot and taproot specific addresses.
+    pub fn encode_bech32_address(program: &WitnessProgram) -> String {
+        let hrp = bech32::Hrp::parse(CPUNET_HRP).unwrap();
```
zaidmstrr: `unwrap()` can we removed from here.

node/src/cpunet.rs
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    #[inline]
+    pub const fn name() -> &'static str {
+        CPUNET_NAME
+    }
+
+    #[inline]
+    pub fn is_cpunet_name(name: &str) -> bool {
+        matches!(name.to_lowercase().as_str(), "cpunet")
```
zaidmstrr: Here allocating a heap is not necessary you can just simply use `eq_ignore_ascii_case`.

node/src/cpunet.rs
```diff
@@ -0,0 +1,214 @@
...
+pub struct Cpunet;
+
+impl Cpunet {
+    #[inline]
+    pub const fn bech32_hrp() -> &'static str {
+        CPUNET_HRP
+    }
+
+    #[inline]
+    pub const fn name() -> &'static str {
```
zaidmstrr: Same for here, these can be called directly without wrapper.

node/src/cpunet.rs
```diff
@@ -0,0 +1,214 @@
...
+    }
+}
+
+/// Implementation of Cpunet specific `Network` and `Address` decoding/encoding functionality.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Cpunet;
+
+impl Cpunet {
+    #[inline]
+    pub const fn bech32_hrp() -> &'static str {
```
zaidmstrr: I don't think necessity for this wrapper and it's also not used anywhere else. One can directly access these constants.

node/src/cpunet.rs
```diff
@@ -0,0 +1,214 @@
...
+
+/// Human-readable part for cpunet bech32 addresses: "tc"
+pub const CPUNET_HRP: &str = "tc";
+
+/// The core-arg name along with that the actual network name as string-slice
+pub const CPUNET_NAME: &str = "cpunet";
+
+/// Cpunet consensus parameters.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CpunetParams {
```
zaidmstrr: This struct is not used anywhere in the codebase. What's the reason for this? Seems like the intended behaviour was to use this later with difficulty adjustment. But currently we don't have such a thing existing, and no final architecture is decided, so it's good to remove this for now. So when later we work on that thing, we can create the structure according to the final needs at that time.

zaidmstrr (commented):
I have some suggestions described below. You can look at them.

Sansh2356:
@zaidmstrr applied the suggestions .

mcelrath:
● PR #398 status: NEEDS‑WORK. Multiple reviewers agree the cpunet hash migration isn’t fully wired through, and there
  are correctness blockers.

  High‑priority blockers

   1. RPC bead IDs are wrong on cpunet. Several endpoints still use block_header.block_hash() instead of the new cpunet 
  hash, which breaks lookups (notably get_genesis, get_parents, get_children, get_node_info). File: 
  node/src/rpc_server.rs (e.g., around lines 386, 611, 660, 870).
   2. DB migration required. Existing cpunet rows hashed with the old algorithm won’t match new lookups; parent 
  reconstruction can fail without a migration/rebuild of stored Bead.hash (db layer / handlers).
   3. Panic paths introduced. consensus_decode uses unwrap() in node/src/uncommitted_metadata/mod.rs:52, and expect() 
  exists in braidpool-common/src/cpunet.rs:36. These can panic on malformed input.

  Medium / notes

   - compute_block_hash checks "cpunet" case‑sensitively, risking silent fallback and hash/PoW divergence 
  (node/src/utils/mod.rs).
   - FinalTemplate::block_hash still uses standard SHA256d; for cpunet this can diverge from the new hash if surfaced 
  externally.
   - Minor perf note: avoid double hashing in stratum.rs (compute once and reuse).

Sansh2356:
> ● PR #398 status: NEEDS‑WORK. Multiple reviewers agree the cpunet hash migration isn’t fully wired through, and there are correctness blockers.
> 
> High‑priority blockers
> 
>     1. RPC bead IDs are wrong on cpunet. Several endpoints still use block_header.block_hash() instead of the new cpunet
>        hash, which breaks lookups (notably get_genesis, get_parents, get_children, get_node_info). File:
>        node/src/rpc_server.rs (e.g., around lines 386, 611, 660, 870).
> 
>     2. DB migration required. Existing cpunet rows hashed with the old algorithm won’t match new lookups; parent
>        reconstruction can fail without a migration/rebuild of stored Bead.hash (db layer / handlers).
> 
>     3. Panic paths introduced. consensus_decode uses unwrap() in node/src/uncommitted_metadata/mod.rs:52, and expect()
>        exists in braidpool-common/src/cpunet.rs:36. These can panic on malformed input.
> 
> 
> Medium / notes
> 
>     * compute_block_hash checks "cpunet" case‑sensitively, risking silent fallback and hash/PoW divergence
>       (node/src/utils/mod.rs).
> 
>     * FinalTemplate::block_hash still uses standard SHA256d; for cpunet this can diverge from the new hash if surfaced
>       externally.
> 
>     * Minor perf note: avoid double hashing in stratum.rs (compute once and reuse).

The issue number 2 have been identified earlier and require separation of DB structures according to the network kind being used, will be filing an issue with regards to this by tomorrow .

mcelrath:
Main blockers

   1. node/src/rpc_server.rs still mixes block_header.block_hash() with the braid’s network-aware bead hash, so cpunet RPC 
  identities are inconsistent across lookup/reporting endpoints.
   2. node/src/main.rs and node/src/utils/mod.rs still treat network_name inconsistently; invalid or mixed-case values can 
  split hashing/PoW behavior across components.
   3. node/src/db/db_handlers.rs has no migration/rebuild path for existing cpunet DB rows whose persisted bead hashes use 
  the old algorithm.
   4. node/src/uncommitted_metadata/mod.rs and node/src/committed_metadata/mod.rs still have panic-on-decode paths via 
  unwrap().
   5. node/src/rpc_server.rs exposes bitcoinproxy as an unauthenticated local relay to privileged bitcoind RPC methods.
   6. braidpool-common/src/lib.rs still exposes public APIs through crate-private error types.

  Prior review status: the earlier May findings are mostly still open. The only material one that looks resolved is the 
  expect() concern in braidpool-common/src/cpunet.rs, because that path now relies on validated WitnessProgram invariants.

nkatha23 (commented):
Concept ACK. The de-forking direction looks correct, especially since we won't need rust-bitcoin after this PR.

tACK  [`55b87da`](https://github.com/braidpool/braidpool/pull/398/commits/55b87da) . Other than the points already raised in the reviews, the changes look good to me.

zaidmstrr:
This needs a rebase on top of current dev.

node/src/template_creator.rs
```diff
@@ -405,17 +383,24 @@ pub fn build_braidpool_coinbase_from_template(
...
-        .require_network(config.network)
-        .map_err(|_| CoinbaseError::AddressNetworkMismatch)?;
+    // According to the type of chain/network derive the pubkeyscript
+    let payout_script = if config.is_cpunet {
+        // Use cpunet module for cpunet addresses (tc1... prefix)
+        Cpunet::decode_bech32_address(&config.pool_payout_address)
+            .map_err(|e| CoinbaseError::InvalidBitcoinAddress(e.to_string()))?
+    } else {
+        // Use standard rust-bitcoin address parsing for other networks
+        let network = config.get_network();
```
zaidmstrr: Here, you can remove this `get_network` function as it's used here and not any other place. And secondly, the current function silently binds to `Regtest` if someone is using the cpunet which can produce wrong results like validation errors and address translations which are unique to the cpunet. You can change this with:
 ``` sh
let network = config.network
        .expect("Config must have a network type");
```
Or for any other reasons. If you think that `get_network` function is necessary then keep that but apply the same changes as mentioned there.

File: node/src/stratum.rs [hunk 7]
```diff
@@ -1255,9 +1277,7 @@ impl DownstreamClient {
 
                 let broadcast_time = std::time::SystemTime::now()
```
zaidmstrr, line 1278: And you can replace this function with:

``` diff
- let broadcast_time = std::time::SystemTime::now()
-                    .duration_since(std::time::UNIX_EPOCH)
-                    .map(|d| Time::from_consensus(d.as_secs() as u32).unwrap())
-                    .unwrap();
+  let broadcast_time = std::time::SystemTime::now()
+                    .duration_since(std::time::UNIX_EPOCH)
+                   .map_err(|e| e.to_string())
+                   .and_then(|d| {
+                      let ts: u32 = d.as_secs().try_into().map_err(|e: 
+ std::num::TryFromIntError| e.to_string())?;
+                        Time::from_consensus(ts).map_err(|e| e.to_string())
+                    })
+                   .map_err(|e| 
+ StratumErrors::ErrorFetchingCurrentUNIXTimestamp { error: e })?;
```
This provides better error handling.

node/src/config.rs
```diff
@@ -106,28 +107,67 @@ impl BraidpoolConfig {
...
+            let core_arg = match network_name {
+                "main" | "mainnet" | "bitcoin" => "main",
+                "testnet" | "testnet3" | "test" => "test",
+                "testnet4" => "testnet4",
+                "signet" => "signet",
+                "regtest" => "regtest",
+                other => other,
+            };
+            //Deriving actual type for utilizing the underlying functionality of deriving script_pubkey
+            let network = Network::from_core_arg(core_arg).unwrap_or(Network::Bitcoin);
```
zaidmstrr: Same here just don't bind the fallback network in case of error, as it can produce unexpected behaviour. Instead, return an error stating the reason for failure.

node/src/config.rs
```diff
@@ -106,28 +107,67 @@ impl BraidpoolConfig {
 
 #[derive(Debug, Clone)]
 pub struct CoinbaseConfig {
-    pub network: Network,
+    /// Network name as string (e.g., "main", "testnet4", "signet", "regtest", "cpunet")
+    pub network_name: String,
```
zaidmstrr: There is no practical use and need of `network_name` here in this struct. The `network` field is sufficient for representing the network. Although it can't represent the `cpunet` but that's fine here. Also, this field is never read by any other function or method.

node/src/config.rs
```diff
@@ -106,28 +107,67 @@ impl BraidpoolConfig {
...
+            }
+        } else {
+            // Convert network name to core arg format if needed
+            let core_arg = match network_name {
+                "main" | "mainnet" | "bitcoin" => "main",
+                "testnet" | "testnet3" | "test" => "test",
+                "testnet4" => "testnet4",
+                "signet" => "signet",
+                "regtest" => "regtest",
+                other => other,
```
zaidmstrr: Using `other => other` is not correct here as it can produce unexpected results, and with the current situation, any other value will bind to default settings internally, so a user will not know whether the args he passed are actually incorrect.

node/src/config.rs
```diff
@@ -106,28 +107,67 @@ impl BraidpoolConfig {
...
+                network_name: "cpunet".to_string(),
+                network: None,
+                pool_payout_address: "tc1qu3cdq9unyhdc3d2hw8mvpfgnnhvp6ucckkl6ft".to_string(),
+                pool_identifier: "Braidpool".to_string(),
+                is_cpunet: true,
+            }
+        } else {
+            // Convert network name to core arg format if needed
+            let core_arg = match network_name {
+                "main" | "mainnet" | "bitcoin" => "main",
```
zaidmstrr: Providing different aliases is good but not a very solid idea in my opinion. All users should be aware of some fixed set of network types they can use. And if they don't, instead of fallback return error.

node/src/config.rs
```diff
@@ -106,28 +107,67 @@ impl BraidpoolConfig {
...
-        Self {
-            network,
-            pool_payout_address,
-            pool_identifier: "Braidpool".to_string(),
+    /// Creates CoinbaseConfig from a network name string.
+    ///
+    /// For cpunet, uses our custom cpunet module.
+    /// For other networks, uses `Network::from_core_arg()` from rust-bitcoin.
+    pub fn from_network_name(network_name: &str) -> Self {
+        let is_cpunet = Cpunet::is_cpunet_name(network_name);
```
zaidmstrr: This is not a bug, but there is a slight redundant edge case with the current approach, as `is_cpunet` can be derived in two ways: first, using the `network_name` or by `network.is_none()`. Which creates a little redundancy here on the existence of the `is_cpunet` flag? This might not create a problem now, but in the future some other dev can miss this detail and can use both the fields wrongly by setting the network: `Some(Network::Bitcoin)` and `is_cpunet: true` which is incorrect here according to the current logic. This issue is more of developers related then the actual users.

node/Cargo.toml
```diff
@@ -37,8 +32,9 @@ strum = { workspace = true }
 if-addrs = { workspace = true }
 anyhow = { workspace = true }
 socket2 = { version = "0.5", features = ["all"] }
+bitcoin-internals = "0.5.0"
```
zaidmstrr: I found no usage of this library anywhere in the codebase.

zaidmstrr (commented):
The thing I have noticed is that when we run the braidpool node but without passing the network type as an argument, the node binds to the `mainnet` by default, and thus further mining using this can lead to invalid shares generated by the miners and creates a confusing state. I think we should make the passing network type compulsory for users. As it's an important factor to decide and take certain actions in the codebase. 

You can discuss any issue. Apart from that, I have identified several more things that can be improved for all of those identifications. I have written the separate comments related to each one of them below.

node/src/config.rs
Sansh2356: Yes, it is better to have an error type in `error.rs` and that will be thrown in case of any other type and the `Network` member of struct in itself can point out current network by making it as an `Option<Network>`if it is one of the bitcoin core network it can be represented as a concrete `Some()` else in case of `None` it will be parsed from provided fixed network or their aliases and converted to Option .

Sansh2356:
> The thing I have noticed is that when we run the braidpool node but without passing the network type as an argument, the node binds to the `mainnet` by default, and thus further mining using this can lead to invalid shares generated by the miners and creates a confusing state. I think we should make the passing network type compulsory for users. As it's an important factor to decide and take certain actions in the codebase.
> 
> You can discuss any issue. Apart from that, I have identified several more things that can be improved for all of those identifications. I have written the separate comments related to each one of them below.

Yes, the users not passing and we defaulting it to mainnet can create a confusing state instead we can make the network argument required in `cli.rs` thus forcing user to spawn with a network flag instead .

braidpool-common/src/cpunet/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+
+        let witness_version = bitcoin::WitnessVersion::try_from(version.to_u8())
+            .map_err(|_| CpunetAddressError::InvalidWitnessVersion(version.to_u8()))?;
+        let witness_program = WitnessProgram::new(witness_version, &data)
+            .map_err(|e| CpunetAddressError::InvalidProgram(e.to_string()))?;
+
+        Ok(ScriptBuf::new_witness_program(&witness_program))
+    }
+    /// Returns the block hash.
+    pub fn block_hash(header: Header) -> BlockHash {
```
nkatha23: Could this take &Header instead of Header?The function only reads the header and doesn't need ownership

braidpool-common/src/cpunet/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+        let mut engine = sha256d::Hash::engine();
```
nkatha23, line 68: Could this take &Header instead of Header?The function only reads the header and doesn't need ownership

braidpool-common/src/cpunet/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+                expected: CPUNET_HRP.to_string(),
+                found: hrp.to_string(),
+            });
+        }
+
+        let witness_version = bitcoin::WitnessVersion::try_from(version.to_u8())
+            .map_err(|_| CpunetAddressError::InvalidWitnessVersion(version.to_u8()))?;
+        let witness_program = WitnessProgram::new(witness_version, &data)
+            .map_err(|e| CpunetAddressError::InvalidProgram(e.to_string()))?;
+
```
nkatha23: Could we preserve the underlying WitnessProgram construction error here instead of stringifying it e.g. `InvalidProgram(bitcoin::witness_program::Error) `,so source() can return it too, consistent with the Bech32 variant

node/src/stratum.rs
```diff
@@ -892,15 +896,31 @@ impl DownstreamClient {
         block_transactions.extend(submitted_job.blocktemplate.transactions.clone());
 
         // Construct and log the complete block using rust-bitcoin's Block struct
-        let complete_block = bitcoin::Block::new_unchecked(header, block_transactions);
+        let complete_block = bitcoin::Block {
+            header,
+            txdata: block_transactions,
+        };
+        // For cpunet, use custom block_hash calculation; otherwise use standard validate_pow
+        let is_cpunet = Cpunet::is_cpunet_name(&self.network_name);
```
nkatha23: The Cpunet network check is recomputed on every share submission, even though `network_name` is fixed for the lifetime of the connection. Could we cache this at connection setup (e.g. as a `bool `on `DownstreamClient`) and reuse it here?

node/src/stratum.rs
```diff
@@ -821,20 +825,20 @@ impl DownstreamClient {
...
+            time: ntime_u32,
             bits: submitted_job.blocktemplate.bits,
             nonce: nonce_u32,
         };
         let compact_target = submitted_job.blocktemplate.bits;
         let target = bitcoin::Target::from_compact(compact_target);
         debug!(
             connection_id = %connection_id_hex,
-            target = %target.to_hex(),
+            target = %hex::encode(target.to_le_bytes()),
```
nkatha23: this looks like it should be to_be_bytes() match the convention used for time_be_hex, bits_be_hex   etc

node/src/stratum.rs
```diff
@@ -1910,6 +1939,7 @@ impl Default for DownstreamClient {
             miner_extranonce2_size: EXTRANONCE2_SIZE,
             monitor_target: None,
             block_submission_tx: None,
+            network_name: String::new(),
```
nkatha23: network_name: String now lives on both Server and DownstreamClient as independent owned copies, nothing enforces these stay in sync.May be consider a shared Arc<str> (or folding it into whatever type ends up representing network identity)rather than two copies that could drift,  idk

also here; [ ```
/// Network name (e.g., "main", "testnet", "cpunet") used for network-specific behavior
    network_name: String,
``` ](https://github.com/braidpool/braidpool/pull/398/changes#diff-0610d2446fcc8639c24afe1f4774a5bba949f356eba7ea45879012da5d8106ceR1962-R1963)    /// Network name (e.g., "main", "testnet", "cpunet") used for network-specific behavior
    network_name: String,

**Review by nkatha23 — COMMENTED** [review]
Code ACK https://github.com/braidpool/braidpool/pull/398/commits/1435e2372d6f9076ea86c0e5ab47671b27366fe0. Went through cpunet/mod.rs, error.rs, and the stratum.rs; the core logic looks correct and matches the intent of the PR.

Follow-up suggestion: make Cpunet and Bitcoin mutually exclusive at the type level, with a single source of truth shared by Server/DownstreamClient, instead of parallel Option<Network> + bool/String fields.

File: node/src/stratum.rs (hunk 19 above)
Comment (zaidmstrr) [reply node/src/stratum.rs]: You can actually simplify the `network_name` type and can reduce its complexity by introducing the enum instead of the String:

```sh
enum PoolNetwork {
    Cpunet,
    Bitcoin(bitcoin::Network),
}
``` 

This can greatly simplify the logic and consistency of the code.

Sansh2356:
I have updated the PR kindly take a look again @nkatha23 and @zaidmstrr  .

nkatha23 (approved):
Code ACK https://github.com/braidpool/braidpool/pull/398/commits/a1ad52680166ebbddb311156e7696017e5fa8ca8

zaidmstrr (commented):
One more thing I have noticed is that. If we are on audit mode, we can still pass the `network` as `cpunet` or any other type. This can lead to wrong share verification. To fix this, I think you can introduce a statement like:

```sh
if args.audit && network != config::PoolNetwork::Bitcoin(bitcoin::Network::Bitcoin) {
    error!(
        network = %network,
        "Audit mode requires mainnet; cannot run on {}",
        network
    );
    return Err("audit mode requires --network mainnet".into());
}
```

Sansh2356:
> One more thing I have noticed is that. If we are on audit mode, we can still pass the `network` as `cpunet` or any other type. This can lead to wrong share verification. To fix this, I think you can introduce a statement like:
> 
> ```shell
> if args.audit && network != config::PoolNetwork::Bitcoin(bitcoin::Network::Bitcoin) {
>     error!(
>         network = %network,
>         "Audit mode requires mainnet; cannot run on {}",
>         network
>     );
>     return Err("audit mode requires --network mainnet".into());
> }
> ```

I have added an additional check for this .

zaidmstrr (approved):
Code ACK [0719de8](https://github.com/braidpool/braidpool/pull/398/changes/0719de849790a0c71a781f2c1deac6633ef3725b)

Thanks for implementing all the suggested changes. This now looks good.

PR #399, opened 2026-03-02
feat: add persistent SQLite database and state snapshotting

zaidmstrr:
This PR adds the functionality of having a persistent database into the audit mode; you can treat this PR as a follow-up of [PR#305](https://github.com/braidpool/braidpool/pull/305). With this you can mine shares on an upstream pool while auditing all the things related to each miner's work (hashrate) statistics etc and save them safely on your local persistence storage. The saved records can we used to verify the active involvement of a miner mining shares or producing some hashrate on some period X. 

This also supports restarts and stopping and starting the mining work for some interval; for example, if someone wants to mine on some particular hours of a day or on some intervals, then he can stop the node and restart later whenever he wants, and the DAG chain will completely restart and recover itself from the last mining info and will start mining on top of previous DAG tips, leading to no breakage of DAG and creating a valid, consistent chain of records. For this the node uses a feature called _State Snapshotting_ or _Pruned Initialization_ in which we restart a node using the valid last snapshot of data without feeding or initializing with the entire data. In our case a node can create a valid in-memory bead as a starting point by fetching the last valid snapshot from the database..

### DIFF
--- a/node/src/audit.rs
+++ b/node/src/audit.rs
@@ -1,9 +1,14 @@
 use crate::bead::Bead;
 use crate::braid::{AddBeadStatus, Braid};
+use crate::committed_metadata::CommittedMetadata;
+use crate::db::audit_db_handlers::AuditDBHandler;
+use crate::uncommitted_metadata::UnCommittedMetadata;
+use crate::{TimeVec, TxIdVec};
 use bitcoin::consensus::serialize;
 use bitcoin::hashes::sha256d;
-use bitcoin::BlockHash;
+use bitcoin::{BlockHash, BlockHeader, CompactTarget, TxMerkleNode};
 use serde::{Deserialize, Serialize};
+use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::SystemTime;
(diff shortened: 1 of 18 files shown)


Sansh2356 (commented):
t-ACK [`5e37e2a`](https://github.com/braidpool/braidpool/pull/399/changes/5e37e2aa4e6ce1be5016f66e633a78cf29b62e70) .

mcelrath:
● Review complete. The audit DB + upstream proxy integration is promising, but there are several correctness issues that
  will break audit mode or risk panics.

  Findings

  🔴 Critical

   - node/src/db/audit_schema.sql defines AuditTips using parent_composite_hash, which doesn’t exist in AuditBeadParent 
  (lines ~60–63). The view will fail or be unusable. This should likely join on parent_block_hash (and 
  AuditBead.block_hash) or add the missing column.

  🟠 High

   - node/src/db/audit_db_handlers.rs uses multiple unwrap()s during DB row parsing (e.g., pubkey, signature, timestamps,
   hash byte conversions) and unwrap_or_default() on the parent query (lines ~155–223). Any malformed row or query 
  failure will panic or silently drop parents. These should map to DBErrors and be propagated.
   - node/src/stratum.rs sends mining.set_extranonce with params: [new_extranonce1_hex, 1] for all miners during 
  commitment updates (lines ~3025–3029). This ignores the per-miner miner_extranonce2_size computed at subscribe. If 
  upstream extranonce2 size ≠ 8, miners will roll the wrong size and mining.submit will be rejected. Use the connection’s
   stored size instead of 1.

  🟡 Medium

   - node/src/db/audit_db_handlers.rs pairs HashSet parents with an ordered timestamp vector (lines ~102–113 and 
  ~212–223). If parent order matters, timestamps can mismatch. Consider storing parent+timestamp as pairs to preserve 
  association.

mcelrath (changes requested):
See the comment with the AI review. Please fix these things or justify why the AI is wrong.

zaidmstrr:
I have removed the unwraps() to be extra cautious. While in high priority (2) is wrong as using 1 byte as extranonce 2 is intentional in our case, and we are constructing the full extranonce2 before submitting the share to the upstream pool and this variant `NotifyCmd::UpdateExtranonce ` only valid for audit mode. The critical (1) is also fixed; a single line was left while I changed the bead unique reference hash from composite to block_hash. This hasn't impacted the working of the system I tested because I'm not using views right now. But thats a good catch.

The medium (1) is a really good review, and it uncovers a real bug in our code related to non-deterministic bead creation. This is not related to this PR; the bug lies in the consensus code itself. To explain more clearly on this. The Bead struct has a `parents` field inside `CommittedMetadata` and `parent_bead_timestamps` field as well. The problem here is that these parents are just `HashSet` containers which is not deterministic, and secondly, there is no relation between parents as their timestamps. So each unique run may produce different parent-to-timestamp ordering, leading to non-deterministic behaviour of the same bead. The bug lies in the normal Braidpool mode inside [[lib.rs](http://lib.rs/)](https://github.com/braidpool/braidpool/blob/df09cc6e5f68f479a7034b720f19ab89fb4b994b/node/src/lib.rs#L303), here the `parent_hash_set` is itself `HashSet` (non-deterministic) and then later the `time_hash_set` is Vec but these two later on contain no direct mapping. After this, the code constructs the bead as [weak_share](https://github.com/braidpool/braidpool/blob/df09cc6e5f68f479a7034b720f19ab89fb4b994b/node/src/lib.rs#L359) and uses this locally for storing and all, but at the time for propagating, it uses [serialize](https://github.com/braidpool/braidpool/blob/df09cc6e5f68f479a7034b720f19ab89fb4b994b/node/src/lib.rs#L407) function which internally sorts the parents using `hashset_to_vec_deterministic` for parents but leaves the timestamp as it is, leading to non-deterministic behaviour. Now the receiving node might contain a different serialized bead because of the parents and timestamp. One fix for this is to establish a deterministic relation between parents and their timestamps.

zaidmstrr:
This PR fixes the above mentioned problem, which is also opened in issue [#426](https://github.com/braidpool/braidpool/issues/426).

PR #402, opened 2026-03-05
removing forked dependency for updated version

### LINKED ISSUE #396: Removing customized `rust-bitcoin` dependency

Sansh2356:
- It alters the current functionalities on the basis of forked dependency for `rust-bitcoin`  to follow the upstream tagged version of `0.32.5` instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -383,15 +383,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "base58ck"
-version = "0.2.0"
-source = "git+https://github.com/braidpool/rust-bitcoin.git?branch=cpunet#156f317b6a90d2ddfb9c7b74ac8ff36b705c52b8"
-dependencies = [
- "bitcoin-internals 0.4.0",
- "bitcoin_hashes 0.16.0 (git+https://github.com/braidpool/rust-bitcoin.git?branch=cpunet)",
-]
-
-[[package]]
 name = "base64"
 version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -417,11 +408,6 @@ checksum = "2af50177e190e07a26ab74f8b1efbfe2ef87da2116221318cb1c2e82baf7de06"
 
 [[package]]
 name = "bech32"
-version = "0.11.0"
-source = "git+https://github.com/braidpool/rust-bech32.git?branch=cpunet#34d8ea70211b4178b6b5c540e17981719faccdd2"
-
-[[package]]
-name = "bech32"
 version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32637268377fc7b10a8c6d51de3e7fba1ce5dd371a96e342b34e6078db558e7f"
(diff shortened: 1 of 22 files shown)


mcelrath (changes requested):
● Rust Review: removing forked dependency for updated version

  Summary

  API updates for rust-bitcoin 0.32.5 are mostly consistent, but there are reliability and correctness regressions in
  template parsing and an unexpected default-network change.

  Findings

  🟠 High

   - build_complete_block now panics on malformed templates instead of returning a CoinbaseError, which can crash the node 
  on bad/partial input.
   - FinalTemplate::block_transaction_count now decodes a fixed-width u64 instead of a CompactSize varint, so most valid 
  templates will read as 0 and trigger misleading warnings.

  🟡 Medium

   - Default network changed from cpunet to regtest while CLI/help text still calls cpunet “preferred.” This silently 
  changes runtime behavior and likely breaks the expected default workflow.

  🟢 Suggestions

   - Remove the accidental garbage string in the IBD persistence log line (and consider restoring the log level to debug).

  Code Samples

```
   // node/src/template_creator.rs
   let (tx_count, varint_size) = deserialize_partial::<VarInt>(&original_block_hex[cursor..])
       .expect("An error occurred while decoding VarInt representation");

   // node/src/template_creator.rs
   match deserialize_partial(body) { // infers u64, not VarInt
       Ok((count, _)) => count,
       Err(_) => 0,
   }

   // node/src/main.rs
   warn!(beadhash=?curr_beadhash,"Bead received in IBD persisted ... 
  sanmdfiasnASBNUIYFBNASUIYBNFYUI*ASBNFGYUIASHBNGYUIASHBNGYUIAGHBN");
```

mcelrath:
● Rust Review: PR 402

  Summary

  Review complete. Two high‑severity regressions in block parsing/assembly plus a couple of medium‑risk config/logging
  issues.

  Findings

  🟠 High

   1. Wrong varint decoding in block_transaction_count.
  deserialize_partial is inferred as u64, so it decodes a fixed 8‑byte integer, not a Bitcoin varint. This returns incorrect 
  tx counts (often zero) and breaks downstream logic.
  File: node/src/template_creator.rs
   2. Panic on malformed templates in build_complete_block.
  deserialize_partial::<VarInt>(...).expect(...) will panic on invalid or truncated templates, turning a recoverable error 
  into a node crash. This should return CoinbaseError.
  File: node/src/template_creator.rs

  🟡 Medium

   1. Default config now targets regtest while comments say cpunet preferred.
  This changes out‑of‑box behavior and conflicts with CLI/help text.
  File: node/src/default_braidpool_config.toml
   2. Garbage string in warning log.
  The log message includes random text and should be removed for operational clarity.
  File: node/src/main.rs

  Code Samples
```
   // Current (wrong: infers u64, not VarInt)
   match deserialize_partial(body) {
       Ok((count, _)) => count,
       Err(_) => 0,
   }
   
   // Suggested
   match deserialize_partial::<VarInt>(body) {
       Ok((count, _)) => count.0,
       Err(_) => 0,
   }

   // Current (panic)
   let (tx_count, varint_size) =
       deserialize_partial::<VarInt>(&original_block_hex[cursor..])
           .expect("An error occurred while decoding VarInt representation");
   
   // Suggested (propagate error)
   let (tx_count, varint_size) =
       deserialize_partial::<VarInt>(&original_block_hex[cursor..])
           .map_err(|_| CoinbaseError::InvalidBlockTemplateData)?;
```

PR #403, opened 2026-03-05
test: add comprehensive unit tests for audit proxy

zaidmstrr:
This adds unit tests for the audit and upstream pool.

### DIFF
--- a/node/src/audit.rs
+++ b/node/src/audit.rs
@@ -31,7 +31,7 @@ pub fn compute_audit_bead_hash(bead: &Bead) -> BlockHash {
     BlockHash::from_byte_array(sha256d::Hash::hash(&combined).to_byte_array())
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct AuditCommitment {
     pub commitment_bytes: [u8; COMMITMENT_BYTES],
     pub parent_bead_hash: Option<BlockHash>,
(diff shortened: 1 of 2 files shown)


mcelrath:
● Rust Review: test: add comprehensive unit tests for audit proxy

  Summary

  Test coverage is materially improved, but a production logic change in audit.rs weakens audit correctness by accepting
  stale commitments as valid.

  Findings

  🟠 High

  Accepting the previous commitment as Valid in verify_share_with_fallback allows stale shares to pass audit
  verification. This undermines the audit chain’s guarantee that miners are committing to the current DAG state. If this
  behavior is intentional, it needs explicit bounds (e.g., only when a commitment update is pending and within a short
  window) and should be recorded as a degraded verification state rather than fully valid.

  Code Samples
```
   // node/src/audit.rs
   if let Some(prev_commitment) = previous_commitment {
       if prev_commitment.verify_in_extranonce1(extranonce1_bytes, &self.miner_prefix) {
           warn!(... "Share used previous commitment, accepting it as valid under conditions");
           return AuditVerificationResult::Valid { ... };
       }
   }
```

  Suggested fix: Treat this as invalid (as before) or downgrade it to a distinct status (e.g., Stale) and ensure 
  add_and_record_bead doesn’t mark it as fully verified unless explicitly allowed by policy.

mcelrath (changes requested):
Please see the last comment with the AI review.

zaidmstrr:
This was already fixed in the PR#403 under commit [feat: improvements and new get_bead_count db method](https://github.com/braidpool/braidpool/pull/399/changes/3db925bba305d790ece38a50c91cbc44d5f94d5e)

Issue #404, opened 2026-03-06
Reduce lock contention on stratum_connection_mapping by using RwLock


Shivrajsoni:
Mentioned in https://github.com/braidpool/braidpool/pull/333#discussion_r2897248570

Currently `stratum_connection_mapping` is protected with a Mutex and is locked on each RPC call such as `get_miner_info`. This can block critical Stratum tasks (e.g., miner connect/disconnect or message dispatch) if RPC calls are frequent.

Using an `RwLock` could allow concurrent read access for RPC calls while still maintaining exclusive write access for state updates.

This issue tracks refactoring the locking strategy for `stratum_connection_mapping` to improve concurrency.

Rustix69:
@Shivrajsoni Will you work on it or I can pick it up ?

Shivrajsoni:
Thanks for asking! I'm currently working on this issue.

Sansh2356:
Closing as complete under PR #405 .

PR #405, opened 2026-03-07
stratum: Use RwLock for connection_mapping to avoid blocking RPC

Shivrajsoni:
Issue : #404 

## Changes

Wrap `connection_mapping` in `Arc<RwLock<ConnectionMapping>>` so that  
`getminerinfo` and the `Notifier` use **read locks**, preventing Stratum from being blocked when miners are added/removed or when job notifications are sent.  

Stratum **add/remove operations use write locks**, keeping the critical sections short.

### Updates

- **main.rs**
  - Create `connection_mapping` using `RwLock`.

- **stratum.rs**
  - `Server` and `Notifier` now use **read/write locks**.
  - Tests updated accordingly.

- **rpc_server.rs**
  - `get_miner_info` now uses `read().await`.
  - `run_rpc_server` and tests updated to use `RwLock` for the mapping.


mcelrath (approved):
● Rust Review: PR #405

  Summary

  This is a safe, focused change that reduces contention on the Stratum connection map by switching to RwLock and updating
  read/write sites accordingly.

  Findings

  🟢 Suggestions

   - No correctness or safety issues found in this diff. The read locks are not held across awaits, and write locks are 
  limited to add/remove paths, so the change should improve concurrency without introducing deadlocks.

PR #424, opened 2026-04-04
feat: adding conditional compilation according to OS archs and build.rs

### LINKED ISSUE #421: Cross compatibility across MacOS,linux and Windows
- Currently we have tested over WSL and linux only and for better accessibility and reach we require cross compatibility hence configuration of current codebase according to OS architecture is required .

Sansh2356:
- Solving issue #421 marginally since we don't have released binary yet but cross-compilation is checked for Windows,MacOS and linux respectively .
- Added compatibility for MacOS and linux, windows will require a different implementation in case of `ipc.rs` due to the use of  `NamedPipes` instead of `UnixStream` being available over `tokio::net` for IPC .


mcelrath (changes requested):
● Rust Review: PR #424

  Summary

  Cross‑platform intent is good, but there are build‑breaking issues on non‑Linux/macOS targets and the build script now
  emits invalid Cargo directives.

  Findings

  🟠 High

   1. Build script emits invalid cargo:: directives.
  Cargo only recognizes lines starting with cargo:. Using cargo::rerun-if-changed, cargo::warning, and cargo::error means 
  rebuild triggers and error/warning reporting are ignored. That can leave stale Cap’n Proto bindings and hide compile 
  failures.
  File: node/build.rs
   2. Windows/other targets won’t compile due to missing defaults.
  DEFAULT_DATADIR, DEFAULT_RPC_COOKIE, DEFAULT_IPC_SOCKET, and get_data_dir() are only defined for linux/macos. On Windows 
  (or any other target), these symbols are undefined and get_data_dir() has no body.
  Files: node/src/cli.rs, node/src/db/init_db.rs

  Code Samples

```rust

   // build.rs (should be `cargo:`)
   println!("cargo::rerun-if-changed=schema");
   println!("cargo::warning=Compiling schema file: {}", path.display());

   // cli.rs (no defaults for windows/other targets)
   #[arg(long, default_value = DEFAULT_DATADIR)]
   pub datadir: PathBuf;

   // init_db.rs (no windows/other cfg)
   fn get_data_dir() -> Result<PathBuf, DBErrors> {
       #[cfg(target_os = "linux")] { ... }
       #[cfg(target_os = "macos")] { ... }
   }
```

Sansh2356:
This PR is aimed at MacOS and linux based architectures only as mentioned in PR description the windows compatibility will require changes in `ipc.rs` interface .

zaidmstrr:
The Bitcoin IPC binaries or in general the IPC infrastructure, are not stable on windows. So its not a good idea to give time for that for now. We should focus more on Linux and Mac binaries first.

mcelrath:
@Sansh2356 could you please:
1. fix build conflicts
2. Add CI hook in .github to test compilation and run tests on macos

Suggested scope of this PR: 
1. set up config directories cross-platform and make them available
2. ensure compilation on macos (not windows) using github CI. 

We will need a MacOS collaborator to fully validate.

mcelrath (dismissed):
Approved once merge conflicts are resolved.

node/src/db/init_db.rs
```diff
@@ -1,23 +1,38 @@
...
-use std::{env, fs, path::Path, str::FromStr};
+use std::{env, fs, path::Path, path::PathBuf, str::FromStr};
 
 use crate::error::DBErrors;
 #[allow(unused_imports)]
 use tracing::{debug, error, info, trace, warn};
 static SCHEMA_SQL: &str = include_str!("schema.sql");
 
+/// Gets the braidpool data directory in a cross-platform manner.
+fn get_data_dir() -> Result<PathBuf, DBErrors> {
```
zaidmstrr, line 10: This will work fine in both Linux and Mac, but if a user tries to run the binary in any other platfrom the compiler will throw the type error with no clear reason explaining the problem. I think it would be best if you introduced a fallback block where it just returns an error to the user stating that the platform they are using is not supported yet.

```diff
+ #[cfg(not(any(target_os = "linux", target_os = "macos")))]
+ {
+   Err(DBErrors::EnvVariableNotFetched {
+       error: "unsupported OS".to_string(),
+        var: "OS".to_string(),
+   })
+ }
```

node/src/cli.rs
```diff
@@ -2,11 +2,35 @@ use clap::Parser;
...
+#[cfg(target_os = "linux")]
+const DEFAULT_RPC_COOKIE: &str = "~/.bitcoin/.cookie";
+
+/// Default Bitcoin RPC cookie path for macOS
+#[cfg(target_os = "macos")]
+const DEFAULT_RPC_COOKIE: &str = "~/Library/Application Support/Bitcoin/.cookie";
+
+/// Default IPC socket path for Linux (Unix domain socket)
+#[cfg(target_os = "linux")]
+const DEFAULT_IPC_SOCKET: &str = "/tmp/bitcoin-cpunet.sock";
```
zaidmstrr: Both of these are the same for Linux and Mac. I think merging them together is better here.

node/src/db/init_db.rs
```diff
@@ -1,23 +1,38 @@
...
+    #[cfg(target_os = "linux")]
+    {
+        let home = env::var("HOME").map_err(|error| DBErrors::EnvVariableNotFetched {
+            error: error.to_string(),
+            var: "{HOME} Directory".to_string(),
```
zaidmstrr: The `{HOME}` inside the error are not format placeholders. It would be good if replaced with some other meaningful message.

node/src/db/init_db.rs
```diff
@@ -1,23 +1,38 @@
...
+        })?;
+        Ok(Path::new(&home).join(".braidpool"))
+    }
+
+    #[cfg(target_os = "macos")]
+    {
+        let home = env::var("HOME").map_err(|error| DBErrors::EnvVariableNotFetched {
+            error: error.to_string(),
+            var: "{HOME} Directory".to_string(),
```
zaidmstrr: Same for here.

node/build.rs
```diff
@@ -1,48 +1,112 @@
...
+    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap_or_default();
+    let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap_or_default();
+
+    let platform_cfg = match target_os.as_str() {
+        "macos" => "macos",
+        "linux" => "linux",
+        _ => "other",
+    };
+
+    println!("cargo::rustc-cfg=target_platform=\"{platform_cfg}\"");
```
zaidmstrr: These are declared and not used anywhere else. Secondly the `rustc-check-cfg` is not initialized so any code using this will emit a warning by the compiler. The best thing is to remove this for now until explicitly needed by some program.

node/build.rs
```diff
@@ -1,48 +1,112 @@
...
+    let arch_cfg = match target_arch.as_str() {
+        "x86_64" => "x86_64",
+        "aarch64" => "aarch64",
+        "x86" => "x86",
+        _ => "other",
+    };
+    println!("cargo::rustc-cfg=target_cpu_arch=\"{arch_cfg}\"");
```
zaidmstrr: Same for this.

zaidmstrr (commented):
Concept ACK [0e297b2](https://github.com/braidpool/braidpool/pull/424/changes/0e297b20891b6260ea362f8229c82e3ca44ae698)

Approach seems fine but found some minor room for improvement. Please take a look at them.

Sansh2356:
@zaidmstrr i have updated the requested changes .

zaidmstrr (approved):
Code ACK [05e086b](https://github.com/braidpool/braidpool/pull/424/changes/05e086b8b4d89a6c604146e204c803bfc14317d1)

Thanks, looks good to me now.

Issue #318, opened 2025-12-19
Raise a Big Red Flag if we can't talk to Bitcoin


mcelrath:
In #297 I think I originally noticed this due to logging spam. But this queue being full and dropping requests is actually a BIG PROBLEM. This relates to #283 which is pending work. When that is complete we will be putting txs into cmempoold, computing block templates, comparing the Merkle root to the one in the bead we just receiveed, and then removing some txs again. Therefore, if our queue to Bitcoin (cmempoold) ever gets so full that requests are dropped, it means that our node will fall out of consensus with the Braid, because this check is part of our consensus algorithm.

Therefore, if this ever happens, we need to halt and crash and make some really noisy log messages.

We should also ensure that this can never happen, one way or another.

krishi-agrawal:
Are these the required changes for PR #316 ?

1. Critical and High queues should be unbounded to prevent queue-full conditions that would break consensus
2. Normal and Low queues should remain bounded 
3. Adding explicit error reporting and logging when bounded queues do fill (replacing the silent drop behavior)

Issue #283, opened 2025-11-05
cmempoold split


mcelrath:
We're currently using only a single bitcoin node to talk to. In order to complete the #125 we need to split this into two bitcoin nodes, one of which (cmempoold) is used for tracking transactions added in beads to the committed mempool, and the other is a standard bitcoin node from which we take mempool transactions and submit blocks.

#270 adds a template cache, which we can use largely as-is and the template cache doesn't necessarily have to care which kind of template it is. I might suggest using a `HashMap<BlockHash,BlockTemplate>` but I leave this as an implementation detail for now.

A detailed description of the logic is:

[Original Discussion](https://github.com/braidpool/braidpool/pull/270) Each bead we generate changes the block template (if it carries any transactions). However we can pre-compute this for beads we mine, to keep our hashers hashing and get our txs in beads. Therefore let's assume we keep:
- `current_bead `: (what we've sent to miners)
- `next_bead`: (bead with a block template including txs from `current_bead`, and `current_bead` as the parent)
- `next_next_bead`: (bead with a block template including txs from `next_bead` and `current_bead`, and `next_bead` as the parent)
- ... more if necessary?
- `next_block_bead`: (bead that is mined next assuming `current_bead` becomes a block)

As far as block templates this means we need:
- `current_block_template`: what we're hashing on. (`current_block_template` = `current_bead_template`)
- `next_bead_template`: next block template assuming **WE** mine `current_bead` as a **BEAD**:
    - Add txs from `current_bead` to cmempoold
    - Call `getblocktemplate` to generate `next_bead_blocktemplate`
    - Generate `next_bead`
    - Remove all txs from `current_bead` from cmempoold
- `next_block_template`: the block template we will mine next assuming **WE** mine `current_block_template` as a **BLOCK** used by `next_block_bead`:
    - Remove all txs from `current_bead` and `current_blocktemplate` from cmempoold
    - Call `getblocktemplate` to generate `next_blocktemplate`
    - Add in all txs from `current_block_template` back in to cmempoold

Work-change scenarios:
1. Successful bead (not block) in mining.submit (~few ms hashrate loss for 1 device)
    - Send cached `next_bead` in `mining.notify` to all devices.
    - `current_bead` = `next_bead`
    - Call `getblocktemplate` to generate `next_bead` (~200ms) and `next_next_bead`
2. Successful bead `incoming_bead` (not a block) from other nodes (no hashrate loss - latency generates siblings)
[...]

PR #436, opened 2026-04-15
ipc: Fail-stop the node when Bitcoin IPC queue overflows

pradhyum6144:
Issue : #318

## Changes

The existing `panic!` in `PriorityRequestQueue::enqueue` never halted the node: it runs inside `tokio::task::spawn_local`, where panics are caught by Tokio and only surface via `JoinHandle::await`. The processor task died silently while callers blocked on `oneshot` channels, leaving the node alive but unable to satisfy the consensus-critical IPC required by `docs/braidpool_spec.md`.

Replace the silent panic with `std::process::abort()` on Normal/High/Critical overflow so operators see the failure. Low-priority overflow stays a soft drop (warn + continue) since it is non-consensus.

### Updates

- **ipc/client.rs**
  - New `EnqueueOutcome` enum returned by `enqueue`, matched by the processor task: Low → warn + drop, Normal/High/Critical → error + `std::process::abort()`.
  - `is_overloaded()` now checks all four priority queues.
  - 7 new unit tests covering each outcome and priority ordering.

### DIFF
--- a/node/src/ipc/client.rs
+++ b/node/src/ipc/client.rs
@@ -191,8 +191,8 @@ impl PriorityRequestQueue {
         let result = match priority {
             RequestPriority::Critical => {
                 if self.critical_queue.len() >= self.max_queue_sizes.critical {
-                    error!("BITCOIN IPC QUEUE FULL - CRITICAL QUEUE EXCEEDED LIMIT");
-                    panic!("FATAL: Cannot communicate with Bitcoin Core. Queue is full.");
+                    error!("FATAL: Bitcoin IPC critical queue full - aborting to prevent consensus desync. See issue #318.");
+                    std::process::abort();
                 } else {
                     self.critical_queue.push_back(request);
                     self.metrics
@@ -203,8 +203,8 @@ impl PriorityRequestQueue {
             }
             RequestPriority::High => {
                 if self.high_queue.len() >= self.max_queue_sizes.high {
-                    error!("BITCOIN IPC QUEUE FULL - HIGH PRIORITY QUEUE EXCEEDED LIMIT");
-                    panic!("FATAL: Cannot communicate with Bitcoin Core. Queue is full.");
+                    error!("FATAL: Bitcoin IPC high queue full - aborting to prevent consensus desync. See issue #318.");
+                    std::process::abort();
                 } else {
                     self.high_queue.push_back(request);
                     self.metrics
@@ -215,8 +215,8 @@ impl PriorityRequestQueue {
             }
             RequestPriority::Normal => {
                 if self.normal_queue.len() >= self.max_queue_sizes.normal {
-                    error!("BITCOIN IPC QUEUE FULL - NORMAL PRIORITY QUEUE EXCEEDED LIMIT");
-                    panic!("FATAL: Cannot communicate with Bitcoin Core. Queue is full.");
+                    error!("FATAL: Bitcoin IPC normal queue full - aborting to prevent consensus desync. See issue #318.");
+                    std::process::abort();
                 } else {
                     self.normal_queue.push_back(request);
                     self.metrics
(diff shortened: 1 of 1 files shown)


pradhyum6144:
Flagging before review: #318 has two asks, this PR only covers one.

Covered (requirement 1- "halt + noisy logs"): Normal/High/Critical overflow now emits error! + std::process::abort(). Note: the previous panic! never actually halted the node  it ran inside tokio::task::spawn_local, so Tokio caught it and the processor task died silently while callers blocked on oneshot channels. abort() bypasses that.

Not covered (requirement 2 - "ensure this can never happen"): Scoped out; I believe it's blocked on #283 (cmempoold split) + p2p rate limiting.

Direction question: @krishi-agrawal suggested unbounding Critical + High queues so overflow is unreachable - a different philosophy than this PR (bounded + fail-stop). Happy to pivot:

(a) Land as-is, track (2) against #283.
(b) Also unbound Critical + High here, keep abort() as safety net on Normal/Low.
(c) Unbound Critical + High, drop abort() entirely (krishi's proposal).
Which would you prefer?

zaidmstrr (changes requested):
Concept ACK [db6e8f8](https://github.com/braidpool/braidpool/pull/436/changes/db6e8f88a4ccbd78fbb398118409c768a3d10aab)

I don't like the idea of introducing `EnqueueOutcome` enum this is making code complex without adding much meaning, even though the test verifies overflow detection but not the abort response, so you end up testing the enum shape rather than the behaviour. Although your concern about the `abort()` is correct. The cleaner and simpler approach here is to use `abort()` inside `enqueue()` and don't treat low priority requests differently than critical and high. Because if a queue is full on low priority, then still that means the Bitcoin node is unresponsive.

node/src/ipc/client.rs
```diff
@@ -187,63 +195,52 @@ impl PriorityRequestQueue {
             BitcoinRequest::GetMiningTipInfo { priority, .. } => *priority,
             BitcoinRequest::SubmitSolution { priority, .. } => *priority,
         };
-
-        let result = match priority {
```
zaidmstrr, line 191: Removing this piece of code is not the intended fix for this issue. I suggest just introducing a simple `abort()` and leaving the old code unchanged.

zaidmstrr (commented):
Thanks for working on the things I suggested previously. This now looks way better than before, but there are some more suggestions below you can look at.

node/src/ipc/client.rs
```diff
@@ -191,8 +191,8 @@ impl PriorityRequestQueue {
         let result = match priority {
             RequestPriority::Critical => {
                 if self.critical_queue.len() >= self.max_queue_sizes.critical {
-                    error!("BITCOIN IPC QUEUE FULL - CRITICAL QUEUE EXCEEDED LIMIT");
-                    panic!("FATAL: Cannot communicate with Bitcoin Core. Queue is full.");
+                    error!("FATAL: Bitcoin IPC critical queue full - aborting to prevent consensus desync. See issue #318.");
```
zaidmstrr, line 194: We don't need to change the previous logs please leave them as they are. And sign your commits.

node/src/ipc/client.rs
```diff
@@ -215,8 +215,8 @@ impl PriorityRequestQueue {
             }
             RequestPriority::Normal => {
                 if self.normal_queue.len() >= self.max_queue_sizes.normal {
-                    error!("BITCOIN IPC QUEUE FULL - NORMAL PRIORITY QUEUE EXCEEDED LIMIT");
-                    panic!("FATAL: Cannot communicate with Bitcoin Core. Queue is full.");
+                    error!("FATAL: Bitcoin IPC normal queue full - aborting to prevent consensus desync. See issue #318.");
```
Sansh2356, line 218: Why is the issue number listed in running logs not required and imo we can revert to previous log messages only .

Sansh2356 (changes requested):

PR #435, opened 2026-04-15
Adding CLI option to set stratum port

kushagra0902:
## Summary
Current implementation doesn't have any option to set the stratum port from the CLI. This is especially needed when testing and running two nodes simultaneously. Further, the RPC bind port was hardcoded in the main file and was marked as a "TODO".

This PR solves the issue by getting both of these from the CLI args.

## Usage:
``cargo run -- --addnode=/ip4/127.0.0.1/udp/8000/quic-v1 --bind=127.0.0.1:9000 --stratum-port=3334 --rpc-bind=127.0.0.1:6683``

Added tage: 
- ``stratum-port``
- `rps-bind`

PR #356, opened 2026-01-27
feat: implement config file loading with CLI override priority

parthdude07:
#327 
- Update `node/src/main.rs` to parse the TOML config file specified by the `--config` flag.
- Implement merging strategy: CLI arguments take precedence over config file values, falling back to defaults if neither are present.
- Ensure `BraidpoolConfig` is actively used to drive node settings (RPC, network, bind address).
- Enable active usage of `default_braidpool_config.tomlfeat: implement config file loading with CLI override priority

- Update `node/src/main.rs` to parse the TOML config file specified by the `--config` flag.
- Implement merging strategy: CLI arguments take precedence over config file values, falling back to defaults if neither are present.
- Ensure `BraidpoolConfig` is actively used to drive node settings (RPC, network, bind address).
- Enable active usage of `default_braidpool_config.toml

PR #437, opened 2026-04-15
node: honor braid_rpc_config.rpc_server_addr at startup

### LINKED ISSUE #431: node: startup ignores braid_rpc_config.rpc_server_addr and always binds RPC to 127.0.0.1:6682
### Summary
The node already exposes an RPC bind address through config, but startup currently ignores that value and always uses a hardcoded address.

In `node/src/main.rs`, the RPC server is started with a hardcoded `127.0.0.1:6682`:

```rust
let rpc_addr = "127.0.0.1:6682"; // TODO: Load from config file
```

However, the config layer already defines this setting in `node/src/config.rs`:

```rust
pub struct BraidRpcConfig {
    pub rpc_server_addr: String,
}
```

and the default config file also includes it in `node/src/default_braidpool_config.toml`:

```toml
[braid_rpc_config]
rpc_server_addr = "127.0.0.1:6682"
```

### Expected behavior
If `braid_rpc_config.rpc_server_addr` is set in config, the RPC server should bind to that value.

### Actual behavior
The node always binds RPC to `127.0.0.1:6682` regardless of the configured value.

### Why this matters
This makes the RPC configuration misleading and prevents users from changing the RPC bind address through the existing config path.

### Possible fix
Use `braid_rpc_config.rpc_server_addr` when starting the RPC server instead of the hardcoded address in `node/src/main.rs`.

Shreyas2004wagh:
## Summary
Closes #431.

The node already exposes `braid_rpc_config.rpc_server_addr` in the config layer and in `default_braidpool_config.toml`, but startup still hardcodes the RPC bind address in `main.rs`.

This PR makes startup honor `braid_rpc_config.rpc_server_addr` by loading node config from `~/.braidpool/braidpool-config.toml` when present and otherwise falling back to the existing default config.

## Changes
- load `BraidpoolConfig` from the node datadir during startup
- use `braid_rpc_config.rpc_server_addr` when starting the RPC server
- add tests covering datadir config loading and fallback to default config
- document the runtime config file location in `docs/braidpool_setup.md`

## Notes
There is related upstream work in:
- `#356`, which is a broader config-loading / CLI precedence PR
- `#435`, which adds CLI flags for RPC bind and stratum port

This PR is intentionally narrower and only addresses the config-backed RPC bind bug reported in `#431`.

## Testing
I added targeted config-loading tests, but I could not run `cargo test` or `cargo fmt` through Cargo locally because the current local Cargo toolchain (`1.81.0`) fails to parse this workspace with an `edition2024` manifest error.

### DIFF
--- a/docs/braidpool_setup.md
+++ b/docs/braidpool_setup.md
@@ -60,7 +60,7 @@ You can find all available command-line arguments in [node/src/cli.rs](https://g
 -   `--ipc-socket <PATH>`: Specifies the path to the UNIX domain socket file (should be the same as bitcoin node).
 -   `--network <NETWORK>`: Sets the network. Valid options are `mainnet`, `testnet4`, `signet`, and `cpunet`. The default is `mainnet`.
 
-If `~/.braidpool/braidpool-config.toml` exists, the node will also load `braid_rpc_config.rpc_server_addr` from that config file when binding the JSON-RPC server.
+If `<datadir>/braidpool-config.toml` exists, the node will also load `braid_rpc_config.rpc_server_addr` from that config file when binding the JSON-RPC server. Here, `<datadir>` is the directory passed via `--datadir` or the default data directory if `--datadir` is not provided. The file must be a complete, valid `BraidpoolConfig` TOML document, not just a `[braid_rpc_config]` snippet.
 
 ## Setting up cpuminer for downstream connection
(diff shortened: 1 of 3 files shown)


zaidmstrr (commented):
Closing as this is the duplicate of [PR#356](https://github.com/braidpool/braidpool/pull/356)

PR #440, opened 2026-04-18
Adding msrv and deps bump

Sansh2356:
- This adds MSRV to clip the tool-chain version which satisfies our dependency tree which i have found to be `1.88.0` and updating the older/archived actions-toolchains.
- Updates the metadata of root package and propagating rustc min version to workspace members.
- Updating the lockfile for stale deps no breaking changes .


.github/workflows/rust-node.yml
```diff
@@ -12,17 +12,36 @@ jobs:
...
-      - uses: actions-rs/toolchain@v1
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # tag=v6.0.2
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@0b1efabc08b657293548b77fb76cc02d26091c7e
         with:
-          profile: minimal
           toolchain: stable
-          override: true
+      - uses: Swatinem/rust-cache@v2
       - uses: actions-rs/cargo@v1
```
zaidmstrr: `actions-rs` is deprecated I guess. It would be good if you just use normal `run` here. Something similar to:

```diff
- uses: actions-rs/cargo@v1
- with:
-    command: check
-    args: --manifest-path Cargo.toml
+ run: cargo check --manifest-path Cargo.toml
```

.github/workflows/rust-node.yml
```diff
@@ -12,17 +12,36 @@ jobs:
...
-      - uses: actions/checkout@v2
-      - uses: actions-rs/toolchain@v1
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # tag=v6.0.2
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@0b1efabc08b657293548b77fb76cc02d26091c7e
         with:
-          profile: minimal
           toolchain: stable
-          override: true
+      - uses: Swatinem/rust-cache@v2
```
zaidmstrr: This should also be pinned to a commit hash like others.

.github/workflows/rust-node.yml
```diff
@@ -12,17 +12,36 @@ jobs:
...
+      - name: Install Cap'n Proto
+        run: |
+          export DEBIAN_FRONTEND=noninteractive
+          sudo apt-get install -y capnproto libcapnp-dev
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # tag=v6.0.2
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@0b1efabc08b657293548b77fb76cc02d26091c7e
+        with:
+          toolchain: "1.88"
+      - uses: Swatinem/rust-cache@v2
```
zaidmstrr: This also needs to be pinned to a commit hash.

.github/workflows/rust-node.yml
```diff
@@ -31,12 +50,12 @@ jobs:
...
-      - uses: actions/checkout@v2
-      - uses: actions-rs/toolchain@v1
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # tag=v6.0.2
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@0b1efabc08b657293548b77fb76cc02d26091c7e
         with:
-          profile: minimal
           toolchain: stable
-          override: true
+      - uses: Swatinem/rust-cache@v2
```
zaidmstrr: This one also here.

.github/workflows/rust-node.yml
```diff
@@ -45,31 +64,12 @@ jobs:
...
-  #         profile: minimal
-  #         toolchain: stable
-  #         override: true
-  #     - run: rustup component add clippy
-  #     - uses: actions-rs/cargo@v1
-  #       with:
-  #         command: clippy
-  #         args: -- -D warnings
+          components: rustfmt
+      - uses: Swatinem/rust-cache@v2
```
zaidmstrr: Using `rust-cache` here is a resource wastage. Because fmt doesn't need to compile the code.

.github/workflows/rust-node.yml
```diff
@@ -31,12 +50,12 @@ jobs:
...
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@0b1efabc08b657293548b77fb76cc02d26091c7e
         with:
-          profile: minimal
           toolchain: stable
-          override: true
+      - uses: Swatinem/rust-cache@v2
       - name: Run tests
         run: |
           cargo test
```
zaidmstrr: This runs on the implicit behaviour. What do you think if we add `--workspace` here?

zaidmstrr (commented):
Concept ACK [245da44](https://github.com/braidpool/braidpool/pull/440/changes/245da44e8b60aaec4926320ed4b60acb83568b3c)

Overall the intent of the PR is reasonable. I have suggested a few things which can be improved.

mcelrath (approved):

PR #444, opened 2026-04-20
removing the lock contention by taking out ownership and dropping gua…

### LINKED ISSUE #443: Reducing lock contentions
- Currently there are certain instances especially in the swarm_loop and stratum components where the locks are held longer then they are required to and hence can be dropped earlier in the spawned threads thus reducing the data races and contention for the `MutexGuard` for braid_lock thus leading to reduced CPU utilization as well as unnecessary holding of locks .
- It also includes holding locks across some `Futures` which can be improved as well.
- I have done some profiling locally with `perf` and `cargo flamegraph` to check potential allocations as well as contentions for cache misses at L1,L2 and L3 level in order to improve to cache-locality and for better performance and on the basis of that these smaller changes can certainly improve this contention reduce IO waited threads .

Sansh2356:
Addresses #443 partially .


node/src/braid/mod.rs
zaidmstrr, line 34: If I'm not wrong this change was also a part of the Orphan [PR #309](https://github.com/braidpool/braidpool/pull/309). Then why its here in this PR also?

zaidmstrr (commented):
The title of the PR says it's for removing the lock contenion, so are you sure that this PR doesn't change anything else besides that? If yes, please tell.

node/src/braid/mod.rs
Sansh2356: I had updated the #309 after cherry-picking my commit so a rebase was pending here .

Sansh2356:
> The title of the PR says it's for removing the lock contenion, so are you sure that this PR doesn't change anything else besides that? If yes, please tell.

No it is dealing with the scope as mentioned in issue mentioned in PR header .

mcelrath (changes requested):
● The risk is that the new lock‑scope changes in node/src/main.rs (around the floodsub bead handler, ~560–610) compute 
  bead_id, bead_index_mapping, and removed_orphans under the braid write lock, then release the lock before awaiting 
  db_tx.send(InsertBeadSequentially{…}). That send is awaited, so other bead events can run and enqueue DB inserts
  first. Because the DB path is sequential and uses bead_id/removed_orphans snapshots, out‑of‑order inserts can violate
  parent‑before‑child assumptions or apply orphan removals based on a stale braid snapshot.

  Concrete interleaving example:

   1. Bead A arrives; braid is extended; bead_id and removed_orphans are computed; lock is released.
   2. Before db_tx.send(A) completes, bead B arrives and is added; db_tx.send(B) runs first.
   3. DB processes B before A. If it enforces parent existence, B can fail; if it doesn’t, the orphan removal list and 
  index mapping applied to B are no longer consistent with DB state, risking dropped beads or inconsistent orphan 
  tracking.

  That’s why it’s flagged as a correctness risk (high severity) even though it looks like a harmless lock‑scope
  reduction.

Sansh2356:
Closing as following #309 .

Issue #452, opened 2026-04-25
Dashboard: Remove Recharts default gray hover band on Bitcoin 24h price bar chart


Rustix69:
**Problem**  
On **Bitcoin Statistics → “Bitcoin Price Range (24h)”**, hovering a bar showed a **large light gray background** behind the bar in addition to the black tooltip. That band is Recharts’ default **axis tooltip cursor** (`recharts-tooltip-cursor`, fill `#ccc`), not part of our bar styling.

**Expected**  
Only the **tooltip** (label + “Price: …”) should appear on hover; no extra highlight rectangle behind the bar.

**Demo**  

https://github.com/user-attachments/assets/a392f594-88fa-40d2-a5a5-afe70bd3d23d

zaidmstrr:
Closing as the discussion leads to not incorporating the changes.

PR #453, opened 2026-04-25
Hide Recharts bar cursor on Bitcoin 24h price chart

Rustix69:
**Summary**  
- Fixes : #452 
- Stops the default Recharts **gray hover band** (`recharts-tooltip-cursor`) on the **Bitcoin Price Range (24h)** bar chart so only the **tooltip** shows on hover.

**Demo**
image

### DIFF
--- a/dashboard/src/components/BitcoinStats/Prices.tsx
+++ b/dashboard/src/components/BitcoinStats/Prices.tsx
@@ -354,16 +354,16 @@ const BitcoinPriceTracker: React.FC = () => {
                   (dataMin: number) =>
                     Math.floor(
                       dataMin -
-                      (priceData
-                        ? (priceData.high24h - priceData.low24h) * 0.1
-                        : 0)
+                        (priceData
+                          ? (priceData.high24h - priceData.low24h) * 0.1
+                          : 0)
                     ),
                   (dataMax: number) =>
                     Math.ceil(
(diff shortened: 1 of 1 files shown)


priyashuu:
It was actually intentional; the grey background helps indicate the area covered by the individual bar and improves hover feedback. Making it fully transparent might reduce interactivity; we could instead fine-tune the colour for better balance.
What do you think?

Rustix69:
Hey @priyashuu  that gray area is actually a tooltip thing that has a gray back by default if we used shadecn then we hadnt faced this thing. The gray area is actually not the area bcz the height of the gray area is always equal to the height of the full graph so most of the times in apps we prefer the shadecn charts to so that it looks good ... 

And I feel that on hovering the tool tip will show the y value of the bar.

Rustix69:
The same thing is in the mempool section. For value 0 also it will show that grayish section.

[Screencast from 2026-04-28 02-29-19.webm](https://github.com/user-attachments/assets/402565a7-b057-47e5-ab85-96e88aa9cb02)

mcelrath:
I can understand why you wanted this, but this is kind of an artifact of test data where the fee is zero. In practice this will never occur in the real network. 

@priyashuu would prefer a hover animation that highlights the actual data, with a line on the x axis if it is zero, and I agree.

PR #456, opened 2026-04-29
Frontend Migration: Deprecate Simulator API and adopt Braidpool RPC

priyashuu:
### **Overview**

This PR depends on another PR and should be merged after it. It introduces frontend changes to deprecate the Simulator API and switch all data fetching to Braidpool RPC methods. The UI now displays Braidpool data instead of simulator data.

### **Changes**

* Removed/deprecated Simulator API usage
* Updated frontend logic to reflect RPC-based data

### **Dependency**

* Depends on: **#427**

### DIFF
--- a/dashboard/src/URLs.ts
+++ b/dashboard/src/URLs.ts
@@ -8,9 +8,6 @@ export const WEBSOCKET_URLS = {
   // Main WebSocket for general real-time updates (used in multiple components)
   MAIN_WEBSOCKET: 'ws://localhost:5000',
 
-  // WebSocket for BraidPool DAG visualization (simulator API)
-  BRAIDPOOL_DAG_WEBSOCKET: 'ws://localhost:65433/',
-
   // WebSocket for block viewer updates (mempool)
   BLOCK_VIEWER_WEBSOCKET: 'http://localhost:8080/api/v1/ws',
 } as const;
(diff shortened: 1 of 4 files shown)


zaidmstrr:
This has some failing checks. Please fix that.

mcelrath (changes requested):
This is a small fix, but there were some comments from the AI, can you respond to these?

● Below are the expanded details for each PR #456 finding.

  1) Perf (medium): main WS JSON.parse filtering
  Where: dashboard/src/components/BraidPoolDAG/BraidPoolDAG.tsx
  Why it matters: The component now listens to the main, multiplexed WebSocket and parses every incoming message just to
  filter for DAG updates. If that main feed is high‑frequency (multiple message types, large payloads), repeated JSON.parse
  + type checks on the UI thread can create frame drops and input jank, especially when DAG rendering is already busy.
  Risk profile: Doesn’t break correctness, but under heavy traffic it can noticeably degrade UI responsiveness.
  Suggested fix: Prefer server‑side routing to a dedicated DAG channel, or include a lightweight message envelope that
  allows filtering without full parse (e.g., a small prefix/type field), or split the WS feed for DAG updates.

  2) Perf (low): per‑tail Array.from + Math.max
  Where: dashboard/src/components/BraidPoolDAG/BraidPoolDAGUtils.ts
  Why it matters: For each update, the code allocates arrays from Sets and then computes Math.max(...). With large DAGs and
  frequent updates, these allocations add GC pressure.
  Risk profile: Minor overhead; not a correctness issue.
  Suggested fix: Iterate the Set directly (manual max loop) to avoid allocations.

  3) TS (low): console.log in WS handler
  Where: dashboard/src/components/BraidPoolDAG/BraidPoolDAG.tsx
  Why it matters: Leaving logs in production can be noisy and can add small runtime overhead, and can leak data if logs
  include payload content.
  Risk profile: Low.
  Suggested fix: Guard logs behind import.meta.env.DEV or remove entirely.

PR #458, opened 2026-05-02
add cmempool module for committed mempool IPC support #125)

### LINKED ISSUE #125: Committed Mempool
Braidpool with "committed mempool" functionality will consist of three components:

1. **braidpool-node** the main braidpool node
2. **cmempoold** an instance of bitcoin used only for its mempool, communicating over IPC with `braidpool-node`
3. **bitcoind** any standard instance of bitcoin core.

# `cmempoold` tasks:
1. Compile an extremely minimal version of `bitcoind` with:
    a. NO network access, wallet, ZMQ, RPC, etc
    b. Configured to use the same chainstate as `bitcoind`
    c. Identify which versions of bitcoin core have the same mempool policy and can be used for `cmempoold` ([discussion](https://discord.com/channels/1270418429412642957/1270423991068266672/1359150338837581844))
2. Runtime configuration that disables network and RPC access (we will only talk to it with IPC).
3. Each bead effectively has its own mempool, but the majority of the UTXO state is shared with ancestors. Figure out how to either have multiple instances of the Mempool object, or quickly create new ones as needed.

When a new bead is seen by `braidpool-node` it will:
1. Extract the set of transactions carried by the bead.
2. Compute the amount of reserved space necessary to include these transactions
3. Call `cmempoold` to merge the mempool object of all parents into a mempool for this bead:
    a. This might be done by extracting individual transactions that differ and adding them according to the highest work sorting
    b. If transactions cannot be accepted to the mempool, and are not on the highest work path, continue
    c. If transactions cannot be accepted to the mempool, and ARE on the highest work path, the bead is invalid.
5. Call `getblocktemplate` to obtain a block template from the committed mempool created in (3), reserving space for new transactions in (2).
6. Add the transactions carried by this bead to the block template and recompute the Merkle root.
7. Randomize the block template and recompute the Merkle root (see below)

`cmempoold` needs to be notified of new blocks. Figure out how to do this with IPC and have braidpool-node feed new blocks to `cmempoold`. This functionality is part of the standard bitcoin operation, but it should:
1. Remove any transactions from its mempool that appeared in the block.

# Discussion

## Transaction randomization functionality
1. We want to randomize the transactions in the block template, subject to the constraint that the transactions themselves imply (for instance if a transaction spends an output created in the same block, the transaction creating the output must appear *before* the transaction that spends it.
2. Use the "cluster mempool" work to identify clusters of transactions that can be re-ordered as a group.
3. Use a PRNG seeded with the parent block hash and extranonce to randomize the order of txs
4. Recompute the Merkle root

The point of this is that a number of "parasitic asset" protocols (Runes, BRC20, DEX'es, etc) depend upon transaction order, creating MEVil. This is not something anyone should depend upon and Bitcoin provides no guarantees at all about transaction order, UNLESS one transaction spends the output of another, which we must respect. Therefore we will seek to break these stupid and dangerous protocols. We won't otherwise try to stop such transactions (as OCEAN and Knots try to do) because it's fundamentally impossible, and opens a door where someone will ask: "If you can censor Runes/BRC20 transactions, why can't you censor <Ukraine Donations|Canadian Truckers|Political Party X|Mongolian Independence>?" and we don't want to create a censorship engine.

## IPC communication instead of RPC/ZMQ/P2P

We wish to primarily (and exclusively) use IPC functionality, so that the `cmempoold`, `bitcoind` and `braidpool-node` are required to be running on the same host. We want to make it as difficult as possible to proxy this functionality over a network. In other words, we don't want people to outsource transaction selection to another party. On a related note, we're unlikely to use the StratumV2 [Job Declaration Protocol](https://stratumprotocol.org/specification/06-Job-Declaration-Protocol/) or [Template Distribution Protocol](https://stratumprotocol.org/specification/07-template-distribution-protocol/) for this reason, but will use the StratumV2 [Mining Protocol](https://stratumprotocol.org/specification/05-Mining-Protocol/) for talking to mining devices. (Research needed)

BHARATH0153:
fixes #125

### changes 

I read every existing file before writing a single line: main.rs, block_template.rs, zmq.rs, rpc.rs, cli.rs, connection.rs, protocol.rs, and all of braid.rs + braid/io_json.rs. The existing node already fetches block templates from bitcoind via RPC on ZMQ hashblock notifications. The issue asks to insert a cmempoold layer between those steps.

### DIFF
--- a/node/src/block_template.rs
+++ b/node/src/block_template.rs
@@ -55,10 +55,11 @@ pub async fn fetcher(
     }
 }
 
+// dummy placeholder function to consume the received block templates
 pub async fn consumer(mut block_template_rx: Receiver<GetBlockTemplateResult>) {
     let mut last_block_template_height = 0;
     while let Some(block_template) = block_template_rx.recv().await {
-           // if block template is from some outdated exponential backoff RPC, ignore it
+    // if block template is from some outdated exponential backoff RPC, ignore it
         if block_template.height > last_block_template_height {
             log::info!(
                 "Received new block template via `getblocktemplate` RPC: {:?}",
--- a/node/src/block_template.rs
+++ b/node/src/block_template.rs
@@ -42,6 +42,8 @@ pub async fn fetcher(
                 }
                 rpc_failure_backoff = u64::checked_pow(BACKOFF_BASE, rpc_failure_counter.clone())
                     .expect("MAX_RPC_FAILURES doesn't allow overflow; qed");
+                
+                // sleep until it's time to try again
                 log::error!("Error on `getblocktemplate` RPC: {}", e);
                 log::error!(
                     "Exponential Backoff: `getblocktemplate` RPC failed {} times, waiting {} \
--- a/node/src/block_template.rs
+++ b/node/src/block_template.rs
@@ -3,6 +3,8 @@ use bitcoincore_rpc_json::{GetBlockTemplateModes, GetBlockTemplateResult, GetBlo
 use tokio::sync::mpsc::{Receiver, Sender};
 use tokio::time::{sleep, Duration};
 
+use crate::cmempool::IpcClient;
+
 const BLOCK_TEMPLATE_RULES: [GetBlockTemplateRules; 4] = [
     GetBlockTemplateRules::SegWit,
     GetBlockTemplateRules::Signet,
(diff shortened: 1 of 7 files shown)


BHARATH0153:
@priyashuu please check when you are free

Sansh2356:
@BHARATH0153 you read everything but you didn't read that all the work is being done over `dev` and not `main`, and these are clearly untested, un-thinked AI slops please do not do this. This work is serious don't waste the time of reviewers as well as yours, just see at your commit history what is even that ?

Sansh2356:
`NACK` closing this PR continuation of these types of PRs will not get you merge anything please take a note to that .

Issue #463, opened 2026-05-12
Test framework implementation plan


kushagra0902:
# Braidpool Functional Testing Framework (Summer of Bitcoin '26)

## Phase 1: Configuration & Logging
- [ ] **PR 1:** `NetworkConfig and PortPool`
  - [ ] Implementing `NetworkConfig` dataclass with timeouts, ports, and binary paths.
  - [ ] Implementing port allocation leveraging `PYTEST_XDIST_WORKER` with `socket.bind()` validation.
- [ ] **PR 2:** `testing NetworkConfig and PortPool`
  - [ ] Testing `PortPool` race conditions and worker ID port isolation.
- [ ] **PR 3:** `LogManager`
  - [ ] Implementing structured, timestamped test run directory based logging system.
- [ ] **PR 4:** `testing LogManager`
  - [ ] Testing directory teardown logic on success vs. failure.

## Phase 2: RPC & Miner Integration
- [ ] **PR 5:** `RpcClient`
  - [ ] Wrap JSON-RPC 2.0 API with exponential backoff for `ConnectionError`.
  - [ ] Mapping core node RPCs (`get_bead`, `get_tips`, `get_cohort_count`, `bitcoin_proxy`, etc.).
- [ ] **PR 6:** `testing RpcClient`
  - [ ] Testing mock HTTP responses to verify retry mechanisms and error code parsing.
- [ ] **PR 7:** `MinerManager and CpuMiner`
  - [ ] Implementing `CpuMiner` class to wrap `minerd` process (`start()`, `stop()`).
  - [ ] Adding `CpuMiner.shares_submitted()` to parse `minerd` logs via regex.
  - [ ] Implementing `MinerManager` factory.
- [ ] **PR 8:** `testing MinerManager`
  - [ ] Testing log parsing logic against mock `minerd` log outputs.

## Phase 3: Process Orchestration
- [ ] **PR 9:** `CleanupManager`
  - [ ] Implementing LIFO stack based CleanupManager class for `atexit` and `SIGTERM`/`SIGINT` handling.
  - [ ] Implementing/including memory manager for CleanupManager to ensure efficient memory cleanup.
- [ ] **PR 10:** `testing CleanupManager`
  - [ ] Testing exception swallowing during forced teardown.
- [ ] **PR 11:** `NodeManager`
  - [ ] Implementing `BitcoinNode` manager (runs Bitcoin Daemon with `-ipcbind` and `-regtest`).
[...]

PR #468, opened 2026-05-16
feat(framework): implementing custom Test runner and base modules for test framework 

kushagra0902:
This PR progresses issue #463, phase 1

This PR sets up the base for the test framework by implementing 4 main parts:
- test_runner: a custom test runner that handles initialising test scripts, their life-cycle and eventually their cleanup
- network_config: all the test configs, defaults and options are being set here. 
- log_manager and log_utils: functions to set up the basic log directory structure and log templates
- util: general util functions such as assertion equality and wait_untils.


tests/functional/framework/network_config.py
```diff
@@ -0,0 +1,60 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import List, Optional
+
+
+@dataclass
+class NetworkConfig:
+    # Number of braidpool nodes in the network
+    num_braidpool_nodes: int = 2
```
zaidmstrr: Setting this to 2 is unnecessary here; it will spawn two nodes every time. There are the cases where we just need to spawn a single node. You can set it to 1.

tests/functional/framework/network_config.py
```diff
@@ -0,0 +1,60 @@
...
+    # Number of cpu miners in the network
+    num_cpu_miners: int = 0
+
+    # Set none as the port is later assigned by the PortPool so to avoid conflicting ports
+    bitcoin_rpc_port: Optional[int] = None
```
zaidmstrr: We are not using RPC to connect Bitcoin to Braidpool. So you can remove this.

tests/functional/framework/network_config.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # Braidpool port lists (populated by PortPool, not by the test)
+    braidpool_bind_ports: List[int] = field(default_factory=list)
+
+    braidpool_rpc_ports: List[int] = field(default_factory=list)
+
+    braidpool_stratum_ports: List[int] = field(default_factory=list)
+
+    # Bitcoin/Braidpool network.  regtest | cpunet | signet.
+    network: str = "regtest"
```
zaidmstrr: Just point this to `cpunet` for now as a default behaviour.

tests/functional/framework/network_config.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # Binary locations (overridden by BITCOIN_NODE_PATH / BRAIDPOOL_BIN_PATH
+    # env vars or auto-discovered — see framework/__init__.py)
+    bitcoin_bin_path: str = "bitcoin-node"
+    braidpool_bin_path: str = "target/debug/node"
+    minerd_bin_path: str = "minerd"
+
+    # Timeouts (seconds)
+    startup_timeout: float = 30.0
```
zaidmstrr: This startup timeout is for Braidpool or Bitcoin? Because both of them can differ?

tests/functional/framework/network_config.py
```diff
@@ -0,0 +1,60 @@
...
+
+    peer_connection_timeout: float = 15.0
+
+    bead_propagation_timeout: float = 60.0
+
+    # Logging & test isolation
+    log_level: str = "DEBUG"
```
zaidmstrr: The default behaviour should not be `DEBUG` it should be `INFO`  and can be set to `DEBUG` if the user wants.

tests/functional/framework/network_config.py
```diff
@@ -0,0 +1,60 @@
...
+    keep_logs_on_success: bool = False
+    datadir_prefix: str = "/tmp/bp-test"
```
zaidmstrr: The test data path is fixed global writable path in the future you need to add some randomness to this mainly in the NodeManager class otherwise two users running tests in parallel or in the shared CI machine collide at the same prefix.

tests/functional/framework/network_config.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # Reproducibility
+    random_seed: int = 42
```
zaidmstrr: Have you planned to use this variable? Like inside `PortPool`, if possible?

tests/functional/framework/port_pool.py
```diff
@@ -0,0 +1,83 @@
...
+
+            self._next = candidate
+            return ports
+
+    # Helper function to check if the port is free
+    @staticmethod
+    def _is_free(port: int) -> bool:
+        """Return True if the OS will let us bind TCP *port* on 127.0.0.1."""
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
```
zaidmstrr: This sets `SO_REUSEADDR` on the probe socket which allows binding to ports in the TIME_WAIT state. You need to fix this.

tests/functional/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+import pytest
+
+# Ensure `framework` package is importable from any working directory by making the path global.
+_HERE = Path(__file__).parent.resolve()
+if str(_HERE) not in sys.path:
+    sys.path.insert(0, str(_HERE))
+
+#Defining the markers so that they can be used in the test code
+def pytest_configure(config: pytest.Config) -> None:
+    config.addinivalue_line(
```
zaidmstrr: The markers are defined in two places here and inside `pytest.ini` please choose one place for that. Preferably `pytest.ini`

tests/functional/framework/port_pool.py
```diff
@@ -0,0 +1,83 @@
+"""
+pytest-xdist assigns each worker a unique ID via the PYTEST_XDIST_WORKER env var: "gw0", "gw1", …  We map each worker to a non-overlapping 100-port band so parallel test processes never try to bind the same port.
+
+Within a band, ports are handed out sequentially.  Before returning a port we do a quick socket.bind() probe to confirm the OS considers it free; if it is already occupied we skip it.
+
+Thread-safety is provided by threading.Lock so multiple fixtures running concurrently inside one worker (pytest-asyncio, etc.) don't race.
```
zaidmstrr: The docstring specifies that we need `pytest-asyncio` but you didn't mention it in the requirement.txt.

zaidmstrr (changes requested):
Thanks for creating the first PR in the direction of functional testing framework. I have reviewed this PR and found some issues and suggestions which might be helpful for you. Please look at them and feel free to discuss about them.

tests/functional/framework/network_config.py
kushagra0902: I have thought of two methods, one is in LofManager(will make the pr soon), there will be a separate nested directory inside of this made for all the test runs, so that will prevent collisions in the same directory. Or the xdist worker ID may also be used to make the folders unique. What do you think will be better? 
I thought of  using the time-stamped approach

tests/functional/framework/network_config.py
kushagra0902: I declared this for later stages in chaos testing, where some random delays or jitters, etc., will be added and for making random beads disconnect.

tests/functional/conftest.py
kushagra0902: I had a small doubt here, and that's why I added it at both places for robustness. If we run from the root of the repo, and pytest finds some other config files, then this will be ignored, right? So in such cases, what is the better practice?

tests/functional/framework/network_config.py
kushagra0902: Done the changes!

tests/functional/framework/network_config.py
kushagra0902: My bad! changes done

tests/functional/framework/network_config.py
kushagra0902: Added different thresholds now

tests/functional/framework/network_config.py
kushagra0902: Changed to info now!

tests/functional/framework/port_pool.py
kushagra0902: Thanks for pointing this out, will keep in mind for future! Removed it now

tests/functional/framework/port_pool.py
kushagra0902: added!

tests/functional/framework/network_config.py
kushagra0902: Done

tests/functional/framework/network_config.py
zaidmstrr: Time-based approach might cause collisions on parallel runs; I think you can also use `mkdtemp()` but for this you can change the approach a little bit more. I will explain this about below in a separate comment.

tests/functional/conftest.py
zaidmstrr: You can remove this from the `pytest.ini`.

zaidmstrr:
The thing I want to add here is that I generally don't want to use the `pytest` for handling most of the tasks. The idea here is to use and write the custom framework similar to Bitcoin Core testing framework. This PR uses `pytest` extensively mainly in the files `conftest.py` and `pytest.ini`. The more accurate idea here is to replace those two with something similar to `test_runner` found inside the Bitcoin codebase. Apart from that `PortPool` and `NetworkConfig` can be stayed there but needs some changes according to the changes behaviour. Another problem is this PR runs a subset of tests in the same process. There is no process isolation here, so if one test fails, the other tests may contain corrupted or unexpected behaviour or info. That's the feature of pytest (`pytest-xdist`) that we don't need. For more info about process isolation you can checkout [test_runner.py](https://github.com/bitcoin/bitcoin/blob/master/test/functional/test_runner.py). The simple idea here is not to use any external dependencies. That's why the scope of this work is just limited to the implementation of test framework infrastructure and not about writing functional tests or caring about the CI/CD. Having a look at the [test_framework](https://github.com/bitcoin/bitcoin/tree/master/test/functional/test_framework) is also good.

kushagra0902:
> The thing I want to add here is that I generally don't want to use the `pytest` for handling most of the tasks. The idea here is to use and write the custom framework similar to Bitcoin Core testing framework. This PR uses `pytest` extensively mainly in the files `conftest.py` and `pytest.ini`. The more accurate idea here is to replace those two with something similar to `test_runner` found inside the Bitcoin codebase. Apart from that `PortPool` and `NetworkConfig` can be stayed there but needs some changes according to the changes behaviour. Another problem is this PR runs a subset of tests in the same process. There is no process isolation here, so if one test fails, the other tests may contain corrupted or unexpected behaviour or info. That's the feature of pytest (`pytest-xdist`) that we don't need. For more info about process isolation you can checkout [test_runner.py](https://github.com/bitcoin/bitcoin/blob/master/test/functional/test_runner.py). The simple idea here is not to use any external dependencies. That's why the scope of this work is just limited to the implementation of test framework infrastructure and not about writing functional tests or caring about the CI/CD. Having a look at the [test_framework](https://github.com/bitcoin/bitcoin/tree/master/test/functional/test_framework) is also good.

Thank you for clarifying this. I will look into these and proceed accordingly

tests/functional/test_runner.py
```diff
@@ -0,0 +1,444 @@
...
+                start_time=time.monotonic(),
+                process=proc,
+                testdir=testdir,
+                stdout_path=stdout_path,
+                stderr_path=stderr_path,
+            )
+        )
+        log_event(logger, "runner_test_started", test=scheduled.display_name, pid=proc.pid, port_seed=scheduled.port_seed)
+
+# Function to terminate a process group, firstly with SIGTERM and a wait, then with SIGKILL if it does not exit, and ignoring cases where the process is already gone. This is used for cleaning up test subprocesses on timeout or interrupt.
```
zaidmstrr: Consider writing the comment in multiple lines rather than a single long phrase.

tests/functional/test_framework/port_pool.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+class PortAllocationError(RuntimeError):
+    """Raised when a deterministic test port cannot be allocated."""
+
+
+class PortSeed:
+    """Global port seed set by the custom functional test runner."""
```
zaidmstrr: You can remove the name `custom` from this comment.

tests/functional/test_runner.py
```diff
@@ -0,0 +1,444 @@
...
+    # Function to start a test subprocess for a scheduled test, with logs directed to the appropriate files in a test-specific subdirectory. The process is launched in a new session for easier cleanup, and the RunningJob state is recorded for monitoring.
+    def _start(self, scheduled: ScheduledTest) -> None:
+        test_argv = scheduled.argv
+        script_name = test_argv[0]
+        test_base = re.sub(r"\.py$", "", script_name)
+        testdir = self.tmpdir / f"{test_base}_{scheduled.port_seed}"
+        testdir.mkdir(parents=True, exist_ok=True)
+        stdout_path = testdir / STDOUT_LOG_NAME
+        stderr_path = testdir / STDERR_LOG_NAME
+        stdout_file = open(stdout_path, "ab", buffering=0)
```
zaidmstrr: There is a possible File Descriptor leak here. The `open()` is called here in two places, but if the second of them fails for any reason (permission denied etc) then the error propagates immediately before the next `try` block hits, leaving the FD entry orphaned. The possible solution here is to use `with` block and `contextlib.ExitStack`. Although a possibility of this happening is quite low, it can only happen when the FD limit is hit after creating the first file.

zaidmstrr (commented):
Approach ACK [3cfad69](https://github.com/braidpool/braidpool/pull/468/changes/3cfad69c6bf23172a0dfa16a822e779f8f7058e0)

The overall direction and structure of this PR is good. It successfully implements the foundation of the functional testing framework by introducing core concepts like logging, scheduling, port classification, and running. There are some small nits I have recommended; please check them out.

tests/functional/test_runner.py
kushagra0902: Here, instead of ExitStack, I considered using a simple try/except approach. Since the file descriptors need to remain open for the lifetime of the child process, letting the ExitStack exit immediately after Popen would close them earlier than intended. Although this can be addressed using stack.pop_all(), I felt a try/except solution was simpler and easier to follow for this specific case.

zaidmstrr: > Since the file descriptors need to remain open for the lifetime of the child process, letting the ExitStack exit immediately after Popen would close them earlier than intended

This argument here is factually incorrect. Because when you pass the reference of a file to `Popen()` it internally calls `fork()` and immediately creates separate independent entries of the files inside the child FD. Thus, closing the parent FD doesn't affect the child's. That's why using `ExitStack` is also safe here. 

> I considered using a simple try/except approach

The approach you are using to fix this problem is try/finally not try/except.

Apart from that, your current solution also fixes the problem by wrapping the files inside a try/finally block.

zaidmstrr (approved):
Code ACK [bcdaa4e](https://github.com/braidpool/braidpool/pull/468/changes/bcdaa4e164f9c451bbba6bb0597eb9e1b790b8c7)

This looks good now.

PR #471, opened 2026-06-01
feat(miner-api): implement database-backed miner management with async support

priyashuu:
- Added database models and services for miner devices.
- Implemented CRUD operations for miners in the API.
- Introduced periodic refresh of miner data from devices.
- Updated API documentation to reflect new endpoints and functionality.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ venv
 .reviews/
 .worktrees/
 deny.toml
+miner_api.db
--- a/dashboard/src/components/MinerInventory/MinerInventoryDashboard.tsx
+++ b/dashboard/src/components/MinerInventory/MinerInventoryDashboard.tsx
@@ -1,16 +1,12 @@
 import { useState, useEffect, useRef } from 'react';
-import {
-  Miner,
-  HistoryPoint,
-  MinerAnalyticsPoint,
-  MinerAlert as Alert,
-} from './Types';
+import { Miner, HistoryPoint, MinerAnalyticsPoint } from './Types';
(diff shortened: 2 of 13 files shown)


zaidmstrr (commented):
ACK [70b6142](https://github.com/braidpool/braidpool/pull/471/changes/70b6142481f607a3ae513c36a3213a5a19837ad2)

Tested with S9 and it works fine with that. The data is persisted in the tabs upon refreshing. There are two things I noticed apart from this PR intended changes and these are not part of this PR.

1) Upon refreshing the page, and let suppose I'm on the page Inventory I will be redirected to the home page. The intended behaviour should be that we stay on the page where we are.
2) The temps showing for each miner under Inventory tab showing like `65/60` which is a little confusing. Here we can fetch the maximum temp of a miner and use that as the upper bound and use the current miner temp as is.

These can be worked on as follow-ups.

zaidmstrr (approved):
ACK [70b6142](https://github.com/braidpool/braidpool/pull/471/changes/70b6142481f607a3ae513c36a3213a5a19837ad2)

Tested with S9 and it works fine with that. The data is persisted in the tabs upon refreshing. There are two things I noticed apart from this PR intended changes and these are not part of this PR.

1) Upon refreshing the page, and let suppose I'm on the page Inventory I will be redirected to the home page. The intended behaviour should be that we stay on the page where we are.
2) The temps showing for each miner under Inventory tab showing like `65/60` which is a little confusing. Here we can fetch the maximum temp of a miner and use that as the upper bound and use the current miner temp as is.

These can be worked on as follow-ups.

PR #472, opened 2026-06-02
fix(stratum): derive extranonce1 from atomic connection_id counter

### LINKED ISSUE #461: No unique identifier for more than one miner connecting to the single node
Recently I found out that the braidpool node does not provide or use any specific parameter for different miner categorization connecting to the single node instance. This will create the problem of duplication of work and wastage of the hashrate because all the miners' connecting to the same node receive the identical mining template. In audit mode implementation I used a 2-byte prefix in extranonce1 to mitigate this issue, but in normal Braidpool mode this problem still exists. The possible solution is to use a 2 bytes prefix in extranonce 1 for miner categorization. But for this we might need to increase the size of the current extranonce1, which is currently 4 bytes.

nkatha23:
Fixes #461.

When multiple miners connect to the same Braidpool node they received identical extranonce1 values, causing them to search the exact same nonce space and duplicate work. Root cause: extranonce1 and connection_id were generated by two independent rand::thread_rng() calls,  neither derived from the other  and random assignment collides with near-certainty around 65k concurrent miners (birthday problem on a 32-bit space). Both FIXME comments in DownstreamClient::default() point at exactly this.

I've Replaced both RNG calls with a single monotonically increasing AtomicU32 counter. extranonce1 is now derived directly from connection_id via to_be_bytes(), so miner 0 gets 00000000, miner 1 gets 00000001, and so on,  guaranteed unique, non-overlapping nonce partitions. Removes the now-unused rand::RngCore import and adds a test asserting monotone IDs, unique extranonces, correct size, and correct big-endian encoding.

Note: Ordering::SeqCst on the counter could be relaxed to Ordering::Relaxed in a followup, the counter only needs to produce unique values, not coordinate memory visibility with anything else.

### DIFF
--- a/node/src/stratum.rs
+++ b/node/src/stratum.rs
@@ -8,10 +8,9 @@ use bitcoin::{absolute::Decodable, Transaction};
 use bitcoin::{BlockHash, BlockHeader, BlockTime, TxMerkleNode, Txid, Witness};
 use futures::{lock::Mutex, FutureExt};
 use num::ToPrimitive;
-use rand::RngCore;
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
-use std::sync::atomic::AtomicBool;
+use std::sync::atomic::{AtomicBool, AtomicU32, Ordering};
 use std::time::UNIX_EPOCH;
 use std::{collections::HashMap, net::SocketAddr, sync::Arc};
 use tokio::{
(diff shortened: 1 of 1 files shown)


mcelrath (approved):
Good catch, and good solution.

mcelrath:
Follow-up for @Sansh2356 -- In order to avoid scanning the same nonce range for a single miner with multiple braidpool nodes using the same pubkey (a misconfiguration, but could happen if they scp the .braidpool directory) -- we're dependent on libp2p to reject the second peer and raise a flag for the operator that he has a misconfiguration problem.

PR #473, opened 2026-06-04
Adding dockerization for braidpool

Sansh2356:
Reopening PR #423 with reduced scope .

- Including miner,braidpool-node and cpunet compatible bitcoin-node Dockerfile .
- `docker-compose` under the docker directory for single point of running docker containers .
- Including layer-caching to reduce the rebuilding of image from base by using dummy file compilation etc .
- `helm` charts have already been pushed upon and if required can be used for spawning k8s cluster locally for testing braidpool-mesh [braidpool-chart](https://artifacthub.io/packages/helm/braidpool-test-mesh/braidpool).
- This does not adds on or improves upon the existing dockerfiles for dashboard which should be requiring the update when the structure settles down .

~~Marking this currently as a draft .~~


nkatha23 (commented):
Tested this locally . The full stack came up, Bitcoin CPUNet node synced to height ~24500, Braidpool node connected via IPC, and the CPUNet miner received jobs over Stratum. The three-container setup with shared IPC socket volume works well.

One issue I hit: the braidpool-node health check (`curl http://localhost:6682/health`) always fails because the RPC server binds to `127.0.0.1:6682` by default, so the check can't reach it. The node is actually running fine,  it's just the health check that reports unhealthy, which then blocks the miner from starting. Fix is to pass `--rpc-bind 0.0.0.0:6682` to the node, either via a command override in `docker-compose.yml` or by adding `BRAIDPOOL_RPC_BIND` handling to `docker-entrypoint.sh`.

:) Also noting [Copilot's flag](https://github.com/braidpool/braidpool/pull/473#discussion_r3353518652) about the runtime image pinning `libboost-filesystem1.81.0` and `libboost-thread1.81.0`  those version-suffixed package names are Debian bookworm specific and could break if the base image changes. Non-versioned names would be safer.

Apart from those two things the setup is solid and a big improvement over building Bitcoin from source manually (which hits Cap'n Proto version mismatches on Fedora).

PR #474, opened 2026-06-06
Modifying `extend` to return  adopted orphan beads 

Sansh2356:
- Changing the `extend` function signatures to return ~~promoted_orphan_beads~~ adopted_orphan_beads .
- Updating the caller sites, the update of the usage of returned beads will be update in coming PRs .

### DIFF
--- a/node/src/braid/mod.rs
+++ b/node/src/braid/mod.rs
@@ -11,7 +11,9 @@ pub struct Cohort(pub HashSet<usize>);
 pub enum AddBeadStatus {
     DagAlreadyContainsBead,
     InvalidBead,
-    BeadAdded,
+    // Returning any promoted orphan beads whose parents have
+    // been extended currently .
+    BeadAdded { promoted_orphans: Vec<Bead> },
     ParentsNotYetReceived,
 }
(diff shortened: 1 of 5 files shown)


nkatha23 (approved):
ACK 48deb1d 

The change is clean. `process_orphan_beads()` now returns the promoted set and the recursive `promoted.extend(self.process_orphan_beads())` correctly handles transitive chains in a single extend call, the transitive test confirms this.

One observation: the call sites in `main.rs`, `lib.rs`, and `rpc_server.rs` all use `{ .. }` to ignore `promoted_orphans` for now. That's fine given the PR description says usage comes in a followup, but it's worth noting that promoted orphans are currently not being persisted to DB or forwarded to peers at those sites,  so during IBD, orphans that get promoted through `extend` won't trigger the downstream DB write path. Assuming that's intentional and the followup PR will wire that up.

Also noting: `test_invalid_json` fails due to a pre-existing port 5050 leak in `server_start_test` unrelated to this PR.

mcelrath (approved):
Looks good to me. Just a note on terminology, I'd prefer if it was called `adopted` rather than `promoted_orphans`.

PR #475, opened 2026-06-06
Feat: extend extranonce1 and extranonce2 to 8 bytes each

nkatha23:
This PR Extends extranonce1 and extranonce2 from 4 bytes to 8 bytes each,giving miners a 16-byte total extranonce space. Modern ASICs exhaust a 4-byte extranonce2 (~4 billion combinations) in milliseconds,  the larger space reduces how frequently miners need to request new jobs.

The change propagates through the stratum parsing, uncommitted metadata struct, consensus encoding/decoding, database layer, and test utilities, all nonce types move from u32 to u64. extranonce1 continues to be derived from the atomic connection_id (zero-padded into 8 bytes via(connection_id as u64).to_be_bytes()), preserving the uniqueness guarantee from #472.

The submit_work_version_rolling test was coupled to a specific CPUNet block vector,  with a different coinbase size the old nonce no longer produces a valid hash. The assertion now verifies that version rolling parsing completed without error, which is what the test is actually checking. PoW validity against a live block belongs in integration tests.

Note: test_invalid_json fails due to a pre-existing port 5050 leak in server_start_test,  this predates this branch and is unrelated to these changes.

### DIFF
--- a/node/src/db/db_handlers.rs
+++ b/node/src/db/db_handlers.rs
@@ -404,9 +404,9 @@ pub async fn fetch_beads_in_batch(
                     MedianTimePast::from_u32(row.get::<u32, _>("broadcast_timestamp")).unwrap();
 
                 bead.uncommitted_metadata.extra_nonce_1 =
-                    u32::from_str_radix(&row.get::<String, _>("extranonce1"), 16).unwrap();
+                    u64::from_str_radix(&row.get::<String, _>("extranonce1"), 16).unwrap();
                 bead.uncommitted_metadata.extra_nonce_2 =
-                    u32::from_str_radix(&row.get::<String, _>("extranonce2"), 16).unwrap();
+                    u64::from_str_radix(&row.get::<String, _>("extranonce2"), 16).unwrap();
 
                 bead.uncommitted_metadata.signature =
                     Signature::from_slice(&row.get::<Vec<u8>, _>("signature")).unwrap();
@@ -493,10 +493,16 @@ pub async fn fetch_bead_by_bead_hash(
             let min_target = CompactTarget::from_consensus(row.get::<u32, _>("min_target"));
             let weak_target = CompactTarget::from_consensus(row.get::<u32, _>("weak_target"));
             let miner_ip = row.get::<String, _>("miner_ip");
-            let extranonce_1 =
-                u32::from_str_radix(&row.get::<String, _>("extranonce1"), 16).unwrap();
-            let extranonce_2 =
-                u32::from_str_radix(&row.get::<String, _>("extranonce2"), 16).unwrap();
+            let extranonce_1 = u64::from_str_radix(&row.get::<String, _>("extranonce1"), 16)
+                .map_err(|e| DBErrors::TupleAttributeParsingError {
+                    error: e.to_string(),
+                    attribute: "extranonce1".to_string(),
+                })?;
+            let extranonce_2 = u64::from_str_radix(&row.get::<String, _>("extranonce2"), 16)
+                .map_err(|e| DBErrors::TupleAttributeParsingError {
+                    error: e.to_string(),
+                    attribute: "extranonce2".to_string(),
+                })?;
             let broadcast_timestamp =
                 MedianTimePast::from_u32(row.get::<u32, _>("broadcast_timestamp")).unwrap();
             let signature = Signature::from_slice(&row.get::<Vec<u8>, _>("signature")).unwrap();
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -100,8 +100,8 @@ pub const EXTRANONCE2_SIZE: usize = 8;
 /// **Separator between `EXTRANONCE1` and `EXTRANONCE2`.**
 ///
 /// This is an array of bytes used to clearly delimit the two extranonce parts.
-/// In this testing configuration, the separator length equals
-/// `EXTRANONCE1_SIZE + EXTRANONCE2_SIZE` (8 bytes total),
+/// The separator length equals
+/// `EXTRANONCE1_SIZE + EXTRANONCE2_SIZE` (16 bytes total),
 /// and is filled with the byte value `1u8` for simplicity.
 /// can be changed accordingly as per discussion .
 pub const EXTRANONCE_SEPARATOR: [u8; EXTRANONCE1_SIZE + EXTRANONCE2_SIZE] =
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -88,15 +88,15 @@ pub fn get_next_template_id() -> TemplateId {
 /// In Stratum mining, the extranonce is split into two parts:
 /// `EXTRANONCE1` (prefix) and `EXTRANONCE2` (suffix).
 ///
-/// This constant defines the size of `EXTRANONCE1` as **4 bytes**.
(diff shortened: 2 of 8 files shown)


node/src/utils/mod.rs
```diff
@@ -104,8 +104,8 @@ pub fn create_test_bead(nonce: u32, prev_hash: Option<BlockHash>) -> Bead {
         start_timestamp: time_val,
         weak_target: weak_target,
     };
-    let extra_nonce_1 = 42;
-    let extra_nonce_2 = rand::random::<u32>();
+    let extra_nonce_1: u64 = 42;
```
mcelrath: Let's make sure this is initialized correctly and 42 doesn't get propagated to miners.

node/src/stratum.rs
```diff
@@ -2408,7 +2408,7 @@ mod test {
             lock_time: LockTime::ZERO,
         };
         let test_template_header = bitcoin::block::Header {
-            bits: bitcoin::pow::CompactTarget::from_unprefixed_hex("1d00ffff").unwrap(),
+            bits: bitcoin::pow::CompactTarget::from_unprefixed_hex("207fffff").unwrap(),
```
mcelrath, line 2412: Why did this change?

nkatha23: The test was built around a real CPUNet block. When I doubled the extranonce size, the coinbase changed, different merkle root, different block hash. The old nonce no longer met the `1d00ffff` target.

Changed `bits` to `207fffff` and also updated the assertion (see line ~2494) 
from:

    assert_eq!(json_response, true);  // was testing PoW validity

to:

    assert!(std_response.result.is_some(), "Expected a result in response");
    assert!(std_response.error.is_none(), "Expected no error in response");

The test is named `submit_work_version_rolling` , its job is to verify that version rolling (BIP310) parsing and masking logic works end-to-end without  error. PoW validity against a specific block hash is an integration test concern, not a unit test concern. The test still exercises the full submission 
path,  coinbase reconstruction, merkle root computation, version mask application, header construction, just without coupling the result to a specific hash value.

nkatha23:
@zaidmstrr, once this merges, I can open a follow-up PR targeting [audit-main](https://github.com/braidpool/braidpool/tree/audit-main) to port the extranonce1/2 size changes there, since this 
affects your Audit Mode PR (#399).

zaidmstrr:
Really appreciate your efforts for doing so. But in audit mode it's not just a single limit increase change, but I also need to change the behaviour of storing the data and removing the SHA256 way of storing the hash inside extranonce. So with that you can leave that to me. Although thanks for trying to help.

node/src/stratum.rs
zaidmstrr, line 1083: The connection ID here uses `AtomicU32` but now the `extranonce` is increased so this can become `AtomicU64`. But apart from this, using `AtomicU32` or `AtomicU64` doesn't fix the problem of collisions. Because every braidpool node starting will be initiated with the same extranonce atomic value, starting from 0. This will make extranonce the same and some deterministic for every node, which we don't absolutely want. Because if extranonce is the same for all, then the only remaining factor randomness left is `comm_pub_key` and as we discussed early, in some extreme cases this can also be the same. So there's no random uniqueness factor left. This is the same problem I mentioned before, but at that time we didn't have 8 bytes of extranonce, so we couldn't technically fix this problem, but now I think we can. So, the better fix here is to use the 8 bytes of random value inside the extranonce.

node/src/lib.rs
zaidmstrr: The documentation needs an update.

node/src/db/db_handlers.rs
```diff
@@ -404,9 +404,9 @@ pub async fn fetch_beads_in_batch(
                     MedianTimePast::from_u32(row.get::<u32, _>("broadcast_timestamp")).unwrap();
 
                 bead.uncommitted_metadata.extra_nonce_1 =
-                    u32::from_str_radix(&row.get::<String, _>("extranonce1"), 16).unwrap();
+                    u64::from_str_radix(&row.get::<String, _>("extranonce1"), 16).unwrap();
```
zaidmstrr, line 407: This still contains the `unwrap()` and the fix you did in commit  [dcb58ee](https://github.com/braidpool/braidpool/pull/475/changes/dcb58ee43ba293bf62426b449e6d313e5ed5abac) is not implemented here.

zaidmstrr (commented):
I reviewed the PR. Found some small inconsistencies around the documentation and one medium level issue, which was also raised previously by me. I think this PR is the best place to fix that issue because now we have 8 bytes of space rather than the 4.

node/src/stratum.rs
zaidmstrr: You removed the check `assert_eq!(json_response, true);` which satisfies that the block is accepted by the PoW check. The new check `assert!(std_response.result.is_some() && std_response.error.is_none());` is more lenient and can return true even if the response from the stratum is like `Okay(StandardResponse{result:Some(false)})`. The ideal situation here is to use the existing check. And its not very hard to find the valid nonce values with the difficulty of `207fffff`. Almost 50% of the value will satisfy this difficulty.

node/src/stratum.rs
nkatha23: Fixed in commit 0fbd192,  extranonce1 now uses rand::random::<u64>().to_be_bytes(), unique across all nodes. connection_id (AtomicU32) is kept for logging only.

node/src/db/db_handlers.rs
nkatha23: The or_insert_with closure has signature || -> Bead — ? is a compile error (E0277) inside it. The entire closure uses .unwrap() consistently across 8 fields. The proper fix is refactoring into a fallible helper 
returning Result<Bead, DBErrors> , out of scope here(will open a follow-up).
The second occurrence (fetch_bead_by_bead_hash, commit dcb58ee) is in a Result-returning function and is properly propagated.

node/src/stratum.rs
nkatha23: Great.  the test now grinds a valid nonce at test time by replicating handle_submit's coinbase construction and iterating until grind_target.is_met_by(header.block_hash()). The original assert_eq!(json_response, true) is restored.

node/src/db/db_handlers.rs
zaidmstrr: Seems valid.

node/src/stratum.rs
```diff
@@ -2559,18 +2602,6 @@ mod test {
...
-        );
-        assert_eq!(
-            client2.extranonce1,
-            client2.connection_id().to_be_bytes().to_vec()
-        );
-        assert_eq!(
-            client3.extranonce1,
-            client3.connection_id().to_be_bytes().to_vec()
-        );
+        // extranonce1 is random — not derived from connection_id, so no relationship to assert
```
zaidmstrr: This comment is not needed here. It can be removed.

zaidmstrr (commented):
Code ACK [11f16fe](https://github.com/braidpool/braidpool/pull/475/changes/11f16feab1b30019fa57fc5fb5f88c6fed95a62a)

I reviewed the code and most of the things look good. Just left with a minor nit

zaidmstrr (approved):
Code ACK [855d006](https://github.com/braidpool/braidpool/pull/475/changes/855d006a1a5c625280233ae0b19001ed2a6f9bda)

Thanks for doing the changes. Now look's good.

PR #476, opened 2026-06-07
feat(test_framework): Implementing RPC client, miner manager and Cleanup Manager for test framework

kushagra0902:
This PR advances #463. 
Three main modules are implemented in this PR:
- rpc_client: for handling and providing RPCs of braidpool nodes to the test framework
- miner_manager: starting and handling the lifecycle of the CPU miner
- cleanup_manager: for handling cleanup of all the processes run by the tests, ensuring memory safety.

This PR is a follow up pr for #468

### DIFF
--- a/tests/functional/test_framework/rpc_client.py
+++ b/tests/functional/test_framework/rpc_client.py
@@ -91,7 +91,7 @@ class JsonRpcClient:
             self._auth_header = f"Basic {encoded_auth}"
 
     def _call(self, method: str, params: Any = None) -> Any:
-        started = time.monotonic()
+        started = time.monotonic() # ensures that even if system time changed, actual time is maintained
         request_id = next(self._ids)
         payload_obj = {"jsonrpc": "2.0", "id": request_id, "method": method}
         if params is not None:
@@ -268,6 +268,15 @@ class RpcClient(JsonRpcClient):
 
     def get_ipc_stats(self) -> dict:
         return self._call("getipcstats")
+
+    def get_node_info(self, bead_hash: str) -> dict:
+        return self._call("getnodeinfo", [bead_hash])
+
+    def staged_transactions(self) -> list[dict]:
+        return self._call("stagedtransactions")
+
+    def unstage_transactions(self, txid: str) -> bool:
+        return self._call("unstagetransactions", [txid])
 
     def bitcoin_proxy(self, method: str, params: Any = None) -> Any:
         return self._call("bitcoinproxy", [method, params or []])
--- a/tests/functional/test_framework/rpc_client.py
+++ b/tests/functional/test_framework/rpc_client.py
@@ -56,11 +56,11 @@ def _default_transport(url: str, payload: bytes, timeout: float, auth_header: st
         method="POST",
     )
     with urllib.request.urlopen(request, timeout=timeout) as response:
-        return response.read()
+        return response.read(_MAX_HTTP_BODY_BYTES)
 
 
-class RpcClient:
-    """Synchronous JSON-RPC 2.0 client for a Braidpool node."""
+class JsonRpcClient:
+    """Synchronous JSON-RPC 2.0 client."""
 
     def __init__(
         self,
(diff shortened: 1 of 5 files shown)


mcelrath:
The new test framework is headed in a good direction, but there are a few follow-ups to address before it’s fully solid.

1. Cap successful RPC response size too. In tests/functional/test_framework/rpc_client.py:58, normal 200 responses are read without a size limit, while error responses already have one. Please apply the same cap to successful responses so a misbound or misbehaving local service can’t force unbounded memory use.

2. Tighten trace logging for RPC params. In rpc_client.py:102, trace mode logs raw params. Key-based redaction won’t help if credentials or secrets are passed positionally, so either redact/suppress params for sensitive methods or avoid logging raw positional args.

3. Separate cleanup registration from process-wide handler ownership. In tests/functional/test_framework/cleanup_manager.py:29, constructing CleanupManager installs signal/atexit handlers immediately. That makes ownership implicit and lets later instances replace earlier handler state. It would be safer if handler installation were owned once at the top level, with CleanupManager otherwise acting as a plain callback stack.

4. Make the miner/node interface explicit. In tests/functional/test_framework/miner_manager.py:150, spawn_miner() relies on getattr(node, "stratum_port"). Please replace that ad hoc runtime contract with a small typed protocol/adapter or a clearer interface so misuse fails earlier and more predictably.

5. Consider splitting RpcClient responsibilities. Right now RpcClient owns transport, retry behavior, API helpers, readiness polling, and fallback dispatch. That works for now, but it will get harder to evolve safely as the framework grows. A clean split between generic JSON-RPC transport and Braidpool-specific helpers would make follow-on work easier.

Overall: good structure, no blocker-level security issue, but the RPC size cap is the most important concrete fix.

kushagra0902:
> The new test framework is headed in a good direction, but there are a few follow-ups to address before it’s fully solid.
> 
> 1. Cap successful RPC response size too. In tests/functional/test_framework/rpc_client.py:58, normal 200 responses are read without a size limit, while error responses already have one. Please apply the same cap to successful responses so a misbound or misbehaving local service can’t force unbounded memory use.
> 2. Tighten trace logging for RPC params. In rpc_client.py:102, trace mode logs raw params. Key-based redaction won’t help if credentials or secrets are passed positionally, so either redact/suppress params for sensitive methods or avoid logging raw positional args.
> 3. Separate cleanup registration from process-wide handler ownership. In tests/functional/test_framework/cleanup_manager.py:29, constructing CleanupManager installs signal/atexit handlers immediately. That makes ownership implicit and lets later instances replace earlier handler state. It would be safer if handler installation were owned once at the top level, with CleanupManager otherwise acting as a plain callback stack.
> 4. Make the miner/node interface explicit. In tests/functional/test_framework/miner_manager.py:150, spawn_miner() relies on getattr(node, "stratum_port"). Please replace that ad hoc runtime contract with a small typed protocol/adapter or a clearer interface so misuse fails earlier and more predictably.
> 5. Consider splitting RpcClient responsibilities. Right now RpcClient owns transport, retry behavior, API helpers, readiness polling, and fallback dispatch. That works for now, but it will get harder to evolve safely as the framework grows. A clean split between generic JSON-RPC transport and Braidpool-specific helpers would make follow-on work easier.
> 
> Overall: good structure, no blocker-level security issue, but the RPC size cap is the most important concrete fix.

I fixed these issues and have updated the PR. Kindly let me know if any other changes are needed.

tests/functional/test_framework/rpc_client.py
```diff
@@ -0,0 +1,321 @@
...
+    def get_peer_info(self) -> dict:
+        return self._call("getpeerinfo")
+
+    def get_ipc_stats(self) -> dict:
+        return self._call("getipcstats")
+
+    def get_node_info(self, bead_hash: str) -> dict:
+        return self._call("getnodeinfo", [bead_hash])
+
+    def staged_transactions(self) -> dict:
```
zaidmstrr: This retruns `dict` here but in the rpc_server it processes through `serde_json::to_value` on `Vec<StagedTxEntry>` and returns a JSON array. But the python code thinks it will return the `dict` but in actual the right type is `list[dict]`

zaidmstrr (commented):
Approach ACK [12ab5a5](https://github.com/braidpool/braidpool/pull/476/changes/12ab5a52b140fa5d938f9fcc43e7250257a0db08)

Most of the things look good and correct. Just a single thing you need to look at, which I have found.

tests/functional/test_framework/rpc_client.py
kushagra0902: Thanks for the review! I solved this one. Kindly let me know if anything else is needed

zaidmstrr (approved):
Code ACK [a061a32](https://github.com/braidpool/braidpool/pull/476/changes/a061a327a2c4e9cbe8f551028dd63902c733e86e)

Thanks for fixing the suggested code. This PR now looks correct.

PR #477, opened 2026-06-08
Fix(tests): replace hardcoded stratum ports with OS-assigned port 0

nkatha23:
While reviewing PRs #309 and #474, I kept seeing test_invalid_json fail with AddrInUse on port 5050. Looking into it, all five stratum tests bind fixed ports (3353, 3356, 3357, 3358, 5050), if a previous test leaves a 
server running or two tests race to bind the same port, the test fails with a misleading error that looks like a code bug rather than a port conflict.

The fix adds an optional oneshot channel parameter to run_stratum_service. After binding, the server sends its actual address back through the channel. Tests use port 0 so the OS assigns a free port, then await the receiver before connecting with no sleep, race windows, or hardcoded ports.

rpc_server.rs already uses the same pattern via server.local_addr(), so this is consistent with how the rest of the codebase handles this.

Production call in main.rs passes None, zero impact outside tests.

### DIFF
--- a/node/src/stratum.rs
+++ b/node/src/stratum.rs
@@ -19,7 +19,7 @@ use tokio::{
         tcp::{OwnedReadHalf, OwnedWriteHalf},
         TcpListener,
     },
-    sync::{mpsc, RwLock},
+    sync::{mpsc, oneshot, RwLock},
 };
 use tokio_stream::StreamExt;
 use tokio_util::codec::{FramedRead, LinesCodec};
@@ -1783,6 +1783,7 @@ impl Server {
         notification_sender: mpsc::Sender<NotifyCmd>,
         swarm_handler: Arc<Mutex<SwarmHandler>>,
         ibd_or_not: Arc<AtomicBool>,
+        bound_addr_tx: Option<oneshot::Sender<SocketAddr>>,
     ) -> Result<(), Box<std::io::Error>> {
         debug!("Starting stratum server");
         let bind_address = format!(
@@ -1796,6 +1797,9 @@ impl Server {
                 return Err(Box::new(e));
             }
         };
+        if let Some(tx) = bound_addr_tx {
+            let _ = tx.send(listener.local_addr()?);
+        }
 
         let endpoints = crate::utils::server_endpoints(
             &self.stratum_config.hostname,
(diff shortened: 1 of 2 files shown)


zaidmstrr (commented):
Concept ACK [17a64a7](https://github.com/braidpool/braidpool/pull/477/changes/17a64a7fa7b5ad5e2ba97549908e35782b667105)

Thanks for creating the PR. The problem seems to be exits. I haven't reviewed the code yet nor looked at the approach but will review it soon.

Sansh2356:
@nkatha23 thanks for this PR, yes indeed this was the issue and using os assigned ports is the intended fix .

node/src/stratum.rs
```diff
@@ -1796,6 +1797,9 @@ impl Server {
                 return Err(Box::new(e));
             }
         };
+        if let Some(tx) = bound_addr_tx {
+            let _ = tx.send(listener.local_addr()?);
+        }
 
         let endpoints = crate::utils::server_endpoints(
```
zaidmstrr, line 1805: After this change a test calling this with `0` shows the log `stratum+tcp://127.0.0.1:0` which is wrong.

zaidmstrr (commented):
Tested the PR, most of the things look good. But in the future we can perform a good refactor as a follow-up PR, which can enhance the code readability and provide a nicer architecture. By moving the `TcpListener::bind` call out of the `run_stratum_service` and into the caller. Here, the caller binds port 0 itself; the server function just accepts the ready listener, not a port.

node/src/stratum.rs
nkatha23: Good catch. Fixed in https://github.com/braidpool/braidpool/pull/477/changes/4207ad8cd855f1fc685c0540f34ee11d5a17db3d `bound_addr.port()` is now used consistently for both `server_endpoints` and the warn log. The endpoint will show the real OS-assigned port instead of 0.

zaidmstrr (approved):
Code ACK [4207ad8](https://github.com/braidpool/braidpool/pull/477/changes/4207ad8cd855f1fc685c0540f34ee11d5a17db3d)

Thanks for doing the changes. Now looks good to me.

PR #478, opened 2026-06-13
Updating db handlers for orphan insertions and batched queries 

Sansh2356:
This is a continuation of PR #309 and PR #474 having scope :-  

- Updating db_handler with Batched query insertion and removing `SequentialQueryInsertion`.
- Removing the tests associated with sequential insertion method and updating the existent method wrt to orphan handling.
- Persisting orphan beads via db command handlers and adding helpers to `Braid` for constructing parents set and passing into `BeadInsertData` containing parent_bead_idx and timestamp as per schema for insertion,to be passed onto the db_handler .
- Adding appropriate error types for handling consensus failures.

### DIFF
--- a/node/src/db/db_handlers.rs
+++ b/node/src/db/db_handlers.rs
@@ -84,8 +84,9 @@ impl DBHandler {
             db_handler_tx,
         ))
     }
+    /// Builds the `(transactions, relatives, parent_timestamps)` JSON value tuples for a single
+    /// bead.
     fn prepare_bead_tuple_values(
-        &self,
         data: &BeadInsertData,
     ) -> (
         Vec<serde_json::Value>,
@@ -131,7 +132,7 @@ impl DBHandler {
             let bead = &data.bead;
             let bead_id = data.bead_id;
 
-            let (txs, relatives, parent_ts) = self.prepare_bead_tuple_values(data);
+            let (txs, relatives, parent_ts) = Self::prepare_bead_tuple_values(data);
 
             all_bead_data.push(json!({
                 "id": bead_id as i64,
@@ -275,7 +276,12 @@ impl DBHandler {
                     inserted_count = inserted_count,
                     "Bulk insert chunk failed, rolling back"
                 );
-                local_transaction.rollback().await.ok();
+                if let Err(rollback_err) = local_transaction.rollback().await {
+                    error!(
+                        error = ?rollback_err,
+                        "Failed to roll back batch transaction after chunk insert failure"
+                    );
+                }
                 return Err(e);
             }
             inserted_count += chunk.len() as u32;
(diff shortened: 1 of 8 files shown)


node/src/db/db_handlers.rs
```diff
@@ -1,62 +1,72 @@
...
+const BULK_INSERT_PARENT_TIMESTAMPS: &'static str =
+    "INSERT INTO ParentTimestamps (parent, child, timestamp) 
+    SELECT json_extract(value, '$.parent'), json_extract(value, '$.child'), 
+        json_extract(value, '$.timestamp') 
+    FROM json_each(?);";
 #[derive(Debug)]
 pub struct DBHandler {
     //Query receiver inherit to handler only
     receiver: Receiver<BraidpoolDBTypes>,
     //Shared across tasks for accessing DB after contention using `Mutex`
```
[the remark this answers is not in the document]
Sansh2356: ACK

node/src/db/db_handlers.rs
zaidmstrr: The same inefficiency is present here in the LEFT JOINs. `Transactions` and `ParentTimestamps` are independent, so for a bead with N txids and M parents, SQL produces N×M rows.

zaidmstrr: Due to this the `num_batches` computation is incorrect.

node/src/db/db_handlers.rs
```diff
@@ -69,214 +79,271 @@ impl DBHandler {
...
-    Ok((txs_json, relatives_json, parent_ts_json))
-}
 //Fetching beads in batch
 pub async fn fetch_beads_in_batch(
-    db_pool: Arc<Mutex<Pool<Sqlite>>>,
+    db_pool: &Pool<Sqlite>,
     batch_size: u32,
 ) -> Result<Vec<Bead>, DBErrors> {
-    let conn = db_pool.lock().await.clone();
     let total_rows: u32 = sqlx::query(
```
zaidmstrr: This entire `total_rows` fetching is highly inefficient. Here you are using a LEFT JOIN, which maps out a one-to-many relationship. A bead containing multiple parents/transactions will create multiple duplicate entries of itself. For example, if a bead has 5 transactions and 3 parents, the database generates 15 rows just for that one bead.

node/src/db/db_handlers.rs
```diff
@@ -665,171 +754,109 @@ pub mod test {
...
+        assert_eq!(
+            bead_insert_data.len(),
+            current_file_braid.beads.len(),
+            "resolve_many dropped beads"
+        );
+        // Splitting a dummy orphan bead from all the beads in the existing braid test file
+        // for testing the orphan insertions as well .
+        let split_at = bead_insert_data.len().saturating_sub(1);
+        let orphans = bead_insert_data.split_off(split_at);
+        println!("{orphans:?}");
```
zaidmstrr: Please use `tracing` instead of `println!`.

node/src/db/db_handlers.rs
```diff
@@ -665,171 +754,109 @@ pub mod test {
...
+
+            assert_eq!(
+                fetched.committed_metadata.parents.len(),
+                bead.committed_metadata.parents.len(),
+                "Parent count mismatch for bead {}",
+                bead.block_header.block_hash()
+            );
         }
+
+        println!("Batch insertion test completed successfully!");
```
zaidmstrr: Use `tracing` instead of `println!`

node/src/db/db_handlers.rs
```diff
@@ -69,214 +79,271 @@ impl DBHandler {
...
+                "txid": hex::encode(tx.to_byte_array()),
+                "bead_id": bead_id,
+            }));
+        }
+        (txs_values, relatives_values, parent_ts_values)
+    }
+
+    /// Inserting chunks for bulk insertions
+    async fn bulk_insert_chunk(
         &self,
```
zaidmstrr: The `&self` parameter is not used inside this function anywhere else. It's good to remove from here.

node/src/db/db_handlers.rs
zaidmstrr: Here you are using `map_err` and then `unwrap()` which doesn't resolve to anything but just a simple `unwrap()` call. Using `map_err` has no extra benefit here; instead, the code becomes more verbose.

zaidmstrr: The same thing happens here and below inside `payout_address` and `comm_pub_key`

node/src/lib.rs
```diff
@@ -366,45 +366,49 @@ impl SwarmHandler {
...
-                        },
-                    })
-                    .await
-                {
-                    Ok(_) => {
-                        debug!(
-                            hash = %bead_hash,
-                            "InsertBeadSequentially sent to DB thread"
-                        );
+                match db::BeadInsertData::resolve(&braid_data, &weak_share) {
```
zaidmstrr: This is too much nested inside `match`, like there are 3 nested `match` clauses. What if we can create a small helper function that do 3 step operation, and inside that function we can use `?` operator for better propagation?

node/src/main.rs
```diff
@@ -90,7 +90,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     // starting from that block as genesis
     let initial_bead_fetch_handle = tokio::spawn(async move {
         let mut guard = braid_ref.write().await;
-        let fetched_beads = fetch_beads_in_batch(db_connection_pool_ref, 50).await?;
+        let fetched_beads = fetch_beads_in_batch(&db_connection_pool_ref, 50).await?;
```
zaidmstrr: `50` is a magic number here. It should be named to some `const`. Also, why did you choose this num here? Like beads are very small in size so can we use a little bit higher number here?

node/src/db/db_handlers.rs
```diff
@@ -1,62 +1,72 @@
...
-FROM json_each(?);
-";
+#[cfg(test)]
+use tracing::info;
+use tracing::{debug, error, trace};
+pub const DB_CHANNEL_CAPACITY: usize = 1024;
+/// Maximum number of beads (including orphans) to insert in a single bulk query to limit the memory consumption
+pub const BATCH_INSERT_THRESHOLD: usize = 500;
+//Bulk insertion sub-queries
+const BULK_INSERT_BEADS: &'static str =
```
zaidmstrr: Writing `'static` lifetime is redundant here. As for `static` items  all reference lifetimes default to `'static` automatically.

zaidmstrr (commented):
I have tested and reviewed this PR manually and found some areas for improvement. I have also identified an issue related to the database where computation can go exponentially higher with the number of beads having more than one parent and transaction. The problem currently is not visible because most of the beads we have contain a single transaction and reference to only one parent. But in the future it can be drastically inexpensive for us. 

I will strongly recommend a re-evaluation of the database implementation (including the previously written code) and possible refactoring of the codebase.

Sansh2356:
> I have tested and reviewed this PR manually and found some areas for improvement. I have also identified an issue related to the database where computation can go exponentially higher with the number of beads having more than one parent and transaction. The problem currently is not visible because most of the beads we have contain a single transaction and reference to only one parent. But in the future it can be drastically inexpensive for us.
> 
> I will strongly recommend a re-evaluation of the database implementation (including the previously written code) and possible refactoring of the codebase.

I will look at the SQL queries again wrt to query optimizations and correctness and update some of those following in accordance with query optimization engine .

node/src/main.rs
Sansh2356: Will add a `const` for it .

node/src/db/db_handlers.rs
Sansh2356: Acknowledged .

node/src/db/db_handlers.rs
Sansh2356: Acknowledged can be changed by directly using `Self` as the object method instead of passing self as reference .

node/src/db/db_handlers.rs
Sansh2356: Agreed, some changes were left locally by me during the last PR which updated `insert_beads_batch` accidentally, sorry for that and thanks for pointing this out will update it no joins are actually required here because currently it is behaving like cartesian product instead .

Sansh2356: Agreed .

node/src/lib.rs
Sansh2356: We can change the return type to `Box<dyn std::error:Error>` instead and propagate errors directly by logging them first and then via `?` for now, since i have already done this change in upcoming PRs since it is required at ibd beads persist, floodsub received beads also and therefore scoped in other PR.

zaidmstrr:
This PR has some conflicts please fix them.

node/src/db/db_handlers.rs
zaidmstrr: This is still open in the latest commits.

zaidmstrr (approved):
Code ACK [072a5b9](https://github.com/braidpool/braidpool/pull/478/changes/072a5b9d5dd36132ed5c0ee301e00c9840730dc8)

Thanks for doing the required changes now everything looks good to me.

PR #479, opened 2026-06-15
Refactor(stratum): move TcpListener binding to caller

nkatha23:
run_stratum_service was doing two things: binding to a port and running the server loop. The caller had no control over how the socket was acquired, and tests had no way to discover the actual bound port without extra machinery.

This PR moves binding to the caller. run_stratum_service now accepts a ready TcpListener as its first argument. The caller binds the socket (using port 0 for tests, the configured port for production) and passes it in. The function calls listener.local_addr() for logging and proceeds directly to the accept loop.

This is the same pattern rpc_server.rs already uses, the server is built first, local_addr() is called to get the real port, then the server starts. Making run_stratum_service consistent with that approach.

Tests now bind their own listener before spawning the server and read the port directly from local_addr(), with no channel needed, no sleep, and no race window. Production call in main.rs binds using the configured hostname and port before spawning.

### DIFF
--- a/node/src/stratum.rs
+++ b/node/src/stratum.rs
@@ -20,7 +20,7 @@ use tokio::{
         tcp::{OwnedReadHalf, OwnedWriteHalf},
         TcpListener,
     },
-    sync::{mpsc, oneshot, RwLock},
+    sync::{mpsc, RwLock},
 };
 use tokio_stream::StreamExt;
 use tokio_util::codec::{FramedRead, LinesCodec};
@@ -102,8 +102,6 @@ impl Default for BlockTemplate {
 pub struct StratumServerConfig {
     /// Hostname or IP address to bind the Stratum server.
     pub hostname: String,
-    /// TCP port for incoming Stratum connections.
-    pub port: u16,
     /// Initial mining difficulty assigned to new clients as per in the `braidpool_spec.md`.
     pub start_difficulty: u64,
     /// Minimum allowed mining difficulty as per in the `braidpool_spec.md`.
@@ -118,7 +116,6 @@ impl Default for StratumServerConfig {
     fn default() -> Self {
         Self {
             hostname: String::from("0.0.0.0"),
-            port: 3333,
             //Placeholders can be changed in default
             start_difficulty: 1,
             minimum_difficulty: 1,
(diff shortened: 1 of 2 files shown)


Sansh2356 (dismissed):
t-ACK [`8555828`](https://github.com/braidpool/braidpool/pull/479/changes/8555828bfa99488fa0039eb78860a0188aace30c), just a small suggestion you can look at by moving listener in main the `StratumConfig` also takes the same input as `args.port` being sent via cli by user and it is not used as much so we can directly use it instead and remove the port field from `StratumConfig` all through i don't think it it is required completely, i am fine with current one as well .

Sansh2356:
@nkatha23 can you also switch to OS-assigned ports for other places as well like for `rpc_server` tests ?

nkatha23:
> t-ACK [`8555828`](https://github.com/braidpool/braidpool/pull/479/changes/8555828bfa99488fa0039eb78860a0188aace30c), just a small suggestion you can look at by moving listener in main the `StratumConfig` also takes the same input as `args.port` being sent via cli by user and it is not used as much so we can directly use it instead and remove the port field from `StratumConfig` all through i don't think it it is required completely, i am fine with current one as well .

On the port field, I've been thinking the same thing. Since binding moved to the caller, stratum_config.port is no longer used inside the server itself. We could remove it and have main.rs use args.stratum_port directly.

node/src/stratum.rs
```diff
@@ -1780,28 +1777,14 @@ impl Server {
     /// * `Err(Box<std::io::Error>)` – If binding to the server address fails.
```
zaidmstrr: Please update the error message as the binding is not happening inside this function. 

```diff
-     /// * `Err(Box<std::io::Error>)` – If binding to the server address fails.
+     /// * `Err(Box<std::io::Error>)` – If the listener's local address cannot be resolved.

```

node/src/stratum.rs
zaidmstrr: Same here you need to update this:

```diff
-     /// This asynchronous function continuously listens on the configured hostname and port
+     /// This asynchronous function continuously listens on the provided `TcpListener`

```

node/src/stratum.rs
```diff
@@ -2349,14 +2331,13 @@ mod test {
         assert_eq!(response["id"], 1);
     }
 
-    //TODO: this test is currently conditional wrt to master branch for our forked rust-bitcoin hence commented out
-
     #[tokio::test]
     async fn submit_work_version_rolling() {
-        /*
-        Test block taken - 00000020e6ebb395a1e2ba60f17650d790309e21af08062229ad955376ac574300000000e8de27818e402a0d5e6028f363be4b47d809ad348e6bc88ac2f9c2bedf0409e9337edf68ffff001d7aeb8b0601020000000001010000000000000000000000000000000000000000000000000000000000000000ffffffff1602611e089495ac0803000000094272616964706f6f6cffffffff0300f2052a01000000160014e470d0179325db88b55771f6c0a5139dd81d73180000000000000000266a24aa21a9ede2f61c3f71d1defd3fa999dfa36953755c690689799962b48bebd836974e8cf900000000000000002a6a286272616964706f6f6c5f626561645f6d657461646174615f686173685f33326201020304050607080120000000000000000000000000000000000000000000000000000000000000000000000000
```
zaidmstrr, line 2357: I think leaving this here without removing it is a better option, as this provides a better idea about the values used inside the test and helpful for debugging. Instead, what you can do is, instead of having a very long line of string you can split it into multiple lines for better readability.

zaidmstrr (commented):
Approach ACK [d35a294](https://github.com/braidpool/braidpool/pull/479/changes/d35a294c76a17cd6556e13b945179942f92eaaa3)

The overall structure and behaviour of the PR look good. I have suggested few nits which can be done.

node/src/stratum.rs
nkatha23: The removed hex was actually the pre-#475 reference block: after the extranonce extension the coinbase changed, so its merkle root, bits, and nonce no longer match what this test constructs. Restoring it verbatim would have put back misleading debug data. Instead I've added an annotated dump of the block this test builds today, split per your suggestion, with the extranonce inputs documented,  verified byte-identical across runs, so it's a stable reference. Additional assertions on this test are coming in a small follow-up PR to keep this one scoped to the listener refactor.

zaidmstrr: Okay, that's fine.

zaidmstrr (commented):
Code ACK [c4e89d2](https://github.com/braidpool/braidpool/pull/479/changes/c4e89d263b5416cc061feade2a856fd8fa1ad2d5)

All of the previous nits are now resolved. Changes look good.

zaidmstrr (approved):

PR #482, opened 2026-06-23
fix(tests): replace hardcoded RPC server ports with OS-assigned port 0

nkatha23:
RPC server tests were binding to hardcoded ports,  9001 through 9101, 8889, 9000 and others. If two tests run in parallel or a previous test leaves a socket in TIME_WAIT, the next test hitting the same port gets AddrInUse and fails with a misleading error that looks like a code bug rather than a port conflict.

This is the same problem that was fixed in stratum tests in PR #477. Same fix here,  pass "127.0.0.1:0" so the OS assigns a free port, then use the actual bound address from the return value.

16 tests using run_rpc_server now pass "127.0.0.1:0" and capture the returned SocketAddr instead of ignoring it. 4 tests using Server::builder().build() directly now call server.local_addr() after build. test_get_peer_info_rpc had two server instances inside one test function (ports 9008 and 9018) , both converted the same way.

Also removed 6 blind tokio::time::sleep calls that existed only to give the server time to start. run_rpc_server returns after the server is bound and ready, so the sleep was never necessary,  just slowing down the test suite:(

CC: @Sansh2356

### DIFF
--- a/node/src/rpc_server.rs
+++ b/node/src/rpc_server.rs
@@ -1213,10 +1213,9 @@ pub async fn test_extend_rpc() {
     let braid: Arc<RwLock<braid::Braid>> = Arc::new(RwLock::new(braid::Braid::new(genesis_beads)));
     let (proxy_tx, _) = mpsc::unbounded_channel();
 
-    let server_addr = "127.0.0.1:9101";
-    let _ = run_rpc_server(
+    let (server_addr, _) = run_rpc_server(
         Arc::clone(&braid),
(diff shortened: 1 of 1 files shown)


Sansh2356 (approved):
t-ACK [`8b86057`](https://github.com/braidpool/braidpool/pull/482/changes/8b86057ffcbba2bd303bcecd45367f2eebaefde1) .

mcelrath (approved):

PR #483, opened 2026-06-25
Removing `futures::Mutex` for `tokio::sync::Mutex`

Sansh2356:
Replace all `futures::lock::Mutex` imports and usages across the node crate with `tokio::sync::Mutex` so every async lock comes from a single, runtime-integrated primitive.

### DIFF
--- a/node/src/stratum.rs
+++ b/node/src/stratum.rs
@@ -6,7 +6,7 @@ use bitcoin::consensus::serialize;
 use bitcoin::io::Cursor;
 use bitcoin::{absolute::Decodable, Transaction};
 use bitcoin::{BlockHash, BlockHeader, BlockTime, TxMerkleNode, Txid, Witness};
-use futures::{lock::Mutex, FutureExt};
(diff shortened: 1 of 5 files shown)


nkatha23 (commented):
t-ACK 1a000d6  builds clean

node/src/stratum.rs
nkatha23, line 2003: In handle_connection here, downstream_client.lock().await is held across handle_client_to_server_request(...).await, a wide critical section that blocks any other task trying to lock downstream_client for the full duration of the request handling. Not introduced by this PR (pre-existing with futures::lock::Mutex too), but worth narrowing in a follow-up now that we're on tokio::sync::Mutex where the behavior difference is more visible.

Sansh2356: Yes, this issue i have already noted long time back i had made followed up PRs as well to address this but it is already present in other PRs which should be merged before this hence not addressing that in this PR or another separate PR see #443 .

PR #484, opened 2026-07-01
Fix(stratum): cap MiningJobMap per miner to prevent unbounded memory growth

nkatha23:
This is the first of five structural fixes identified for stratum scalability at 10k miners.
MiningJobMap stored every job ever created with no eviction. Two HashMaps grew indefinitely:

```rust
mining_jobs: HashMap<TemplateId, JobDetails>,     // never pruned
job_id_to_template: HashMap<u64, TemplateId>,     // never pruned
```

At bead rates of 150ms–1s ([roadmap.md](docs/roadmap.md): "beads will come in with a mean time between 150ms-1000ms"), a single miner connection generates up to 576,000 entries per day. Each `JobDetails` holds a full `BlockTemplate` including `Vec<Transaction>`,  actual size depends on mempool fullness but is realistically 100KB–500KB per entry with a populated mempool. With real miners connected the node runs out of memory and crashes within hours

Fix: add capacity: usize to MiningJobMap. Since next_job_id is monotonically increasing, the oldest surviving ID is always (next_job_id - capacity). On insert, evict that entry if the map is at capacity.

Added MAX_JOBS_PER_MINER = 10 constant in lib.rs alongside the existing MAX_CACHED_TEMPLATES = 90.

Two call sites updated: run_stratum_service (production) and the test helper.

Test: test_mining_job_map_eviction, inserts capacity+2 jobs,asserts oldest two are evicted, newest three remain accessible.

### DIFF
--- a/node/src/stratum.rs
+++ b/node/src/stratum.rs
@@ -1184,19 +1184,18 @@ pub struct JobDetails {
 ///Declaring as `Arc` object for shared reference across different tasks due to
 /// multiple threads serving requests according to the new process of serving requests .
 pub struct MiningJobMap {
-    // template_id to job details
     mining_jobs: HashMap<TemplateId, JobDetails>,
-    // numeric job_id to template_id
     job_id_to_template: HashMap<u64, TemplateId>,
-    // Generate sequential numeric job IDs for miners
     next_job_id: u64,
+    capacity: usize,
 }
 impl MiningJobMap {
-    pub fn new() -> Self {
+    pub fn new(capacity: usize) -> Self {
         Self {
             mining_jobs: HashMap::new(),
             job_id_to_template: HashMap::new(),
             next_job_id: 0,
+            capacity,
         }
     }
     ///Inserting a suitable mining job which has been passed to the downstream being constructed from a suitable block template .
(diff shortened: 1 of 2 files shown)


node/src/stratum.rs
```diff
@@ -1209,10 +1208,18 @@ impl MiningJobMap {
...
 
-        // Store job by template_id
-        self.mining_jobs.insert(template_id, job_details);
+        // Evict the oldest job when at capacity. next_job_id is monotonically
+        // increasing so the oldest surviving job id is always (next_job_id - capacity).
+        if self.job_id_to_template.len() >= self.capacity {
+            if let Some(oldest_id) = numeric_job_id.checked_sub(self.capacity as u64) {
+                if let Some(old_template) = self.job_id_to_template.remove(&oldest_id) {
+                    self.mining_jobs.remove(&old_template);
+                    debug!(evicted_job_id = %oldest_id, "Evicted oldest job from MiningJobMap");
```
nkatha23, line 1217: should this be `warn!` instead of `debug!`? Eviction is expected normal behavior so `debug!` seems right, 
but if eviction is happening frequently it could indicate the capacity is too low.

@Sansh2356  @zaidmstrr

mcelrath (changes requested):
This is a good catch but I don't think it's the right solution.

1. When a mining device is given a new work unit, we should expire any previous `mining_job` given to it.
2. There's a tremendous amount of duplication here. Within a single mine and 10k devices, There will in general usually only be 1-5 block templates in flight. (Not 1 because of interleaving jobs previously given to miners and new jobs generated when a new bead comes in)
3. Storing `Vec<Transaction>` is an optimization. It allows us to quickly assemble the block if a valid solution is found. In principle we could re-request the txs from cmempoold.

mcelrath:
MiningJobMap now caps by numeric job ID, but eviction deletes mining_jobs by template_id. That breaks when the same template is sent to one miner more than once:
- repeated non-submit requests can re-trigger latest-template resend at node/src/stratum.rs:313-325
- that resend path inserts the current template again at node/src/stratum.rs:1591-1686
- when the oldest alias is evicted, self.mining_jobs.remove(&old_template) deletes the shared template entry at node/src/stratum.rs:1215-1216
- newer job IDs that still point at that template then fail with MiningJobNotFound

So the PR fixes unbounded growth, but can incorrectly reject valid late submits.

PR #486, opened 2026-07-02
fix(miner-api): update IP and MAC handling in miner device management

### LINKED ISSUE #485: Bug: Duplicate Miner Records Created When IP Address Changes
## Description

Currently, miners are identified using their IP address. When a miner's IP changes (for example, due to DHCP lease renewal, router restart, or network reconfiguration), the dashboard treats it as a new device instead of recognising it as the existing miner.

As a result:

* A new database record is created for the same physical miner.
* The previous record remains associated with the old (unreachable) IP.
* The dashboard can no longer communicate with the original record.
* Users are forced to manually add the miner again.

This leads to duplicate entries for a single device and an inconsistent user experience.

## Suggested Solution

* Introduce a stable unique identifier for each miner (MAC address, serial number, UUID, etc.).
* Use this identifier as the primary lookup key during discovery and synchronisation.
* Treat the IP address as mutable metadata that can be updated whenever it changes.
* During synchronisation:

  * If the unique identifier already exists, update the stored IP.
  * Otherwise, create a new miner record.

priyashuu:
This PR improves miner device management by using the MAC address as the primary device identifier while treating the IP address as mutable. It resolves issues where miners receiving a new IP address (e.g., due to DHCP) were incorrectly treated as new devices, resulting in duplicate database entries.
### Workflow
add_miner(ip)
├── Device unreachable (offline)
│   ├── IP already in DB? → return already_exists
│   └── Create offline record (IP-only, no MAC yet)
│
└── Device online
    ├── No MAC returned → return error (hard stop)
    ├── MAC found in DB → update IP + fields on existing record
    └── MAC not in DB → create new record
        └── Race-condition duplicate MAC? → IntegrityError → return already_exists

fix #485

### DIFF
--- a/miner_api/db_models.py
+++ b/miner_api/db_models.py
@@ -10,7 +10,7 @@ class MinerDevice(Base):
     __tablename__ = "miner_devices"
     
     id = Column(String(36), primary_key=True, default=lambda: str(uuid.uuid4()))
-    ip = Column(String(45), nullable=False, index=True)  # IPv4 or IPv6
+    ip = Column(String(45), nullable=False, index=True)  # IPv4 or IPv6 — mutable, not unique
     name = Column(String(255), nullable=True)  # User-defined name for the miner
     
     # Device identification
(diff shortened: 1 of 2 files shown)


zaidmstrr:
Closing as the current requirement changes to use `asic-rs`

PR #490, opened 2026-07-04
db: read parent_timestamp as u32 to avoid 2038 overflow

### LINKED ISSUE #449: Possible overflow in `parent_timestamp` in db
According to this issue, the bead may panic with an overflow or will produce invalid entries of parent timestamps after Jan 19, 2038 03:14:08. The issue is in the type of `parent_timestamps` which is currently set to `i32` inside `db_handlers.rs`.

```sh
let parent_timestamp = parent_beads.get::<i32, _>("timestamp");
```
A simple fix of this is to change the type from `i32` to `u32`.

Atishyy27:
### What

In `fetch_beads_in_batch` (`node/src/db/db_handlers.rs`), `parent_timestamp` is read from the query result as `i32` and then cast back to unsigned with `ts as u32`:

```rust
row.try_get::<i32, _>("parent_timestamp"),
...
.push(MedianTimePast::from_u32(ts as u32).unwrap());
```

Since the column stores a Bitcoin block timestamp (unsigned seconds since the epoch), reading it as `i32` overflows for any timestamp at or after **2038-01-19 03:14:08**, which will panic or produce invalid parent timestamps.

### Change

Read the column as `u32` directly and drop the now-redundant `as u32` cast. This matches how the sibling timestamp columns in the same query are already read (`nTime`, `start_timestamp`, `broadcast_timestamp` all use `get::<u32, _>`).

### Notes

- Minimal, behavior-preserving for all in-range timestamps.
- I wasn't able to run the full build locally as I don't have the `capnp` toolchain set up on this machine; the change is a two-line type fix consistent with the surrounding code, so CI should cover the build/lint.

Fixes #449

### DIFF
--- a/node/src/db/db_handlers.rs
+++ b/node/src/db/db_handlers.rs
@@ -422,7 +422,7 @@ pub async fn fetch_beads_in_batch(
             }
             if let (Ok(parent_hash), Ok(ts)) = (
                 row.try_get::<Vec<u8>, _>("parent_hash"),
-                row.try_get::<i32, _>("parent_timestamp"),
+                row.try_get::<u32, _>("parent_timestamp"),
             ) {
                 if parent_hash.is_empty() {
                     //Genesis bead case
@@ -439,7 +439,7 @@ pub async fn fetch_beads_in_batch(
                 bead.committed_metadata
                     .parent_bead_timestamps
                     .0
-                    .push(MedianTimePast::from_u32(ts as u32).unwrap());
+                    .push(MedianTimePast::from_u32(ts).unwrap());
             }
         }
     }


zaidmstrr:
Please sign the commits.

Atishyy27:
@zaidmstrr
amended the commit with a `Signed-off-by` line and force-pushed. Ready for another look

Sansh2356:
> @zaidmstrr amended the commit with a `Signed-off-by` line and force-pushed. Ready for another look

Hey @Atishyy27 you will have to amend with rebase in interactive mode instead of `signoff` as both are different things, can have a look at git documentation .

Atishyy27:
Signature is sorted now. The commit (aee61b2) is signed with an SSH signing key and GitHub reports it as Verified, so the 'Commits must have verified signatures' rule is satisfied.

For context on what was actually wrong: the earlier confusion was between a DCO Signed-off-by line (just plain text) and a cryptographic signature. Those are two different things. On top of that, the repo's local .git/config had a user.signingkey pointing at a stale key that was not registered on my GitHub account, which silently overrode my global config, so the amend produced a signature GitHub could not verify (unknown_key). I pointed the config at the correct SSH signing key, re-signed with 'git commit --amend -S', and force-pushed. The GitHub commits API now returns verification.verified = true, reason = valid.

The diff itself is unchanged: still the two line u32 read fix for parent_timestamp. Ready for another look whenever you get a chance, thanks @zaidmstrr @Sansh2356.

zaidmstrr (approved):
ACK [aee61b2](https://github.com/braidpool/braidpool/pull/490/changes/aee61b240c8fd035b9ca38cf90055ed0943e74a3)

Thanks for your contribution.

PR #491, opened 2026-07-04
Testframework phase3

kushagra0902:
This PR advances #463 
It is phase 3 PR and sequentially after PR #468  and PR #476 

It includes:
- Test_Node.py for managing braidpool nodes for the test  framework
- Bitcoin_node: for managing bitcoin node
- Node_Manager: To handle all the binary nodes being used in a test case.
 
There are two test and example files included in the framework, feature_node_startup and feature_framework_unit_tests. 

Docs for running these are included in tests/functional/docs


zaidmstrr (commented):
This implementation is using RPC to communicate with a Bitcoin node while we need to talk via a UNIX socket in our case. And also don't think what can be the use of using `bitcoin-cli` here? You can explain if you think a usecase.

kushagra0902:
> This implementation is using RPC to communicate with a Bitcoin node while we need to talk via a UNIX socket in our case. And also don't think what can be the use of using `bitcoin-cli` here? You can explain if you think a usecase.

I am a bit confused here. The test site uses RPC to communicate with the Bitcoin Node just for setting up, like creating a wallet, let, generating addresses, or getting chain info, etc and uses these RPC calls in related functions like wait_ready. 
The Braidpool node that is running in the test suite is using IPC (UNIX Socket) communication.  I think the BTC node doesn't expose the setup functions mentioned above as an IPC layer. Please correct me if I understood something wrong. 

And yes, btc cli is dead code left by mistake; I will remove it.

zaidmstrr:
Yes, right now Bitcoin doesn't expose the functions you mentioned through IPC. And it's a part of the Chain interface which is currently not merged into Bitcoin core. I think until then we can use RPC for setting up the initial Bitcoin nodes, later if Bitcoin explicitly exposes those functions then we can move to pure IPC implementation. I have also tested that both the RPC and IPC functionality spawned in via the same node can work co-existingly. There are some extra new lines left in the `bitcoin_node.py`.

kushagra0902:
> Yes, right now Bitcoin doesn't expose the functions you mentioned through IPC. And it's a part of the Chain interface which is currently not merged into Bitcoin core. I think until then we can use RPC for setting up the initial Bitcoin nodes, later if Bitcoin explicitly exposes those functions then we can move to pure IPC implementation. I have also tested that both the RPC and IPC functionality spawned in via the same node can work co-existingly. There are some extra new lines left in the `bitcoin_node.py`.

So as of now, the implementation works, right? I will look onto bitcoin node file

zaidmstrr:
Yes, for now we can move with this.

tests/functional/test_framework/network_config.py
zaidmstrr: These 3 fields aren't used anywhere else in the code. As mentioned setting them will override the ports used by `NetworkManager` but the `NetworkManager.start_braidpool_nodes()` never read them. Can you tell me the possible reason for this or any future updates?

zaidmstrr (commented):
Approach ACK [02a8f51](https://github.com/braidpool/braidpool/pull/491/changes/02a8f51e3fa5ced516ae89ae7f228124c29f4b9f)

Most of the things seem correct to me. There is only one question I want to ask; otherwise, it's fine.

tests/functional/test_framework/network_config.py
kushagra0902: > These 3 fields aren't used anywhere else in the code. As mentioned, setting them will override the ports used by `NetworkManager`, but `NetworkManager.start_braidpool_nodes()` never reads them. Can you tell me the possible reason for this, or any future updates?

These were included with the thought that if the user wants to reserve the ports manually in some cases, but yes, these are not actually used as of now. I think I will remove these for now, and if needed in future, then will introduce them.

kushagra0902: Removed.

zaidmstrr (approved):
ACK [7a9902e](https://github.com/braidpool/braidpool/pull/491/changes/7a9902e3a22198784a892ba68378c1f56fab37e9)

PR #492, opened 2026-07-06
refactor(stratum): replace per-miner MiningJobMap with GlobalJobStore

nkatha23:
This is a rewrite of the job storage architecture, PR #484
Every miner connection had its own MiningJobMap. The notify loop cloned the full block template (~100–500 KB) once per miner on each new bead:

```rust
for (peer, map) in job_map_arc.iter() {
    let template_for_job = template.clone(); // ~100–500 KB per miner
    map.insert_mining_job(template_id, job_details).await;
}
```

At 10,000 miners that's 5GB allocated per new template. At 150ms bead rates the node OOM-crashes within hours.

The fix replaces all per-miner maps with a single GlobalJobStore. JobDetails is wrapped in Arc, all miners share one allocation. The notify loop builds the job once and sends N pointer copies:

```rust
// Build once, insert once
let job_details = Arc::new(JobDetails { ... });
let job_id = global_job_store.lock().await.insert(template_id, job_details);

// Send the same job_id to every connected miner, no template clone
for (peer, connection) in &connection_snapshot {
    send_mining_notify(peer, job_id, &notification);
}
```
Memory: ~5 GB → ~500 KB per template at 10k connections.

The resend path for newly connecting miners reuses the existing job_id for the current template (latest_job_id_for) instead of minting a new one. Minting per-connection advances next_job_id and eventually evicts the job_id every already-connected miner is submitting against enough reconnects inside one bead window would invalidate all in-flight shares. Reuse makes the notify loop the only job_id source in steady state. If the template was never stored (broadcast arrived while no miners were connected), the resend path builds and inserts normally.

When at capacity (GLOBAL_JOB_STORE_CAPACITY = 20, ~3s retention at bead rate), the oldest job_id is removed. Template data is only freed once no remaining job_id references it,  this fixes the correctness bug in #484 where evicting by template_id silently invalidated other job_ids pointing at the same template via the resend path.
Capacity note: dev's per-miner MAX_JOBS_PER_MINER was 10 (~1.5s retention). An earlier revision of this PR used 5, which halved it. The 10 was constrained by per-miner memory pressure; with one shared store that no longer applies, so 20 gives better slow-miner tolerance at negligible cost.

### DIFF
--- a/node/src/stratum.rs
+++ b/node/src/stratum.rs
@@ -271,7 +271,7 @@ impl DownstreamClient {
     pub async fn handle_client_to_server_request(
         &mut self,
         client_request: StandardRequest,
-        mining_job_map: Arc<Mutex<MiningJobMap>>,
+        global_job_store: Arc<Mutex<GlobalJobStore>>,
         response_message_sender: mpsc::Sender<String>,
         notification_sender: mpsc::Sender<NotifyCmd>,
         peer_addr: String,
@@ -310,7 +310,7 @@ impl DownstreamClient {
                 Self::handle_submit(
                     self,
                     &req_params,
-                    mining_job_map,
+                    global_job_store,
                     client_request_id,
                     swarm_handler,
                     audit_dag,
@@ -508,7 +508,7 @@ impl DownstreamClient {
     pub async fn handle_submit(
         &mut self,
         submit_work_params: &Value,
-        mining_job_map: Arc<Mutex<MiningJobMap>>,
+        global_job_store: Arc<Mutex<GlobalJobStore>>,
         client_request_id: u64,
         swarm_handler: Arc<Mutex<SwarmHandler>>,
         audit_dag: Option<Arc<Mutex<crate::audit::AuditDAG>>>,
@@ -640,7 +640,7 @@ impl DownstreamClient {
 
         //Acquiring lock on the mining map and fetching the submitted job from the memory
         let (submitted_job, template_id) = {
-            let job_mapping = mining_job_map.lock().await;
+            let job_mapping = global_job_store.lock().await;
 
             // Try string lookup first (upstream jobs)
             if let Ok((job, tid)) = job_mapping.get_by_string_job_id(job_id_str) {
@@ -649,7 +649,7 @@ impl DownstreamClient {
                     mode = "upstream",
                     "Found job by string ID"
                 );
-                (job.clone(), tid.clone())
+                (job, tid)
             } else {
                 // Fall back to numeric parsing (Braidpool jobs)
                 let numeric_job_id = match job_id_str.parse::<u64>() {
@@ -671,7 +671,7 @@ impl DownstreamClient {
                     mode = "braidpool",
                     "Found job by numeric ID"
                 );
-                let job = job_mapping.get_by_job_id(numeric_job_id)?.clone();
+                let job = job_mapping.get_by_job_id(numeric_job_id)?;
                 let templateid = job_mapping
                     .template_id_from_job_id(numeric_job_id)
                     .ok_or_else(|| StratumErrors::MiningJobNotFound {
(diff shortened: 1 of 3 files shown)


zaidmstrr (commented):
Approach ACK [798566d](https://github.com/braidpool/braidpool/pull/492/changes/798566d2de25d8baca195176d6cf917971ba7c58)

This PR introduces a great refactor and optimization. Approach wise this PR looks neat, and I have also reviewed the code very briefly. Will conduct a detailed review later on, but the changes of this PR are dependent on [PR #509](https://github.com/braidpool/braidpool/pull/509) so you need to rebase on top of that after merging, and most probably you need to make changes on the upstream code also because that PR uses the existing stratum infrastructure. So we can merge this after PR #509. And thanks for this change.

node/src/stratum.rs
```diff
@@ -2398,13 +2423,9 @@ impl Notifier {
...
             is_upstream_job: true,
         };
-        let upstream_job_id = curr_peer_mining_job_map
-            .insert_upstream_job(
-                job_notification.job_id.clone(),
-                template_id.clone(),
-                job_details,
-            )
-            .await;
+        job_store_guard.insert(template_id, Arc::new(job_details));
```
zaidmstrr: There is a duplication and template exhaustion issue here. When a miner subscribes or reconnects, it fetches the cached upstream job and calls `send_upstream_job_notification_to_miner()` and that helper function then calls

``` sh       
 job_store_guard.insert(template_id, Arc::new(job_details));
```
again for the same `Template::Upstream`. This duplicates entries inside `job_id_to_template` leading to disappearing the old jobs cache due to frequent miners reconnecting or joining.

zaidmstrr (commented):
The changes here look good, and you have successfully simplified the previous approach and reduced its complexity while eliminating the existing unbounded resource constraints bug. I have reviewed the code and found no major issues with the current approach, just a single issue for now you can look for.

zaidmstrr:
And I think the current cache limit of jobs which is `20` is a tradeoff between miner latency and braidpool acceptance. Setting this limit needs complex evaluation of the miner's interaction with the pool. But I think for now we can proceed with this. And I also think the cohort thickness can also affect this.

nkatha23:
> And I think the current cache limit of jobs which is `20` is a tradeoff between miner latency and braidpool acceptance. Setting this limit needs complex evaluation of the miner's interaction with the pool. But I think for now we can proceed with this. And I also think the cohort thickness can also affect this.

Agreed. 20 was chosen to give 3s retention at 150ms bead rate, which is enough buffer for slow miners without excessive memory pressure. we'll work on it when a different value is needed.

zaidmstrr:
One more thing I noticed is the current variable name `make_braidpool_job` is a little imprecise and does not properly distinguish between normal jobs and upstream jobs. The better alternative is `build_local_job_details` or anything else you can think of.

zaidmstrr:
> One more thing I noticed is the current variable name `make_braidpool_job` is a little imprecise and does not properly distinguish between normal jobs and upstream jobs. The better alternative is `build_local_job_details` or anything else you can think of.

I also observed that we are using Braidpool and upstream names as categorization in most of the places. I think we need to do a further refactor on this on some other PR. As these names might confuse the other contributors. So I'm going to merge this for now. But we can do the naming refactor in some other PR.

nkatha23:
> One more thing I noticed is the current variable name `make_braidpool_job` is a little imprecise and does not properly distinguish between normal jobs and upstream jobs. The better alternative is `build_local_job_details` or anything else you can think of.

 Done,  renamed `make_braidpool_job` to `build_local_job_details ` across the file. Agreed on deferring the broader braidpool/upstream naming refactor to a follow-up PR

zaidmstrr:
Code ACK [2d6a69e](https://github.com/braidpool/braidpool/pull/492/changes/2d6a69e2d997dffab969f87ded0c1760a6906d09)

Thanks for implementing all the suggestions. I have tested this PR using my S9 over audit mode, and it's working fine.

zaidmstrr (approved):

PR #493, opened 2026-07-06
tests: fix mutable default argument in all_ancestors

Atishyy27:
### What

`all_ancestors` and `all_ancestors_recursive` in `tests/braid.py` use a mutable default argument, `ancestors={}`. Python evaluates default arguments once at definition time, so every call that omits the argument shares one dict.

When callers reuse bead indices across different graphs — which `load_braid` does, since it renumbers every fixture's beads `0..n` — later calls find those indices already cached and return **stale ancestor sets** instead of recomputing. Concretely, `test_all_ancestors` in `test_braid.py` calls with the default dict; after the first fixture populates beads `0..49`, every later fixture finds those indices cached and skips recomputation, so most fixtures are no longer actually exercised.

### Change

- Switch both functions to the `ancestors=None` / `if ancestors is None: ancestors = {}` idiom so each defaulted call starts fresh. All internal callers already pass an explicit `ancestors` dict, so their behaviour is unchanged.
- Add a regression test (`test_default_ancestors_dict_is_isolated`) that calls both functions without an explicit dict on two different graphs and asserts no state leaks between calls. It fails on the shared-default behaviour and passes with the fix.

### Notes

- `tests/` unittest and pytest suites pass. (`test_simulator_api.py`'s websocket test errors on Windows due to signal handling — unrelated to this change.)
- I left `all_ancestors_recursive` delegating to the iterative traversal on purpose: `simulator.py` raises the recursion limit and notes that deep recursion can blow the stack on large cohorts, so making it genuinely recursive seemed like a separate, riskier decision better left to a maintainer.

### DIFF
--- a/tests/braid.py
+++ b/tests/braid.py
@@ -74,11 +74,13 @@ def generation(beads, children=None):
         retval |= children[b]
     return retval
 
-def all_ancestors_recursive(b, parents, ancestors={}):
+def all_ancestors_recursive(b, parents, ancestors=None):
     """ Gets all ancestors for a bead <b>, filling in ancestors of
         any other ancestors encountered, using a recursive
         algorithm.  Assumes b not in parents and b not in ancestors.
     """
+    if ancestors is None:
+        ancestors = {}
     ancestors[b] = set(copy(parents[b]))
     for p in parents[b]:
         if p not in ancestors:
@@ -86,11 +88,13 @@ def all_ancestors_recursive(b, parents, ancestors={}):
         ancestors[b].update(ancestors[p])
     return ancestors
 
-def all_ancestors(b, parents, ancestors={}):
+def all_ancestors(b, parents, ancestors=None):
     """ Gets all ancestors for a bead <b>, filling in ancestors of any other
         ancestors encountered, using an iterative algorithm. Assumes b not in
         parents and b not in ancestors.
     """
+    if ancestors is None:
+        ancestors = {}
     work_stack = [(b, False)]  # (bead, is_processed)
 
     while work_stack:
--- a/tests/test_braid_parametrized.py
+++ b/tests/test_braid_parametrized.py
@@ -150,6 +150,23 @@ class TestAllAncestors:
         # With distinct explicit dicts, bead 1 should only record bead 0.
         assert a2[1] == {0}
 
+    def test_default_ancestors_dict_is_isolated(self):
+        """Calls that omit the ancestors argument must not share state.
+
+        Regression test: all_ancestors/all_ancestors_recursive previously used a
+        mutable default argument (ancestors={}), so a single dict was reused
+        across every call. Repeated calls on different graphs then returned stale
+        ancestor sets cached from an earlier call.
+        """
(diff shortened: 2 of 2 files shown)


PragnyaKhandelwal:
> ### What
> `all_ancestors` and `all_ancestors_recursive` in `tests/braid.py` use a mutable default argument, `ancestors={}`. Python evaluates default arguments once at definition time, so every call that omits the argument shares one dict.
> 
> When callers reuse bead indices across different graphs — which `load_braid` does, since it renumbers every fixture's beads `0..n` — later calls find those indices already cached and return **stale ancestor sets** instead of recomputing. Concretely, `test_all_ancestors` in `test_braid.py` calls with the default dict; after the first fixture populates beads `0..49`, every later fixture finds those indices cached and skips recomputation, so most fixtures are no longer actually exercised.
> 
> ### Change
> * Switch both functions to the `ancestors=None` / `if ancestors is None: ancestors = {}` idiom so each defaulted call starts fresh. All internal callers already pass an explicit `ancestors` dict, so their behaviour is unchanged.
> * Add a regression test (`test_default_ancestors_dict_is_isolated`) that calls both functions without an explicit dict on two different graphs and asserts no state leaks between calls. It fails on the shared-default behaviour and passes with the fix.
> 
> ### Notes
> * `tests/` unittest and pytest suites pass. (`test_simulator_api.py`'s websocket test errors on Windows due to signal handling — unrelated to this change.)
> * I left `all_ancestors_recursive` delegating to the iterative traversal on purpose: `simulator.py` raises the recursion limit and notes that deep recursion can blow the stack on large cohorts, so making it genuinely recursive seemed like a separate, riskier decision better left to a maintainer.

Thanks for this. I checked it out and ran it locally.

**Verified**
- On `pr-493`, `test_braid.py` + `test_braid_parametrized.py`: 558 passed.
- I ran the new `test_default_ancestors_dict_is_isolated` against the `dev` version of `braid.py`. It fails (`assert set() == {0}` for `all_ancestors`), so it does catch the shared-default bug.
- I grepped `tests/*.py` for other mutable default arguments (`={}`, `=[]`, `=set()`) and found none left. The internal callers in `braid.py` (`cohorts`, `check_cohort_ancestors`, the call in the work computation) all pass an explicit `ancestors` dict, so their behaviour is unchanged.
- `python-unittest.yaml` runs `pytest test_*.py`, so CI picks up the new test.

LGTM. Two optional follow-ups that I think belong in separate PRs, as you noted:
1. `all_ancestors_recursive` calls `all_ancestors` (iterative) for the parents, not itself. So `test_all_ancestors` compares iterative code against a function that is only recursive at the top level. Worth an issue to decide whether to make it truly recursive or drop the pretence.
2. The existing `test_all_ancestors` passes both before and after this fix, so it wasn't protecting against the bug. Comparing against an independently computed ancestor set (a plain BFS over `parents`) would make it meaningful.
Tested on Windows / Python 3.11 with the pinned `tests/requirements.txt`. I didn't run the websocket simulator tests.

PR #495, opened 2026-07-08
db: validate persisted bead metadata on read to prevent crash-on-restart

### LINKED ISSUE #382: db: Persisting unvalidated UnCommittedMetadata can cause fragile persistence and crash-on-restart
### Description
The DB persistence layer writes `UnCommittedMetadata` to disk without validation and later reconstructs it using multiple `unwrap()` calls. This creates a fragile persistence invariant where invalid data can be permanently stored and cause node panics on reload.

---

### Affected code
- **Module:** `node/src/db/db_handlers.rs`
- **Branch:** `dev`

**Write path**
- `insert_bead` persists `bead.uncommitted_metadata` fields (signature, broadcast timestamp, extranonces) without validation.

**Read paths**
- `fetch_beads_in_batch`
- `fetch_bead_by_bead_hash`

Both reconstruct uncommitted metadata using `unwrap()` on:
- `Signature::from_slice`
- `MedianTimePast::from_u32`
- `u32::from_str_radix`

---

### Problem
The DB layer assumes persisted `UnCommittedMetadata` is always valid, but this assumption is not enforced at insertion time.

As a result:
- Invalid or partially constructed uncommitted metadata can be persisted.
- A single malformed row can cause panic during restart or batch fetch.
- The failure is delayed and non-recoverable because the invalid state is already on disk.

This is a persistence-layer correctness issue, not a consensus or security change.

---

### Scope and impact
- **Category:** Stability, robustness
- **Impact:** Crash-on-restart due to persisted invalid state
- **Consensus behavior:** Unchanged

---

### Suggested direction
- Validate `UnCommittedMetadata` before DB insertion.
- Fail fast at persistence time instead of committing invalid state.
- Replace `unwrap()` calls in DB fetch paths with structured error handling.

---

### Notes
The DB layer is the final persistence boundary and should explicitly enforce the invariants it relies on. Doing so prevents silent corruption and improves long-term node robustness.

Atishyy27:
## Problem

The DB read paths reconstruct a `Bead` from persisted rows using `unwrap()` on several fallible conversions:

- `Signature::from_slice`
- `MedianTimePast::from_u32`
- `u64::from_str_radix` (extranonces)
- prev/merkle hash `try_into()`

Because `insert_bead` writes `UnCommittedMetadata` without validating it, any partially-written or corrupted row becomes a permanent panic-on-reload: `fetch_beads_in_batch` and `fetch_bead_by_bead_hash` will `unwrap()` on the bad value and take the node down on startup. In `fetch_bead_by_bead_hash` there was also a latent bug where the inner parse `Result` was discarded by the outer `match`, so a malformed row silently returned a half-built default `Bead` instead of an error.

Closes #382.

## Change

- Add `build_bead_from_row()` — reconstructs the base bead and returns `DBErrors::TupleAttributeParsingError { error, attribute }` for every field instead of panicking.
- `fetch_beads_in_batch` now builds via that helper (through a `HashMap::entry` match so the fallible build can `?`-propagate) and validates the `txid` / parent-hash / parent-timestamp conversions too.
- `fetch_bead_by_bead_hash` propagates the parse error instead of swallowing it, and returns `Ok(None)` when no row matches.
- No remaining `unwrap()` in either reconstruction path, per the "no `unwrap()` in production code" rule in `CONTRIBUTING.md`.

Malformed persisted data is now a recoverable, logged error rather than a crash.

## Notes

- Pure error-handling change; the happy-path behaviour and DB schema are unchanged.
- `cargo fmt` applied. Full local `cargo check` isn't possible on Windows (the `async_zmq`/`mio::unix` dependency is Unix-only), so I relied on type-level review + CI to compile-verify; happy to adjust if CI surfaces anything.

### DIFF
--- a/node/src/db/db_handlers.rs
+++ b/node/src/db/db_handlers.rs
@@ -13,7 +13,9 @@ use bitcoin::{
 use futures::lock::Mutex;
 use num::ToPrimitive;
 use serde_json::json;
+use sqlx::sqlite::SqliteRow;
 use sqlx::{Pool, Row, Sqlite};
+use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use tokio::sync::mpsc::{Receiver, Sender};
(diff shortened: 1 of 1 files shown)


Atishyy27:
Closing this — it's been superseded on `dev`.

PR #478 ("Updating db handlers for orphan insertions and batched queries", merged 2026-07-09, `20070816b`) introduced a validating `build_bead_from_row` that returns `TupleAttributeParsingError` per field instead of unwrapping. It covers every attribute this PR was validating, including the uncommitted-metadata fields named in #382 (`broadcast_timestamp`, `extranonce1`/`extranonce2`, `signature`), and `db_handlers.rs` no longer carries the `unwrap()`/`expect()` calls that motivated the change.

I went to rebase this onto `dev` and the conflict was the tell: resolving it would have meant re-adding validation that #478 already provides, while fighting its batched-query refactor. Better to take #478's version.

@mcelrath — worth noting #382 is still open, but as far as I can tell `dev` now validates on read, so it may be closeable too.

Issue #417, opened 2026-03-22
rpc: bitcoinproxy has no auth and no method filtering


rx18-eng:
## Body

the rpc server in node/src/rpc_server.rs has no auth at all - run_rpc_server only adds a logging middleware. and the bitcoinproxy method just forwards whatever method + params you give it straight to bitcoin core using the node's stored credentials.

so any process on the machine can do something like:

```json
{"jsonrpc":"2.0","method":"bitcoinproxy","params":["dumpprivkey",["<address>"]],"id":1}
```

and get private keys out of bitcoin core. or sendtoaddress to drain funds.

## why it matters

right now its bound to 127.0.0.1:6682 so its localhost only, but theres a TODO   to make the bind address configurable. once someone binds to 0.0.0.0 without auth this becomes a critical vulnerability.  
   
## what i think we need

* auth on the rpc server
* a method allowlist on bitcoinproxy - only let through read-only stuff like getblockchaininfo, getmininginfo etc
* auth should be required not optional before bind address becomes configurable

not fully confident about how this fits into the bigger picture since braidpool is still early, but wanted to flag it before the rpc gets exposed.

rx18-eng:
@priyashuu @zaidmstrr , please let me know your thoughts on this !

Rustix69:
@rx18-eng For the Auth on the RPC Server #412

zaidmstrr:
No auth is a valid issue and already mentioned inside the codebase as the TODO. For the allowlist I think there should not be any problem because its the same user who is running the Bitcoin and Braidpool nodes. And after auth no other user can acess it so it will fix the issue.

zaidmstrr:
Closing this wrt [#502](https://github.com/braidpool/braidpool/issues/502).

PR #496, opened 2026-07-08
rpc: restrict bitcoinproxy to a read-only method allowlist

Atishyy27:
## Problem

`bitcoin_proxy` in `node/src/rpc_server.rs` forwards **any** method straight to Bitcoin Core using the node's stored RPC credentials, with only a logging middleware in front — no method filtering. Any process that can reach the RPC server can therefore run:

```json
{"jsonrpc":"2.0","method":"bitcoinproxy","params":["dumpprivkey",["<address>"]],"id":1}
```

to pull private keys out of Bitcoin Core, or `sendtoaddress` to move funds. It is bound to `127.0.0.1` today, but there is already a path toward making the bind address configurable, at which point this becomes a critical remote vulnerability.

Part of #417.

## Change

- Add `BITCOIN_PROXY_ALLOWED_METHODS` — an explicit allowlist of read-only, informational methods (chain/mempool/mining/network info, `getblocktemplate`, `estimatesmartfee`, `uptime`, …).
- Reject any method not on the list in `bitcoin_proxy` **before** it reaches Bitcoin Core, returning a clear JSON-RPC error and a `warn!` log.
- Matching is exact and case-sensitive (Bitcoin Core dispatch is case-sensitive, so casing can't be used to smuggle a method past the filter).
- Unit tests cover allowed methods, wallet/control methods (`dumpprivkey`, `sendtoaddress`, `importprivkey`, `walletpassphrase`, `stop`), and unknown/empty/mis-cased input, plus an allowlist-dedup check.

## Scope / follow-up

This closes the concrete key-exfiltration / fund-draining vector by adding the **method filtering** the issue asks for. Adding **authentication** to the RPC server (the other half of #417) is intentionally left as a separate follow-up so this stays atomic.

## Notes

`cargo fmt` clean. A full local `cargo check` isn't possible on Windows (`async_zmq`/`mio::unix` is Unix-only), so I verified types by hand and via the new unit tests; CI will compile-check.

### DIFF
--- a/node/src/rpc_server.rs
+++ b/node/src/rpc_server.rs
@@ -41,6 +41,45 @@ use {
     jsonrpsee::ws_client::WsClientBuilder,
 };
 
+/// Read-only Bitcoin Core RPC methods that `bitcoinproxy` is allowed to forward.
+///
+/// The proxy talks to Bitcoin Core using the node's own credentials, so without
+/// filtering any local caller could invoke wallet/control methods such as
+/// `dumpprivkey` (exfiltrate private keys) or `sendtoaddress` (drain funds).
+/// Only the informational, read-only methods a dashboard or miner needs are
+/// permitted; everything else is rejected before it reaches Bitcoin Core.
+const BITCOIN_PROXY_ALLOWED_METHODS: &[&str] = &[
+    "getblockchaininfo",
+    "getbestblockhash",
+    "getblock",
+    "getblockcount",
+    "getblockhash",
+    "getblockheader",
+    "getblockstats",
+    "getchaintips",
+    "getchaintxstats",
+    "getdifficulty",
+    "getmempoolinfo",
+    "getrawmempool",
+    "gettxout",
+    "gettxoutproof",
+    "gettxoutsetinfo",
+    "getmininginfo",
+    "getnetworkhashps",
+    "getblocktemplate",
+    "getnetworkinfo",
+    "getpeerinfo",
+    "getconnectioncount",
+    "getnodeaddresses",
+    "estimatesmartfee",
+    "uptime",
+];
+
+/// Returns `true` if `method` may be forwarded through `bitcoinproxy`.
+fn is_allowed_proxy_method(method: &str) -> bool {
+    BITCOIN_PROXY_ALLOWED_METHODS.contains(&method)
+}
+
 //server side trait to be implemented for the handler
 //that is the JSON-RPC handle to initiate the RPC context
 //supporting both http and websockets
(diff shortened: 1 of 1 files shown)


zaidmstrr (commented):
What I think is we should not need to call the Bitcoin IPC directly through the Braidpool node socket. As that socket is already used continuously for receiving and passing the import notifications and data between Braidpool and Bitcoin for e.g. new block templates, submitting of the Bitcoin block, and many other updates. We should not interrupt this process as a user for fetching general info from bitcoin. For this I have implemented [multiprocess RPC,](https://github.com/zaidmstrr/bitcoin-rpc-multiprocess) which allows the query of these methods over IPC. So we can use that in the future for such information and for the dashboard also. 

Apart from that, the Bitcoin schema implementation under `ipc/[client.rs](http://client.rs/)` doesn't implement any other methods that are required for mining purposes. This is also a reason we can't call any of those methods, so this is redundant here. I think we should remove any code from the `rpc_[server.rs](http://server.rs/)` which calls Bitcoin directly to get some info.

With this requesting to close this PR.

Atishyy27:
Thanks for the review @zaidmstrr — that's a fair point and I think you're right.

I approached #417 as "this proxy forwards arbitrary methods, so constrain it", but your framing is the better one: the question isn't *which* methods the proxy should forward, it's whether `rpc_server` should be reaching Bitcoin over the node socket at all. If that socket is already carrying block templates, submissions and notifications, then adding general info queries to it is contention we don't need — and an allowlist just makes a path we don't want safer to keep. Your `bitcoin-rpc-multiprocess` approach is clearly the right destination for those queries and for the dashboard.

Your second point also lands: if `ipc/client.rs` doesn't implement the mining methods, then most of what the proxy nominally exposes isn't reachable anyway, so hardening it is guarding dead code.

So I'm happy to close this. If the maintainers agree with the direction, I'd rather do the change you're actually describing — strip the direct-Bitcoin call paths out of `rpc_server.rs` — and I'm glad to open that PR. @mcelrath @pool2win, is that the direction you want, and should the removal wait on multiprocess RPC landing, or go in ahead of it?

Either way I'll follow your call rather than leave this one sitting.

zaidmstrr:
Please don't tag anyone without urgent need or for general queries.

Atishyy27:
Understood, and noted on the tagging — my apologies, I'll keep direction questions to Discord.

Closing this as you asked. Thanks for the steer on the architecture; the multiprocess RPC route makes more sense than hardening a path that shouldn't carry these queries in the first place.

zaidmstrr:
Thanks for understanding. You ask for direction questions here or on our Discord.

PR #497, opened 2026-07-09
IBD state management de-coupled from `PeerManager`

Sansh2356:
- Introduces a sans-I/O sync engine for Initial Block Download (IBD) under a new node::sync module.
- Decoupling sync state from `PeerManager/PeerInfo` and setting up to replace the current `ibd_manager` module. Single-peer sync target, wire protocol unchanged.
- Adding `Retry` engine for each of the request being timed out after specific intervals but capped with maximum number of retries to avoid retry loop increasing overload on the executing thread .

### DIFF
--- a/node/src/sync/mod.rs
+++ b/node/src/sync/mod.rs
@@ -1,0 +1,355 @@
+use crate::bead::{Bead, BeadHashes, BeadRequest};
+use crate::utils::BeadHash;
+use libp2p::PeerId;
+use std::collections::{HashMap, HashSet};
+use std::time::Duration;
+
+pub mod peer_state;
+pub mod retry;
+
+pub use peer_state::SyncPeerState;
+pub use retry::RetryPolicy;
+
+/// Maximum number of beads requested per `GetBeads` round-trip.
+pub const IBD_BATCH_SIZE: usize = 500;
+/// Maximum number of bead hashes a single `GetBeadsAfter` page may contain.
+/// A larger response is treated as a protocol violation.
+pub const IBD_HASH_PAGE_MAX: usize = 5_000;
+
+/// Where IBD currently is. Single-peer: at most one active sync target.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SyncState {
+    /// Not syncing. Either pre-trigger or between retries.
+    Idle,
+    /// Sent `GetTips`, awaiting the peer's tips.
+    AwaitingTips,
+    /// Sent `GetBeadsAfter`, awaiting a page of bead hashes.
+    FetchingHashes,
+    /// Downloading the beads for the current hash page in batches.
+    FetchingBeads,
+    /// IBD finished.
+    Complete,
+}
+
+/// Inputs to the engine, translated by the adapter from swarm responses, timers,
+/// and braid reads.
+#[derive(Debug, Clone)]
+pub enum SyncEvent {
+    /// Begin (or retry) IBD against a peer the adapter has already selected.
+    Start { peer: PeerId },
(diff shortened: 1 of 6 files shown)


node/src/sync/mod.rs
```diff
@@ -0,0 +1,355 @@
...
+            SyncEvent::BeadsReceived { peer, beads } => self.on_beads(peer, beads),
+            SyncEvent::RequestFailed { peer }
+            | SyncEvent::Timeout { peer }
+            | SyncEvent::PeerDisconnected { peer } => self.on_failure(peer),
+        }
+    }
+
+    fn on_start(&mut self, peer: PeerId) -> Vec<SyncAction> {
+        self.active = Some(SyncPeerState::new(peer));
+        self.state = SyncState::AwaitingTips;
```
nkatha23, line 168: on_start is the only handler with no guard,  it resets the engine unconditionally from any state, including mid-FetchingBeads. A late-firing retry timer would clobber an in-flight sync. Either add if self.state != SyncState::Idle && self.state != SyncState::Complete { return vec![]; } or add a doc comment documenting that the caller must never fire Start while active.  Which was intended?

node/src/sync/mod.rs
```diff
@@ -0,0 +1,355 @@
...
+
+/// The IBD protocol state machine.
+#[derive(Debug)]
+pub struct SyncEngine {
+    state: SyncState,
+    /// The single active sync target, if any.
+    active: Option<SyncPeerState>,
+    /// Consecutive IBD failures per peer. A peer reaching the retry ceiling is
+    /// surfaced via [`SyncEngine::exhausted_peers`] so the adapter excludes it
+    /// from sync-peer selection  according to MAX_LIMIT
```
nkatha23: No MAX_LIMIT constant exists anywhere in this file or the crate

node/src/sync/retry.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// Whether `attempts` consecutive failures have reached the ceiling.
+    pub fn is_exhausted(&self, attempts: u64) -> bool {
+        attempts >= self.max_retries
+    }
+
+    /// Basic Exponential backoff: `base * 2^attempts`, saturating and capped at `max`.
+    pub fn delay(&self, attempts: u64) -> Duration {
+        let factor = 1u64.checked_shl(attempts as u32).unwrap_or(u64::MAX);
+        let secs = self.base.as_secs().saturating_mul(factor);
+        let candidate = Duration::from_secs(secs);
```
nkatha23, line 40: as_secs() truncates sub-second bases: Duration::from_millis(500) produces a 0-delay hot loop on every retry. Imo, Use self.base.checked_mul(factor as u32).unwrap_or(self.max) instead,  it operates on Duration directly and preserves millisecond precision.

nkatha23 (commented):
Concept ACK 6a6d8f4

node/src/sync/mod.rs
Sansh2356: It won't be called more than once apart from retry which should reset the request initiation from `AwaitingTips` but it will shift the start of fetching the `beadhashes` from the current tip that has perhaps been updated from a previous failed attempt of IBD. A late firing retry timer won't be the case because the request timeout will be dependent on `NetworkBehaviour` configuration defined under `src/behaviours/mod.rs`

node/src/sync/mod.rs
Sansh2356: I will change the comment here i was referring to max_retries in the `RetryPolicy` configuration during the initialization  thanks for spotting this, i meant with max_limit as general not a defined constant  will update though .

PR #503, opened 2026-07-14
feat(stratum): add per-miner share counters (accepted/stale/invalid)

nkatha23:
Today the pool has no visibility into miner health. When a share is submitted, handle_submit decides whether it was accepted, stale, or invalid,  then discards that decision entirely. An operator cannot answer basic questions: which miners are slow? which are submitting bad work? what is the pool's overall error rate?
This PR adds per-miner share accounting to fix that.

What changes:

`DownstreamClient` gains a `ShareCounters` struct tracking three outcomes:

- accepted — share passed all validation and became a bead
- stale — valid share for a job that no longer exists (timing issue, not miner error)
- invalid — malformed params, bad hex, mask violation, PoW below target, or unauthorized submit

Counters are logged when a miner disconnects. 

Audit mode share accounting is unaffected.

### DIFF
--- a/node/src/stratum.rs
+++ b/node/src/stratum.rs
@@ -201,6 +201,15 @@ pub struct SuggestDifficultyResponse {
 }
 ///This will persist the client specific information for each of the new downstream connected
 /// to the stratum service which are setup during either `mining.subscribe` or `mining.configure` or `mining.authorize`
+
+/// Per-connection share accounting logged at disconnect.
+#[derive(Debug, Default, Clone, Copy)]
+pub struct ShareCounters {
+    pub accepted: u64,
+    pub stale: u64,
+    pub invalid: u64,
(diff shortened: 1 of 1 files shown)


nkatha23 (commented):
After rebasing onto dev following #509 :

Audit-mode submissions return early via validate_and_forward_upstream_share before any counter, so these counters cover the Braidpool submit path only. Audit mode has its own accounting via the audit DAG and MinerStatsView, though that view tracks valid beads only, so it doesn't give an error rate.

Dev's authorization gate now sits upstream of every bump, so an unauthorized mining.submit is counted nowhere. I've left it that way deliberately,  it reads as a protocol-state violation rather than a share-quality signal, but open to classifying it as invalid if reviewers:) see it differently.

PR #505, opened 2026-07-16
test: add missing assertion to reverse_cohorts_testcases

shivv23:
## Description

The reverse_cohorts_testcases test in node/src/braid/tests.rs computed a result but never asserted anything. This PR adds the missing assertion: cohorts computed from the reversed DAG must equal the original file cohorts in reverse order.

## Type of Change

- [x] Bug fix (non-breaking change fixing an incomplete test)

## Checklist

- [x] Tests pass (cargo test -p node -- braid::tests::reverse_cohorts_testcases)
- [x] No new warnings

### DIFF
--- a/node/src/braid/tests.rs
+++ b/node/src/braid/tests.rs
@@ -928,12 +928,11 @@ pub fn test_cohorts_braid_testcases() {
 }
 
 #[test]
-#[allow(unused)]
 pub fn reverse_cohorts_testcases() {
-    let ancestors = std::env::current_dir().unwrap();
-    let ancestors_directory: Vec<&Path> = ancestors.ancestors().collect();
-    let parent_directory = ancestors_directory[1];
-    let test_absolute_path = parent_directory.join(BRAIDTESTDIRECTORY);
+    let test_absolute_path = std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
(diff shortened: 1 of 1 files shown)


Sansh2356:
@shivv23 please add signatures to the commit .

nkatha23 (commented):
Good idea, wrong execution, the assertion logic is correct but the test runs zero iterations locally so it never actually checks anything. Fix the file path resolution so the loop runs at least once

shivv23:
Hi @nkatha23! I have made some fixes, kindly review them.
@Sansh2356 Hi! The commits are now verified, do have a look.

nkatha23 (approved):
t-ACK c8dff92

Thanks for the path fix,  CARGO_MANIFEST_DIR is the right approach.

Sansh2356 (approved):
t-ACK [c8dff92](https://github.com/braidpool/braidpool/pull/505/changes/c8dff9205f2739a1a7b14fe0ab69b312206f0139) .

zaidmstrr (approved):
Tested ACK [c8dff92](https://github.com/braidpool/braidpool/pull/505/changes/c8dff9205f2739a1a7b14fe0ab69b312206f0139)

The new approach seems good with the correct `dir` path.

PR #506, opened 2026-07-18
feat: add dynamic extranonce selection on upstream pool

zaidmstrr:
This PR adds the functionality to choose the dynamic upstream extranonce size. Which means now any upstream pool that either supports 4 bytes or 8 bytes as the extranonce1 size can connect with the Braidpool as an auditing layer. The Braidpool node will readjust itself and further call to the miner and submit the request back to the upstream pool internally with respect to the upstream extranonce size. As seen in many pools, the already defined ranges are the most common ones (e.g. 4 bytes of extranonce1 and 8 bytes of extranonce2 or 8 bytes of extranonce1 and 8 bytes of extranonce2); any other values or sizes are not common and maybe hardly used by any pool. Thus, users can treat it like the maximum cap bytes they can use is 16 bytes where `total_extranonce_bytes - extranonce1_bytes >= 8` to make it work.

### DIFF
--- a/node/src/audit.rs
+++ b/node/src/audit.rs
@@ -199,29 +199,35 @@ impl AuditCommitment {
         hex::encode(&self.commitment_bytes)
     }
 
-    pub fn verify_in_extranonce1(&self, extranonce1_bytes: &[u8], miner_prefix: &[u8]) -> bool {
-        if extranonce1_bytes.len() != TOTAL_EXTRANONCE1_BYTES {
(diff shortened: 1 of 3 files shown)


Sansh2356 (approved):
Code ACK [`778aceb`](https://github.com/braidpool/braidpool/pull/506/changes/778acebcd5650bc0306cde1dfc07a8a84fb03bcd) and tested as well no nits required LGTM .

PR #508, opened 2026-07-20
Test: fix shared DB path causing intermittent CI failures

### LINKED ISSUE #504: Test: shared ~/.braidpool/braidpool.db breaks test isolation (intermittent CI failures)
server_start_test fails intermittently on CI with two distinct errors depending on run scheduling:

database is locked (code: 5) — seen on #479's CI run
table Bead already exists (code: 17) — seen on #503's CI run

Root cause: every test that constructs DBHandler::new() reads/writes the same fixed path (~/.braidpool/braidpool.db). There is no per-test isolation, and schema creation is not atomic across concurrent test tasks,  two tests can both observe the missing table and both issue CREATE TABLE, the second erroring. Local repro supports this: with a fresh path, single-test and full-suite runs pass, and a second full run also passes, so it's a within-run race, not a rerun-idempotency problem.

This will get worse as the suite grows :(

Suggested fixes, in increasing order of thoroughness:

CREATE TABLE IF NOT EXISTS in schema execution (makes the race harmless, one-line)
Per-test temp paths (e.g. tempfile::TempDir) — real isolation
In-memory SQLite (sqlite::memory:) for unit tests — isolation + speed

nkatha23:
server_start_test and other stratum tests fail intermittently on CI. All stratum tests call DBHandler::new() which writes to a shared file path (~/.braidpool/braidpool.db). With parallel test execution, two tests race at schema creation,  one errors with table Bead already exists or database is locked.

changes:
schema.sql:  IF NOT EXISTS on all CREATE TABLE, CREATE INDEX, CREATE VIEW. Eliminates the already-exists race for any file-based test.
DBHandler::new_in_memory():  new #[cfg(test)] constructor using sqlite::memory:. Schema embedded via include_str!, no file path dependency, works on any machine.
stratum.rs:  all six DBHandler::new() callsites in tests switched to new_in_memory(). Each test now gets a private isolated database.

Fixes #504

### DIFF
--- a/node/src/db/db_handlers.rs
+++ b/node/src/db/db_handlers.rs
@@ -84,6 +84,45 @@ impl DBHandler {
             db_handler_tx,
         ))
     }
+
+    #[cfg(test)]
+    pub async fn new_in_memory() -> Result<(Self, Sender<BraidpoolDBTypes>), DBErrors> {
+        use sqlx::{sqlite::SqliteConnectOptions, SqlitePool};
+        use std::str::FromStr;
+        const SCHEMA: &str = include_str!("schema.sql");
+
+        let pool_options = SqliteConnectOptions::from_str("sqlite::memory:")
+            .map_err(|e| DBErrors::ConnectionUrlNotParsed {
+                error: e.to_string(),
+                url: "sqlite::memory:".to_string(),
+            })?
+            .foreign_keys(true)
+            .journal_mode(sqlx::sqlite::SqliteJournalMode::Wal);
+
+        let db_connection_pool = SqlitePool::connect_with(pool_options).await.map_err(|e| {
+            DBErrors::ConnectionToSQlitePoolFailed {
+                error: e.to_string(),
+            }
+        })?;
+
+        sqlx::query(SCHEMA)
+            .execute(&db_connection_pool)
+            .await
+            .map_err(|e| DBErrors::SchemaNotInitialized {
+                error: e.to_string(),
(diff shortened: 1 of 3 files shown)


Sansh2356:
@nkatha23 can you then also either replace the `test_db_initializer` with this or move the existing function to test utils and used that because 2 different functions for same purpose is not required due to DRY principle and also update the test functions in db_handlers with the same , a small suggestion .

nkatha23:
> @nkatha23 can you then also either replace the `test_db_initializer` with this or move the existing function to test utils and used that because 2 different functions for same purpose is not required due to DRY principle and also update the test functions in db_handlers with the same , a small suggestion .

Done, removed test_db_initializer() and updated test_batch_insertion_beads to use new_in_memory() directly, pulling the pool via db_connection_pool for the raw-query assertions. Also cleaned up the now-unused imports.

Sansh2356 (dismissed):
Code ACK and t-ACK [`52bfaff`](https://github.com/braidpool/braidpool/pull/508/changes/52b4aff7149d524eb3d26d8f6effad7abf222f62) .

Sansh2356:
@nkatha23 can you rebase this ?

zaidmstrr (approved):
Code ACK [1fa26f1](https://github.com/braidpool/braidpool/pull/508/changes/1fa26f1c3e303ab922ab5cfa55802779f9df5d08)

The changes look correct.

PR #517, opened 2026-07-27
db: read parent timestamp as u32 in fetch_bead_by_bead_hash

### LINKED ISSUE #449: Possible overflow in `parent_timestamp` in db

### LINKED ISSUE #516: fetch_bead_by_bead_hash still reads parent timestamp as i32 (2038 overflow, panics on decode)
### Summary

`node/src/db/db_handlers.rs` reads the parent bead timestamp as `i32` inside `fetch_bead_by_bead_hash`:

```rust
let parent_timestamp = parent_beads.get::<i32, _>("timestamp");
...
let parent_ts = MedianTimePast::from_u32(parent_timestamp as u32).map_err(|e| { ... })?;
```

This is the same class of bug as #449. That issue was closed by #490, but #490 only changed the sibling batch path (`fetch_beads_in_batch`, `parent_timestamp` -> `u32`); the single-bead lookup here still uses `i32`.

### Why it matters

Timestamps are unsigned. `sqlx::Row::get::<T, _>()` panics if the stored value cannot be decoded into `T`, so any parent bead with a timestamp at or beyond 2147483648 (2038-01-19) panics here rather than returning a `DBErrors`. The panic happens on the `get` call, so the `.map_err(...)` added below it never runs. The trailing `as u32` cast is also the silent-truncation pattern that #490 removed from the batch path.

Every other timestamp column in this file is already read as `u32` (`start_timestamp`, `broadcast_timestamp`, `nTime`).

### Fix

Read the column as `u32` and drop the cast, matching `fetch_beads_in_batch`. I have the change ready and will open a PR referencing this issue.

Atishyy27:
Fixes #516.

### What

`fetch_bead_by_bead_hash` read the parent bead timestamp with `get::<i32, _>("timestamp")` and then cast it with `as u32`. Timestamps are unsigned, so a value at or beyond 2038-01-19 cannot be decoded into `i32` — `sqlx::Row::get` panics at that point, before the surrounding `map_err` can turn it into a `DBErrors`.

### Why it was missed

#490 fixed exactly this for the batch path (`fetch_beads_in_batch`) and closed #449, but #449 actually quoted the line in `fetch_bead_by_bead_hash`, which was never changed. This finishes that fix.

### Change

Read the column as `u32` and drop the `as u32` cast — matching `fetch_beads_in_batch` and every other timestamp column in this file (`start_timestamp`, `broadcast_timestamp`, `nTime`), which already use `u32`.

Two lines, no behaviour change for in-range timestamps.

### DIFF
--- a/node/src/db/db_handlers.rs
+++ b/node/src/db/db_handlers.rs
@@ -742,7 +742,7 @@ pub async fn fetch_bead_by_bead_hash(
             }
         };
     for parent_beads in parent_timestamp_rows {
-        let parent_timestamp = parent_beads.get::<i32, _>("timestamp");
+        let parent_timestamp = parent_beads.get::<u32, _>("timestamp");
         let parent_bead_id = parent_beads.get::<i64, _>("parent");
         //Fetching parent_bead from DB
         let parent_bead_hash_raw_bytes = match sqlx::query("SELECT  hash FROM Bead WHERE id = ?")
@@ -768,7 +768,7 @@ pub async fn fetch_bead_by_bead_hash(
             }
         };
         //Extending parent bead timestamp
-        let parent_ts = MedianTimePast::from_u32(parent_timestamp as u32).map_err(|e| {
+        let parent_ts = MedianTimePast::from_u32(parent_timestamp).map_err(|e| {
             DBErrors::TupleAttributeParsingError {
                 error: format!("Invalid parent timestamp value {}: {}", parent_timestamp, e),
                 attribute: "parent_bead_timestamps".to_string(),


zaidmstrr (approved):
Code ACK [6d8406f](https://github.com/braidpool/braidpool/pull/517/changes/6d8406fa396f181cffaf1bae64df392a65f8f875)

Thanks for spotting this.

PR #519, opened 2026-07-29
Replace simulator with live node WebSocket RPC

priyashuu:
The dashboard DAG visualizer previously relied on a local simulator to generate bead/cohort data. This PR replaces that with direct WebSocket JSON-RPC calls to the running braidpool node, making the dashboard reflect real node state.

### Changes

- Removed simulator dependency; all DAG data (`getbraidinfo`, `getcohortbyid`, `getparents`, `gethighestworkpathbycount`) is now fetched from the node over a persistent WebSocket JSON-RPC connection
- Added `subscribebead` subscription so the graph updates automatically when the node mines or receives a new bead
- Implemented exponential backoff reconnection (1 s → 30 s) with a disconnection banner shown when the node is unreachable
- Fixed cohort window switching: stale beads are cleared immediately on selection change and in-flight fetches are cancelled via a generation counter (`fetchGen`) to prevent race conditions
- Fixed viewport reset when switching to a larger cohort window by only persisting user-initiated zoom/pan events (`event.sourceEvent` check) — programmatic auto-fit calls no longer corrupt the stored transform
- Increased bead label font size (`40` → `55`) and horizontal spacing between beads (`COLUMN_WIDTH` `400` → `600`) for improved readability

- Updated the `limit` parameter in `GetHighestWorkPathByCount` command from `u8` to `u64` in the CLI.
- Removed the `fetchbeadinfo.js` file and its associated functionality from the dashboard API.
- Enhanced the `BraidPoolDAG` component to handle WebSocket connections more robustly, including connection status management and error handling.
- Updated the `rpc_server.rs` to ensure the limit parameter is handled correctly and to improve error messages.
e fetchBeadInfo, and update WebSocket handling


node/src/rpc_server.rs
```diff
@@ -766,33 +765,19 @@ impl RpcServer for RpcServerImpl {
             ));
         }
 
-        // Validate that limit is greater than zero
-        if requested_limit == 0 {
+        if limit == 0 {
             return Err(ErrorObjectOwned::owned(
                 6,
-                format!(
```
zaidmstrr, line 774: Why have you removed these error logs?

priyashuu: Removing the detailed error messages was unintentional, so I’ve restored them.

The change from `.take(requested_limit)` to `.skip(skip_count)` was intentional, since `highest_work_path()` returns beads from oldest to newest (genesis → tips). Using `.take(N)` would return the oldest beads, while `.skip()` correctly gets the most recent beads from the tip side.

I’ve kept this change while restoring the original error messages. Thanks

PR #522, opened 2026-08-05
fix(node): drop peer connection instead of panicking on a malformed message

Atishyy27:
### what

`message_received()` in `node/src/connection.rs` decoded peer bytes with `Message::from_bytes(...).unwrap()`. a peer sending anything that fails to deserialize would panic and take down the whole node.

### why

the bytes here come straight off the network and are fully untrusted. `message_received()` already returns `Result<(), &'static str>` and the read loop closes the connection whenever it gets an `Err`, so the fix is to return an error instead of unwrapping. a malformed message now just drops that one peer.

### how tested

added `it_returns_err_on_malformed_bytes` in `protocol.rs`: `from_bytes` returns `Err` on garbage bytes and on empty input. verified the underlying flexbuffers decode rejects both cases, and a valid round-trip still decodes. CI compiles the node crate and runs the test on linux.

### risk

low. only turns a panic on bad input into a dropped connection. no change for valid messages.

### DIFF
--- a/node/src/connection.rs
+++ b/node/src/connection.rs
@@ -69,7 +69,10 @@ impl Connection {
     async fn message_received(&mut self, message: &Bytes) -> Result<(), &'static str> {
         use futures::SinkExt;
 
-        let message: Message = protocol::Message::from_bytes(message).unwrap();
+        // A peer can send arbitrary bytes; a failed decode must drop this
+        // connection, not panic the whole node.
+        let message: Message = protocol::Message::from_bytes(message)
+            .map_err(|_| "failed to deserialize peer message")?;
         match message.response_for_received() {
             Ok(result) => {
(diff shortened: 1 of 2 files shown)


priyashuu:
Hey! The latest changes have been added to the dev branch. Please base your work on dev and ensure your branch is up to date with dev, not main.

Atishyy27:
Closing in favour of #524.

The `dev` refactor removed the `connection.rs`/`protocol.rs` decode path this PR hardened, so it can't be rebased onto dev. #524 fixes the same class of bug — a malformed peer message panicking the node — against dev's current decode path (`CommittedMetadata`/`UnCommittedMetadata::consensus_decode`, which were `.unwrap()`ing untrusted input). Thanks @priyashuu for the pointer to base on dev.

Issue #514, opened 2026-07-23
Fix Vite CJS Node API deprecation warning by migrating to ESM configuration


priyashuu:
**Description:**

When starting the dashboard in development mode, Vite prints the following warning:

```bash
npm run dev

> dashboard@0.1.0 dev
> vite

The CJS build of Vite's Node API is deprecated.
See https://vite.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
```

### Current Behavior

The dashboard starts successfully, but Vite emits a deprecation warning indicating that the CommonJS (CJS) Node API is being used.

### Expected Behavior

The dashboard should start without deprecation warnings by using Vite's ESM-based configuration and APIs.

### Possible Cause

The project is likely using a CommonJS configuration or importing Vite through `require()` somewhere, for example:

* `vite.config.js` using CommonJS syntax
* `require("vite")`
* Other build scripts relying on the deprecated CJS API

### Proposed Solution

* Investigate what is causing Vite to use the CJS Node API.
* Migrate the project to ESM where appropriate.
* Update configuration files (e.g., `vite.config.mjs` or `vite.config.ts`) if necessary.
* Ensure all Vite imports use ESM syntax (`import`) instead of `require`.
* Verify that the warning no longer appears when running `npm run dev`.

### Acceptance Criteria

* [ ] Running `npm run dev` starts the dashboard without the Vite CJS Node API deprecation warning.
* [ ] The dashboard continues to function as expected after the migration.
* [ ] Any required documentation or configuration updates are included.

chahat-101:
Hey there, I can take this on! 

**Proposed Solution:**
* Refactor `vite.config.js` to strictly use ESM (`import`/`export`).
* Rename to `vite.config.mjs` if the workspace doesn't enforce `"type": "module"`.
* Verify `npm run dev` boots cleanly without the CJS deprecation warning.

Could you assign this to me so I can submit a PR?

priyashuu:
Yeah sure

chahat-101:
Looked into this on dev. The warning doesn't reproduce on a clean install — rm -rf node_modules && npm ci, then npm run dev runs clean on the pinned Vite 6.2.6. On Vite 6 the config's vite import resolves to the ESM build, so index.cjs (where the warning comes from) never loads. It only fires on an older Vite 5.x, so this was probably a stale node_modules — a reinstall fixes it.

Don't think there's anything to migrate here. Probably worth closing as not-reproducible.

[...]

PR #523, opened 2026-08-06
Remove unused tailwind.config.js from dashboard

chahat-101:
## Summary

Removes the unused `dashboard/tailwind.config.js`.

The project uses Tailwind CSS v4 with the `@tailwindcss/vite` plugin, so this config file is no longer used. There is no `@config` directive referencing it, making it effectively dead code.

The file also uses `module.exports` even though the package is configured as ESM (`"type": "module"`). While it isn't currently loaded, keeping it around could cause confusion or parsing errors if it ever is.

## Background

While looking into #514, I noticed the reported Vite CJS deprecation warning could not be reproduced on the pinned Vite version (6.2.6).

The project is already using ESM:
- `package.json` has `"type": "module"`
- `vite.config.ts` is ESM

A clean install starts the dev server without any CJS deprecation warnings. The only CJS-related issue I found was this unused Tailwind config, which this PR removes.

## Verification

- `vite build` produces identical CSS before and after removing the file (same `index-*.css` hash, 53.46 kB).
- `npm run dev` starts without any CJS deprecation warnings.

## Note

`npm run build` (`tsc && vite build`) currently fails on `main` due to existing TypeScript errors unrelated to this change. This PR does not modify `src/` or address those issues.

### DIFF
--- a/dashboard/tailwind.config.js
+++ b/dashboard/tailwind.config.js
@@ -1,8 +1,0 @@
-/** @type {import('tailwindcss').Config} */
-module.exports = {
-  content: ['./index.html', './src/**/*.{js,ts,jsx,tsx}'],
-  theme: {
-    extend: {},
-  },
-  plugins: [],
-};


priyashuu:
Hey! The latest changes have been added to the dev branch. Please base your work on dev and ensure your branch is up to date with dev, not main.

chahat-101:
Rebased this onto `dev` as requested. Since `dashboard/tailwind.config.js` doesn't exist on `dev` (it was only on `main`), there's nothing to delete, so this PR ends up being empty. I'll close it.

The original finding from #514 still stands though: I couldn't reproduce the Vite CJS deprecation warning on the pinned Vite 6.2.6. `dev` already uses ESM (`"type": "module"` and an ESM config), and a clean `npm ci` starts the dev server without any warning. It was likely caused by a stale `node_modules` with an older Vite version.

PR #525, opened 2026-08-12
 fix(stratum): parse ntime and nonce before coinbase reconstruction

nkatha23:
ntime and nonce are extracted from the submit params but not parsed until (180 lines) later, after job lookup (mutex acquisition), full coinbase hex decode, Transaction::consensus_decode, merkle branch decoding, and calculate_merkle_root. Neither parse depends on any of that work, so a submission with unparseable ntime pays the entire validation cost before being rejected.

Moves both u32::from_str_radix blocks to immediately after the audit-mode early return, they can't move above it since that path forwards the raw strings upstream.

No change to what is accepted or rejected. Deliberately not adding strict length validation: a miner formatting with {:x} rather than {:08x} sends "3" for nonce 3, which from_str_radix parses correctly. A width check would reject shares that are currently valid work.

Also points the extranonce2 length error message at miner_extranonce2_size rather than extranonce2_len. Both fields hold the same value at every assignment site today, but the message computes expected length from one and reports the byte count from the other, correct by coincidence rather than construction.

### DIFF
--- a/node/src/stratum.rs
+++ b/node/src/stratum.rs
@@ -4584,28 +4584,26 @@ mod test {
     }
 
     #[tokio::test]
-    async fn unpadded_nonce_is_accepted_by_parser() {
-        // A miner formatting nonce with {:x} instead of {:08x} sends "3" for nonce 3.
-        // from_str_radix parses this correctly — the value is right, just unpadded.
-        // This test pins that behaviour: strict 8-char length validation would
-        // reject valid work. The submit reaches PoW check (returns Ok(false)) rather
-        // than failing at the parse step (which would return Err).
+    async fn unpadded_nonce_not_rejected_by_length_check() {
+        // Pins the decision not to enforce strict 8-char width on nonce.
+        // A miner using {:x} formatting sends "3" for nonce 3 — correct value, just unpadded.
+        // A length check would reject valid shares like this one.
         let (mut client, map, swarm, job_id) = submit_setup().await;
         let params = json!([
             "miner",
             job_id.to_string(),
-            "0000000000000000", // extranonce2
-            "68df7e33",         // valid ntime
-            "3",                // nonce "3" — unpadded, correct value
+            "0000000000000000",
+            "68df7e33",
+            "3", // unpadded nonce, correct value
         ]);
         let result = client
             .handle_submit(&params, map, 1, swarm, None, None, None)
             .await;
-        // Must NOT be a parse error. PoW will not be met (nonce=3 at any real difficulty),
-        // so we expect Ok(false) — the submit got through the parse and into validation.
+        // Passes the parse step; fails later at coinbase decode (InvalidCoinbase),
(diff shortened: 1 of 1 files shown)


zaidmstrr (commented):
Approach ACK [19c5eb2](https://github.com/braidpool/braidpool/pull/525/changes/19c5eb269e6dba2ecf05db2840605569ef20310d)

The changes proposed by this PR are correct and valid. While I don't think so, the tests introduced by this PR are very relevant here. As the test named `unpadded_nonce_is_accepted_by_parser` just checks the behaviour of `from_str_radix` which is already part of the standard library and stable. Similarly, the test `non_hex_ntime_returns_err_before_coinbase_work` just checks the error type. Which is also obvious, and reverting the current changes will make no impact on the current test output behaviour. So, I think you should just keep the current changes and remove the new tests.

nkatha23:
> The changes proposed by this PR are correct and valid. While I don't think so, the tests introduced by this PR are very relevant here. As the test named `unpadded_nonce_is_accepted_by_parser` just checks the behaviour of `from_str_radix` which is already part of the standard library and stable. Similarly, the test `non_hex_ntime_returns_err_before_coinbase_work` just checks the error type. Which is also obvious, and reverting the current changes will make no impact on the current test output behaviour. So, I think you should just keep the current changes and remove the new tests.

Actually the tests are doing different job than testing `from_str_radix` itself.

The `non_hex_ntime_returns_err_before_coinbase_work` test is mainly checking the error mapping/ordering in `handle_submit`. With the old ordering, the test reaches coinbase reconstruction first and returns a different error, so the test does distinguish the old and new behaviour.
For `unpadded_nonce_is_accepted_by_parser`, the intention is to pin the decision not to enforce strict nonce length. A miner using `{:x}` can send an unpadded value like `3`, which is valid for the current behaviour. I’d rather have a test protecting that decision so a future length check doesn’t accidentally reject those shares.

I agree the second test name makes it sound like it’s testing the stdlib parser, though. I’ll rename it to make the intent clearer and the comments as well.

zaidmstrr:
Okay, so the test `non_hex_ntime_returns_err_before_coinbase_work` returns a different error message than the current one, and then it will be fine. Sorry for missing that one. And on the second `unpadded_nonce_is_accepted_by_parser` I think yes, changing the doc is more useful as from just reading it sounds a little but different than intended.

zaidmstrr:
Also please rebase this branch on top of the latest dev.

Sansh2356:
t-ACK [`c7d9ee6`](https://github.com/braidpool/braidpool/pull/525/changes/c7d9ee68ba60b66f68773cd30f163082ca6db9ca) .

zaidmstrr (approved):
Code ACK [c7d9ee6](https://github.com/braidpool/braidpool/pull/525/changes/c7d9ee68ba60b66f68773cd30f163082ca6db9ca)

Thanks for the changes. Now looks good.

PR #531, opened 2026-08-28
Worker name parsing for payout address of miner

Sansh2356:
- The worker_name received from miners serve the purpose of getting miners payout address during `mining.submit` for each submission .
- Parsing it and passing it onto the bead formation in `lib.rs` .

### DIFF
--- a/node/src/stratum.rs
+++ b/node/src/stratum.rs
@@ -1,6 +1,7 @@
 use crate::config::PoolNetwork;
 use crate::error::StratumErrors;
 use crate::template_creator::calculate_merkle_root;
+use crate::utils::validate;
 use crate::utils::compute_block_hash;
 use crate::{SwarmHandler, TemplateId, EXTRANONCE1_SIZE, EXTRANONCE2_SIZE, EXTRANONCE_SEPARATOR};
 use bitcoin::absolute::Time;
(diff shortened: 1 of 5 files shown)


Sansh2356:
To be merged after #398 .

node/src/utils/mod.rs
```diff
@@ -42,6 +49,87 @@ pub fn compute_block_hash(block_header: &BlockHeader, network: PoolNetwork) -> B
...
+                    });
+                }
+            }
+
+            if address_part.len() > 50 {
+                return Err(StratumErrors::UserNameParseError {
+                    error: crate::error::UsernameValidationError::InvalidAddress {
+                        address: address_part.to_string(),
+                    },
+                });
```
nkatha23, line 100: why 50? The length check  rejects valid addresses. 
Bitcoin P2WSH (bech32) addresses are up to 62 characters, so this fails valid submissions..  Either raise to 62 or remove the check entirely

node/src/stratum.rs
```diff
@@ -544,6 +545,19 @@ impl DownstreamClient {
             Ok(name) => name,
             Err(error) => return Err(error),
         };
+        //Parsing payout address from worker name
+        let payout_address = match validate(worker_name, self.network) {
+            Ok((address, _worker)) => address.to_string(),
```
nkatha23, line 550: the worker name is being silently discarded here. The worker name is useful for per-worker share accounting,  if a miner connects multiple rigs under the same address, the worker suffix distinguishes them. This should at minimum be stored on `DownstreamClient `even if not used yet.

nkatha23 (commented):
Code ACK https://github.com/braidpool/braidpool/pull/531/changes/25f6ea31d7fc951e46b03a1f73e2d0893b204abe

Issue #546, opened 2026-09-12
Panic in MerklePathProof::get_merkle_hashing_order on an empty merkle path


Atishyy27:
calculate_corresponding_merkle_root panics for a MerklePathProof with is_right_leaf = true and an empty merkle_path. get_merkle_hashing_order indexes self.merkle_path[0] in the is_right_leaf branch (braidpool-primitives/src/utils/bitcoin/mod.rs:36) before checking the path is non-empty, so an empty path is an out-of-bounds index.

Found while adding a cargo-fuzz target for the primitives (PR to follow); the target hits it within seconds of a run.

There are no callers yet, but MerklePathProof is a proof verified against peer-supplied data, so it should reject a malformed path rather than panic. I can send the fix; the open question is the contract, whether calculate_corresponding_merkle_root should return Option<MerkleRoot> or a Result, since it currently returns MerkleRoot directly.

zaidmstrr:
The issue you are stating is not part of the current `dev` branch, which is our current working branch. You might have pointed to the `main` which is currently outdated. So this is not valid for now.

Atishyy27:
Closing this. It's against braidpool-primitives on main, which is frozen and not part of dev, where all real development happens. The code this bug was in doesn't exist on dev. Apologies for the noise.

Issue #133, opened 2025-04-16
Set Up Fuzzing Framework for Braidpool


zaidmstrr:
This issue describes why fuzzing is important for a project like Braidpool and also keeps track of the current stage of implementation. 

You might think, Why do we need to fuzz Braidpool? Well, the answer is quite simple—we want robustness and security. Fuzzing is a testing methodology that is used in many security-critical and high-assurance softwares, to protect system integrity. Braidpool also lies in this category, so we need extra precautions. In general, fuzzing is a technique to find bugs/vulnerabilities like crashes, memory leaks, buffer overflows, or logic errors. These types of errors are common and hard to catch by other types of testing like functional or unit testing.

The question that can rise in your mind is, "Is code written using Rust also vulnerable to these flaws (mentioned above), as Rust is a memory-safe language? And the answer is "yes", fuzzing still plays a critical role in catching logic bugs, panics, and security vulnerabilities that type-checking alone can’t catch.

For Braidpool, we can implement different types of fuzzing, e.g., **structure-aware**, **property-based**, **stateful**, **differential**, and **mutation-based**, but we will start with coverage-guided mutation-based fuzzing. It's powerful and easy to integrate with the current codebase, and then later we can implement all the other fuzzing strategies. There are many potential targets in Braidpool for which we can write fuzz tests, so soon I will disclose some of these and their implementation. As the Briadpool codebase increases, the more possible fuzz targets are generated, which means we want more coverage and robustness. Fuzzing can be very critical to derivatives and payout mechanisms, which we will introduce in a while.

We are planning to use [libFuzzer](https://github.com/rust-fuzz/cargo-fuzz) which is simply provided by cargo-fuzz in Rust crates. This fuzzer is capable of providing most of the things we need right now and in the future.

###   Stage 1

- [ ]  Initializing the fuzz directory structure

- [ ] Writing targets for system critical functions (consenses, validation etc)

### Stage 2

- [ ]  Adding different fuzzing strategies

Feel free to contribute ideas, fuzz targets, or improvements below 👇

anuragShingare30:
@Sansh2356 Can I look into this?

zaidmstrr:
Thanks for your interest. We haven't decided to move on with this issue yet.

pradhyum6144:
Hi @zaidmstrr, I saw your earlier reply that you hadn't decided to move forward with this yet. Has that changed in the last couple of months, or is it still on hold?

[...]

PR #547, opened 2026-09-12
fuzz: add cargo-fuzz harness with a merkle path proof target

Atishyy27:
Sets up the fuzzing framework from #133 using cargo-fuzz.

It adds a fuzz workspace under braidpool-primitives/fuzz with one target, merkle_path_proof, that builds an arbitrary MerklePathProof and runs calculate_corresponding_merkle_root, covering empty and single-element paths on both is_right_leaf branches. A short README covers running it and adding targets. The target is kept thin so it can be retargeted as the core structures change, and more can follow as the types stabilise.

On its first run it surfaces a panic in get_merkle_hashing_order for an empty merkle path with is_right_leaf set, where merkle_path[0] is indexed without a bounds check. I reported that in #546 so the fix can be reviewed on its own rather than bundled here.

To run: cargo +nightly fuzz run merkle_path_proof

### DIFF
--- a/braidpool-primitives/fuzz/.gitignore
+++ b/braidpool-primitives/fuzz/.gitignore
@@ -1,0 +1,5 @@
+target
+corpus
+artifacts
+coverage
+Cargo.lock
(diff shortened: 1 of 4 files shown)


zaidmstrr:
This PR is not valid against the issue it refers to. The wrong branch is selected. Requesting to close this PR.

Atishyy27:
Closing this. It targets main, which turns out to be frozen (last commit 2025-06-02) — real development is on dev, and braidpool-primitives (the crate this fuzz target was built against) isn't part of dev at all; the bead/braid code there lives under node/src instead. I'll requeue #133 against the actual dev code.

PR #556, opened 2026-09-20
node: key block template staleness on the previous block hash

amritraj7237:
`block_template.rs:63` compares heights to drop stale templates, which also drops the template built on the new tip after a reorg back to the same height — exactly the case we need. Comparing `(height, previous_block_hash)` keeps it.

Worth noting the stale-template case the original comment describes can't actually happen today: `fetcher` is awaited inline by the ZMQ loop and sends exactly one template before returning, so arrivals are strictly ordered. That ordering is what makes "anything that isn't a repeat is new" safe, and the comment says so — if `fetcher` is ever spawned concurrently this needs a sequence number instead.

`BlockHash` comes from `bitcoincore_rpc_json`'s re-export rather than the top-level `bitcoin` dependency, because the lockfile carries three different `bitcoin` versions (0.32.4-cpunet, 0.32.5, 0.33.0-alpha.0) and the direct import risks a different type.

Tested with `cargo check`, `cargo fmt --all -- --check` and `cargo test`.


priyashuu:
Hey @amritraj7237, you should change the base branch from `main` to `dev` and make sure that the issue you’re solving still exists in the `dev` branch, since the latest changes are there.

If the issue no longer exists in `dev`, please close this PR.

zaidmstrr (commented):
Does not exist in the current `dev` branch.

PR #557, opened 2026-09-20
test: write the save/load temp file outside the fixture directory

amritraj7237:
`test_save_load_braid` (`braid_tests.rs:663`) creates `temp_braid_test.json` inside `tests/braids/` — the directory eight other tests in the same file enumerate with `read_dir` while it runs. They pick up the entry, then fail to open it once the file has been removed.

On ext4 the window is small enough that it never fires; I ran the suite there 8 times unpatched with no failure, so CI isn't affected. On a Windows drive mounted under WSL (`/mnt/c`), where deletions take longer to become visible, it fails 4-8 tests on every single `cargo test` run:

```
test result: FAILED. 11 passed; 8 failed;
test result: FAILED. 11 passed; 8 failed;
test result: FAILED. 15 passed; 4 failed;
```

which makes the suite unusable on a Windows checkout without `--test-threads=1`. Writing to `std::env::temp_dir()` leaves the fixture directory read-only for the suite.

5/5 parallel runs green on /mnt/c afterwards, still 19/19 on ext4.


priyashuu:

zaidmstrr (commented):
Created against the wrong branch, and this issue doesn't exist on the current `dev` branch.

PR #558, opened 2026-09-20
node: keep the ZeroMQ listener alive and supervise it

amritraj7237:
Two separate problems that were tangled together in `zmq.rs:26`.

A single unparseable notification did `return Err(err)`, ending the hashblock subscription for the rest of the node's life. The ZeroMQ socket reconnects on its own, so this is transient — log it and keep listening.

The stream *ending* is the opposite case and was being ignored. `main` spawned the listener and dropped its `JoinHandle`, so if it ever returned the node carried on with a block template that would never refresh again: mining on a stale tip, no log, no exit. `main` now awaits the handle and halts if it exits for any reason, panics included.

Stopping is the conservative choice here. Mining on a template that can't be updated is worse than being down, since it's silent.

Tested with `cargo check`, `cargo fmt --all -- --check` and `cargo test`.


priyashuu:

zaidmstrr (commented):
We are not using this feature, and the targeted branch is incorrect; the current working branch is `main`

PR #559, opened 2026-09-20
node: handle peer and RPC failures instead of panicking

amritraj7237:
Four places where routine failures take down a connection task or the whole node.

`connection.rs:72` deserializes peer frames with `unwrap()`. Any peer can panic its connection task by sending a well-framed payload that isn't a flexbuffer, and no handshake is needed to get there — `start_from_accept` goes straight into the read loop. `message_received` already returns a `Result` whose `Err` closes the connection, so it just uses that.

`main.rs:78` throws away the address `accept()` returns and rereads it with `stream.peer_addr()?`. That fails with ENOTCONN once the peer has sent RST, and the `?` propagates out of `main`, so connect-then-immediately-close kills the listener and the process. Using the address we were already given fixes it and removes a syscall.

`main.rs:54` panics on an unreachable `--addnode` seed, before the listener has even bound, although the option is documented as repeatable. It now logs and moves to the next seed.

`rpc.rs:49` calls `get_blockchain_info().unwrap()` on the line before the `if let Err(..)` that prints the cookie-authentication hint, so that hint is unreachable — and `get_best_block_hash()?` at line 46 returns even earlier. Since `--rpccookie` defaults to `~/.bitcoin/.cookie`, the default path is the one that hits it. Now one liveness check, reported properly.

Tested with `cargo check`, `cargo fmt --all -- --check` and `cargo test`. No new tests: all four are on I/O paths that need a live bitcoind or a socket peer, and there's no harness for either yet. Glad to add one if preferred.


priyashuu:

zaidmstrr (commented):
This issue doesn't exist in the current `dev` branch, which is our current working branch. Closing this for now.

